PR #16, opened 2021-07-19
Add tooltip desribing beneficiary

nervetrip:


packages/app/src/components/Dashboard/PayoutModsList.tsx
```diff
@@ -106,9 +107,14 @@ export default function PayoutModsList({
                     @<ProjectHandle projectId={m.projectId} />:
                   </div>
                   <div
-                    style={{ fontSize: '.8rem', color: colors.text.secondary }}
+                    style={{ fontSize: '.8rem', color: colors.text.secondary, marginLeft: 10 }}
                   >
-                    Beneficiary: <FormattedAddress address={m.beneficiary} />
+                    <TooltipLabel
+                      label={'Beneficiary:'}
+                      tip={`This address will receive any tokens minted when the recipient project gets paid.`}
```
nervetrip, line 114: I really wanted to do this:
```
`This address will receive any tokens minted when ${m.projectName} gets paid.`
```

But the only way I found to lookup the project name was via the `<ProjectHandle>` component. And the Tooltip component doesn't allow JSXElements as a `tip` value.

Since that was going to be a fair bit more effort, I figured it was best to get this landed first.

PR #19, opened 2021-07-19
Fix over fetching

teddywilson:
I'm fairly convinced that the memoized filter does not need to be a part of the `useEffect()`(s) for fetching events; it would make since that this is causing the effects to refresh several times. It would be great if @peripheralist could verify this.

On the other hand, as exemplified by the tiles payment history, there's probably no need to fetch > 3000 events when the page loads. We can use the queryFilter `fromBlock` and `toBlock` params to get around this.


peripheralist:
i may be missing something, but i've always assumed effects should list any value that can change without a parent re-render in their dependencies, which makes me think `filter` does belong there. do you have a good resource that would confirm taking it out?

also worth noting that right now, eventListener is only used for getting Pay events in the app, and i think those events will be one of the first things i'd like to deprecate with the subgraph. if you haven't noticed yet (and admittedly this might be due to something i'm missing) the pay events aren't always loaded reliably

teddywilson:
As discussed over chat, due to the ambiguity, we can instead stringify the memoized object to achieve the same effect.

PR #42, opened 2021-07-26
bugfix for displaying payouts with no funding limit

bonkrat:
Fix for https://github.com/jbx-protocol/juicehouse/issues/41

Screenshot 2021-07-26 181737

Screenshot 2021-07-26 181757


peripheralist:
👏

PR #60, opened 2021-08-09
Change project id to hex string format

odd-amphora:
I think that this probably makes more sense if people are trying to grab their project id to deploy a contract/run a certain function.

e.g., 12 ->0x0c which is a bit misleading

Screen Shot 2021-08-09 at 6 28 03 PM


peripheralist:
hmmm mixed feels. we aren't showing hex formatted numbers anywhere else, and i imagine any app that needs that hex string for a contract call will be used to converting numbers to hex format. i think it'll be useful to people for more than just plugging into code—it's also the order number of when your project was created

nnnnicholas:
I prefer project as a human readable number. Perhaps we could add the HEX formatted number to the right-side menu that pops out when you click on the wrench?

image
image

PR #67, opened 2021-08-11
Upgrade walletconnect

odd-amphora:
Looks like wallet connect isn't working currently with gnosis. Upgrading it solves the issue as they fixed a bunch of bugs in 1.4+. Tested locally and it appears to be working now.


peripheralist (approved):
thank u!

PR #101, opened 2021-08-21
Move example contracts into example directory

odd-amphora:


mejango (changes requested):
Get this "Showtime" business outta here. It's Shwotime.

PR #121, opened 2021-09-07
Add blog to navbar

odd-amphora:
Also a few cleanups:
* Reduce navbar padding on mobile (has always felt a little wide for me)
* Abstract menu so we don't have to duplicate it for mobile/desktop


peripheralist (approved):
love it, thanks!

PR #131, opened 2021-09-10
update wording and add a line between two sections

### LINKED ISSUE #86: Wording
The top section should reference the fact that the overflow also becomes your runway.
---
When both toggles are turned on, the section run into one another. itd be nice to have a line, or slight color change, between them.

image

tjrvl:
fixes #86 

f3


peripheralist:
great, thanks! just made one change with how the color is used—no need to rely on `isDarkMode` :)

PR #147, opened 2021-09-20
[WIP][DO NOT MERGE] Support subgraph on multiple networks

odd-amphora:


peripheralist:
@odd-amphora how are files in /generated created? also i have a /build directory, but i think this may be obsolete?

image

odd-amphora:
@peripheralist `./generated` is created after `yarn prepare:{network}` and `./build` is created on deploy. That reminds me – these should be added to the standalone .gitignore. Done.

peripheralist (approved):

PR #155, opened 2021-09-23
Use contracts NPM dependency and new format

odd-amphora:


peripheralist (commented):
Huge! this feels great. any reason to not use the numerated ContractName? makes creating instances of useContractReader a little easier i think

odd-amphora:
@peripheralist done!

peripheralist (approved):

PR #201, opened 2021-11-17
fix/bugs

palevoo:


peripheralist:
hey @palevoo I feel like we discussed this somewhere in discord or on a call, my bad for never commenting here as now I can't remember where this got left off.

you were playing with an interesting `isWindowVisible` pattern here—do you want to save this or should I close it?

Issue #203, opened 2021-11-17
[BUG] Link on FAQs is broken


mporracindie:
**Describe the bug**
On the [homepage](https://juicebox.money/#/) under FAQs the second question `What does Juicebox cost?` has a link that does work.
https://juicebox.money/#/p/constitutiondaojuicebox

mporracindie:
Here is the line to fix: https://github.com/jbx-protocol/juice-interface/blob/main/src/components/Landing/Faq.tsx#L22

mporracindie:
My mistake but the bug is still there. {window.location.href} does not always point to `https://juicebox.money/`

peripheralist:
fixed

PR #204, opened 2021-11-17
Fix broken link

mporracindie:
{window.location.href} does not always point to https://juicebox.money/

https://github.com/jbx-protocol/juice-interface/issues/203

### DIFF
--- a/src/components/Landing/Faq.tsx
+++ b/src/components/Landing/Faq.tsx
@@ -19,7 +19,7 @@ const QAs: {
   {
     q: `What does Juicebox cost?`,
     a: [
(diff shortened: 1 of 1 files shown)


peripheralist:
thanks for catching the bug and for the PR @mporracindie !

could we also make this a <a></a> tag so it's clickable?

mporracindie:
> thanks for catching the bug and for the PR @mporracindie !
> 
> could we also make this a tag so it's clickable?

I would but tbh I'm not familiar with ts at all and I wasn't able to set up the env to test it locally :(

peripheralist:
all good @mporracindie !

i'll fix in a separate branch and close this, but thanks for your help in spotting it and throwing in a PR 🤝

if you're interested in contributing elsewhere lmk and i'll make sure you can get set up running locally

PR #205, opened 2021-11-18
feat/projects-warning

palevoo:


nnnnicholas (changes requested):
> The Juicebox protocol is open to anyone. There are risks associated with interacting with all projects on the protocol. Projects built on the protocol are not endorsed or vetted by JuiceboxDAO, so you should do your own research before committing your funds.

peripheralist (approved):

Issue #145, opened 2021-09-20
Project pages should have project-specific OpenGraph metadata


nnnnicholas:
When shared in Discord and Twitter, project pages currently show Open Graph data for Juicebox: An image of Banny, the title "Juicebox", and a description of the protocol. This makes all project page shares look the same. 

## Issues

- [x] https://github.com/jbx-protocol/juice-interface/issues/1144

**Proposed solution**
Populate Open Graph metadata for project pages with project-specific metadata (title, image, description).

**Screenshots**
BrainDAO ograph data:
![image](https://user-images.githubusercontent.com/1418561/133950724-a1002fc2-b813-4363-9e64-2e926135d362.png)
![image](https://user-images.githubusercontent.com/1418561/133950847-3a261951-ffe4-4aba-81ae-8bd042d87186.png)

peripheralist:
the main thing preventing us from doing this is the app's "hash routing", which apps hosted on IPFS need to use to be compatible with IPFS gateway route resolving. you'll notice that JB (and a lot of other IPFS apps, e.g. Uniswap) use a # in the url route. it makes it impossible to dynamically load <html/> holding the opengraph info for a specific page—we can only do it for the top level domain.

that said, there is probably a clever workaround we could come up with to make it still work, and i'd love to find it. anyone can feel free to tackle this.
 
![image](https://user-images.githubusercontent.com/79433522/134229023-fd6f3de5-6ca4-4f6e-84d0-17aca1635b21.png)

peripheralist:
assigning this to you @palevoo since you're working on this, but i know it's still blocked for now so feel free to unassign if you get to an impass and wanna focus elsewhere

jnoh:
One possible partial solution to this is to migrate from create-react-app to next.js so that we can get the [static generation](https://nextjs.org/docs/basic-features/pages#static-generation-recommended) magic that it provides (maybe we could figure out how to do this with create-react-app as well).

We can statically generate the top N projects so that there meta tags are properly filled in at build time. For any project that doesn't fall within N we would use generic text as we are doing now.

We'd want to keep it in N so that build times don't get out of hand as projects grow.

Not a full solution but an idea.

---

One total solution is to stop using IPFS/Fleek and have a proper backend so that we can SSR.

peripheralist:
[...]

PR #209, opened 2021-11-18
feat/project-metadatas

palevoo:
So with this PR when people share link of their project's page, metadata for preview on the web (discord, twitter, telegram), should show information specific to the project instead of generic Juicebox one, as well as in the tab title they will see the dao name too.

I've used existing metadata from project such as name, description, logoUrl and twitter. In future we could further improve this by creating auto-generated preview images from provided project info

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,6 +22,7 @@
     "ethers": "^5.1.0",
     "react": "^17",
     "react-dom": "^17.0.1",
+    "react-helmet": "^6.1.0",
     "react-redux": "^7.2.2",
     "react-router-dom": "^5.2.0",
     "react-scripts": "4.0.1",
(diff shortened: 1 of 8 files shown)


peripheralist:
so stoked on this. @palevoo could you add a description here explaining what we're looking at?

odd-amphora (changes requested):
Let's make sure we get a working preview before this is merged. But super great work on getting this started!

peripheralist:
@palevoo from your recent discord post it sounds like this isn't going to work as expected? should we leave this open?

palevoo:
@peripheralist yep, that is true, I'm still looking around into possible workarounds, will close this one if don't come up with anything, in meantime I've opened a PR #213 with only updating client side `document.tile` with project's name

peripheralist:
this relates to #145

peripheralist:
closing this since we're now exploring other options for supporting custom project page meta tags

PR #212, opened 2021-11-19
fix/default-payout-address

### LINKED ISSUE #200: [BUG] Project handle form input doesn't autofill the defaultValue
image

Even when this form has a value for the address, the form input will not populate it and requires the user to retype it.

palevoo:
add autofill when wallet is connected fixes #200


peripheralist:
I'm not sure if i'm following here. The goal should be to populate the field with the address for a payout if one's already been defined, and the user is editing it. If one hasn't been set yet we can probably default to empty—autofilling to the project owner might be confusing since funds go to the project owner by default.

peripheralist:
closing this as it was fixed by another PR

PR #215, opened 2021-11-20
Add eslint helpers and cleanup eslint warnings

jnoh:
As more people come into the project, a clean workspace becomes essential. 

**Dev Workflow Modifications**  
* Adds pre-commit hook via husky that automatically runs eslint and prettier before every commit
* Switch vscode recommended plugin to eslint and prettier from [prettier-eslint](https://marketplace.visualstudio.com/items?itemName=rvest.vs-code-prettier-eslint). Prettier-eslint uses it's own version of prettier and eslint and collides with the project's eslint and prettier versions.
* VSCode setting to format automatically on save.

**New eslint Rules**  
* Error on @typescript-eslint/no-unused-vars 
* Error on @typescript-eslint/no-redeclare

**Clean Up**  
* Cleans up `@typescript-eslint/no-unused-vars` warnings
* Cleans up `jsx-a11y/alt-text` warnings
* Cleans up `no-useless-escape` warnings
* Cleans up `@typescript-eslint/no-redeclare` warnings
* Cleans up `react-hooks/exhaustive-deps` warnings
* Wraps objects with useMemo and functions with useCallback to stop unnecessary rerenders


peripheralist:
thanks for giving this some attention 🙏 

what part of these changes prevents checking in code with errors?
should we use a pre-commit hook like husky to run lint check + cleanup?

jnoh:
@peripheralist eslint is run as a part of the `fleek-ci` check and will fail the build - so the gate exists at merge to main.

We can add a pre-commit hook if we want to the check to be more explicit. My suggestion is that we add this if we need it pulled out as to keep things simpler for the time being. Adding husky is one more point of complexity to onboarding new devs.

What do you think?

peripheralist:
I'm leaning towards using husky so that the commit failure would be caught on commit rather than during the fleek build, which would take a few minutes. If we're going to enforce these rules (which we should) we might as well be running lint:fix before each commit, and I've never felt like husky created any added friction for new devs aside from adding a couple seconds to each commit

src/User.tsx
```diff
@@ -17,10 +17,12 @@ export default function User({ children }: { children: ChildElems }) {
   })
 
   useEffect(() => {
-    contracts?.TerminalV1.functions.fee().then((res: [BigNumber]) => {
-      if (!adminFeePercent?.eq(res[0])) setAdminFeePercent(res[0])
-    })
-  }, [setAdminFeePercent, contracts])
+    async function fetchData() {
```
jnoh, line 20: Prefer async/await over promise chains

src/components/App.tsx
```diff
@@ -26,7 +26,7 @@ function App() {
     if (!signerNetwork) return
 
     setSwitchNetworkModalVisible(signerNetwork !== networkName)
-  }, [setSwitchNetworkModalVisible, signerNetwork])
+  }, [networkName, signerNetwork])
```
jnoh, line 29: setState methods are stable and don't need to be included in deps

image
([source](https://reactjs.org/docs/hooks-reference.html))

src/components/Create/index.tsx
```diff
@@ -342,52 +375,75 @@ export default function Create() {
...
-        queuedTicketMods: undefined,
-        balance: BigNumber.from(0),
-        balanceInCurrency: BigNumber.from(0),
-        tokenSymbol: undefined,
-        tokenAddress: constants.AddressZero,
-        isPreviewMode: true,
-        isArchived: false,
-      }}
-    >
+    <ProjectContext.Provider value={project}>
```
jnoh, line 446: Tons of unnecessary rerenders because the value is not memoized

src/components/shared/inputs/FormattedNumberInput.tsx
```diff
@@ -50,10 +51,10 @@ export default function FormattedNumberInput({
...
 
   useLayoutEffect(() => {
-    const accessory = document.getElementById(accessoryId)
-    if (!accessory) return
-    setAccessoryWidth(accessory.clientWidth)
-  }, [accessory, accessoryId])
+    const accessoryContainer = document.getElementById(accessoryId)
+    if (!accessoryContainer) return
+    setAccessoryWidth(accessoryContainer.clientWidth)
+  }, [accessoryId])
```
jnoh, line 57: Should not be referencing accessory prop, should reference container of accessory

src/components/shared/inputs/ImageUploader.tsx
```diff
@@ -41,6 +41,7 @@ export default function ImageUploader({
       <Col xs={24} md={7}>
         <Space align="start">
           {url && (
+            // eslint-disable-next-line jsx-a11y/img-redundant-alt
```
jnoh, line 44: Disables warning for using the word image in the alt text

jnoh:
@peripheralist whenever you get a chance, can you give this a review? (see the PR description for all the changes made)

src/components/App.tsx
peripheralist: awesome. pretty sure this needs correcting in a lot of places

src/components/Create/IncentivesForm.tsx
```diff
@@ -30,7 +30,7 @@ export default function IncentivesForm({
   useLayoutEffect(() => {
     setDiscountRate(initialDiscountRate)
     setBondingCurveRate(initialBondingCurveRate)
-  }, [])
+  }, [initialBondingCurveRate, initialDiscountRate])
```
peripheralist, line 33: some of these dependency arrays were empty to ensure they only run once when the component loads. i'm not sure if a better practice there is to just omit the second argument to useEffect entirely. 

have you tested to see if adding dependencies breaks anything?

peripheralist (commented):
So stoked for these changes, the codebase really needed this kind of housekeeping love.

How are you catching missing effect dependencies—is it the linter or are you looking through manually? I haven't played around with your latest yet but I'm curious if adding these will cause unexpected behavior anywhere.

also, great alt texts on the landing page

src/components/Create/IncentivesForm.tsx
jnoh: Using useEffect without the dependency array will cause it to run on every render which is probably not what we want in these cases ([source](https://reactjs.org/docs/hooks-reference.html#useeffect)).

From my experience, I've found that not using exhaustive-deps typically is a sign that useEffect is not being used properly:
image

The FAQ [here](https://reactjs.org/docs/hooks-faq.html#is-it-safe-to-omit-functions-from-the-list-of-dependencies) describes a bunch of reasons why deps should be exhaustive and why exhaustive-deps is a lint rule.

I manually updated each of the dependencies that the linter caught. You probably noticed some of the refactoring tha had to happen as a result.

jnoh:
@peripheralist As mentioned above, I used the linter to catch the `exhaustive-dep` issues and I manually updated those issues by hand. I also checked the product to ensure things are working as expected. Dependency issues tend to be pretty easy to catch because it typically causes infinite rerendering. 

I spent the majority of time perfecting/refactoring the alt text, so appreciate that you noticed 😆

src/components/App.tsx
jnoh: I didn't catch them all but I think I got a bunch of them.

peripheralist:
Solid work @jnoh, been playing around and everything feels smooth.

Last thoughts:

- Did you remove the lint:fix command? It'd be nice to have a command to fix simple things like unused imports across the whole codebase.

- Since Prettier will get run in the precommit hook, I'm not sure we need to enforce format on save. Wdyt?

jnoh:
🙏 @peripheralist 

> Did you remove the lint:fix command? It'd be nice to have a command to fix simple things like unused imports across the whole codebase.
Yeah I agree. Just added it back in. `yarn lint-staged` also runs `eslint --fix` (you'll see it specified in the package.json under the "lint-staged". But might be nice to have a helper that's isolated from prettier.

> Since Prettier will get run in the precommit hook, I'm not sure we need to enforce format on save. Wdyt?
I think the auto formatting on save is a really positive mod on dev workflows across the board. I think in a similar way to your line of thinking of moving the check to commit instead of build, moving it to save speeds up the feedback loop. I save frequently and constantly to check that my code is free of syntax errors and also to clean up really large tsx-prop/interfaces/objects structures as I write them. Would you be open to test-driving it for some trial period? We can remove it if we feel that it's too invasive of an experience.

peripheralist:
@jnoh My take would be:

Have Prettier check for linting, but not actually run a fix. Having your code get changed _after_ you're done working with it feels awk, especially with linting (vs formatting). I've had cases before where the lint fix made breaking changes, like with order of imports that were then committed without me realizing.

For the format-on-save, I think it's a great workflow, but don't see why a project should impose it on devs. Devs should be able to turn it off so long as the code is compliant before it's committed.

jnoh:
Cool thanks for the thoughtfulness here. 

Just to clarify. Your request is to:
* Remove the automatic fixes for both eslint the precommit hook. The precommit hook would fail if there was a lint issue. In these cases, devs would should use the automatic lint fix helper or go in and lint by hand?
* Same as above for prettier
* Remove the autoformat on save

peripheralist:
copying my latest from our discord convo here for documentation 👍 

> my only reservations about having the precommit hook run a lint fix come from past experience, where sometimes those fixes can cause breaking changes that won't be caught by the dev until after they've been committed. the chances of that happening are certainly slim. but since code is often getting committed/merged quickly in juicebox and deployments happen automatically, i'd rather minimize the chance of shipping any bugs

> regarding the autosave/formatting, i'm definitely in favor of running prettier to format on pre-commit (just not lint fixes). in that case the code should be format-compliant before ever looking at a diff, no?

peripheralist:
@jnoh could we make sure this also enforces some strategy for ordered imports? if it's supposed to be doing that already it doesn't seem to be working for me

jnoh:
To get autoformatting on save for import ordering, you can add this to your vscode settings.json file:
```
{
   ...
    "eslint.options": {
        "extensions": [
            ".js",
            ".jsx",
            ".ts",
            ".tsx"
        ]
    },
    "eslint.validate": ["javascript", "javascriptreact", "typescript", "typescriptreact"],
    "editor.codeActionsOnSave": {
        "source.fixAll.eslint": true
    },
   ...
}
```

jnoh:
@peripheralist I think after we tweak this ordered imports, we ought to merge this and adjust linting rules as we go. Want to merge this PR as it's quite large as it is.

.eslintrc.json
```diff
@@ -0,0 +1,16 @@
+{
+  "plugins": ["import"],
+  "extends": ["react-app", "react-app/jest", "prettier"],
+  "rules": {
+    "react-hooks/exhaustive-deps": "error",
+    "@typescript-eslint/no-redeclare": "error",
+    "@typescript-eslint/no-unused-vars": "error",
+    "import/order": [
+      "warn",
```
jnoh, line 9: Warn to start since this will cause builds to break. We'll add a followup PR will update all the imports and flip this to an error.

peripheralist:
awesome. happy to keep tweaking rules as we go. let's merge it in

peripheralist (approved):

PR #217, opened 2021-11-20
Proposal: Use React Query as the "data fetching"/"server state management" implementation

torvusbug:
I've used [React Query](https://react-query.tanstack.com) professionally in multiple projects and I swear by it. It's a great way to manage data fetching to provide solid and fast user experiences, as well as great developer experience with the built-in devtools. Loading and error states, cache invalidation, manual refetching, and cross-page data sharing are a breeze.

After poking around Juicebox for a while, I noticed the Projects list loaded fresh every visit, disrupting the browser-back scroll position. The app also loads Project details twice, once for the grid cell and a second time when clicking into a Project. This can be solved with a good [key strategy](https://react-query.tanstack.com/guides/query-keys), such that a Project's "details" are fetched with a consistent key used on multiple screens.

Since Juicebox is currently an exclusively client-side rendered app, we could even get away with localStorage-based caching. Say a 24-48 hour TTL, we can provide great UX speed to returning users. That's out of scope of this proposal, IMO, but something [relatively easy to implement](https://react-query.tanstack.com/plugins/createWebStoragePersistor).

I noticed much of the API data handling is already in isolated React Hooks, a great practice, making a migration to React Query even easier.

Here's a walkthrough example of a change I made in this PR:

In `src/hooks/Projects.ts`, the `projects` state was managed by a `useState` hook. It then filtered the response into archived _and_ active projects regardless of the selected `filter`.

Before getting into what I changed in this hook, I created a new hook `useSubgraphQuery` that wraps the existing Graph API. It uses the entire options payload as a key, which should automatically handle paging, filtering, etc.: any change to the options will create a new request.

The change I made to this hook is to return the entire React Query response. It uses the `useSubgraphQuery` hook to get the list of projects, and I overwrote the returned `data` property to add the custom filtering logic, filtering only the currently selected filter.

Since I changed the return value of the projects hook, I had to change its usage. Instead of `const projects = ...`, it's now `const { data: projects } = ...`. You can have easy access to the other properties  `error`, `isLoading`, `refetch`, etc.

I also changed the gas price hook, removing usage of the custom polling hook as React Query has this built-in.

This is my first contribution, and I'm interested in anyone's thoughts.


peripheralist:
thanks for the pr and the clear description @torvusbug !

stoked to play around with this and learn some more about React Query. i'll share some thoughts on this soon

peripheralist:
@torvusbug pulled this down to play around with it. it definitely cuts down querying a lot and makes the app feel smoother. i'm excited ❤️ 

To avoid having multiple patterns in the app at the same time, let's expand this PR to replace the other calls to the graph with react-query hooks as you've already done with useProjects. this would be
- pay events
- withdraw (tap) events (and subsequent `distributeToPayoutModEvent` queries)
- redeem events
- distribute reserves events
- project participants (holders modal in the UI)
- calls used in the balance/volume graph (BalanceTimeline.tsx)

What are your thoughts on using react-query to also:
- replace the current ENS cacheing system in FormattedAddress.tsx?
- avoid redundant contract calls from the useContractReader hook?

^happy to leave these outside the scope of this PR

torvusbug:
@peripheralist 

To your first point:
> can we expand this PR to replace the other calls to the graph with react-query hooks as you've already done with useProjects?

Absolutely. After I posted this PR, I realized a critical mistake in my PR description was a lack of "Next Steps." Before integrating React Query, I looked through the usage of the existing subgraph function, which led me to make such a generic `useSubgraphQuery` hook. I believe it should cover any and all use cases for The Graph interaction, specifically improved error handling. I can certainly begin tackling the aforementioned usage of The Graph API.

On your second question about React Query usage in `FormattedAddress.tsx` and a cachable `useContractReader` hook:

I need to understand the fundamentals of these two pieces of business logic a bit better before I can say exactly how they can benefit. However, I'm sure they can be simplified once the data fetching questions are answered more reliably.

peripheralist:
Sounds great, the generic `useSubgraphQuery` makes sense to me.

On `useContractReader`, this just needs to be responsible for making a single contract call and returning the response. The logic in that hook now is just for passing arguments, formatting the response, `updateOn` for configuring when a call should be made again, and `valueDidChange` a helper for checking if a new value changed from the previous, which helps prevent unnecessary state updates + renders. Wrapping it as a react-query hook would be great if it could either slim down the logic in the hook, or in any way help prevent redundant calls/renders, but I'm not totally sure what that would look like yet or if it'd be worth it.

Let's focus on the graph queries for this PR and come back to the rest afterwards

torvusbug:
@peripheralist Sounds good, I'll replace subgraph calls that are wrapped in `useEffect` where I can.

torvusbug:
@peripheralist I pushed new code in this PR.

A few notes:
- I modified the `useSubgraphQuery` to take care of the manual parsing from the subgraph API to the internal type structure (mostly converting strings to dates and to BigNumber). This seriously reduces code complexity, as you don't need to parse each entity type in a `useEffect` somewhere in a React component. `src/components/Landing/Payments.tsx` is a good example of this simplification.
- I removed the `useMemo` I noticed some React components return. If you're compelled to memo a component, it's probably better to use `React.memo`. If you disagree, I'm happy to add it back, but I avoid `useMemo` unless you have a serious user-perceptible performance issue.
- I added a few `key` props to some React array mapping I noticed were missing a key.

I didn't modify the subgraph API usage in two places:
- The second usage in the participants modal had a fairly unique interaction that can probably be solved by React Query's `useMutation` API, but I haven't given that a shot yet.
- The balance timeline has a very complex interaction that I'm _sure_ can be solved more simply, but that would take a lot more time to figure out.

src/hooks/SubgraphQuery.ts
```diff
@@ -0,0 +1,51 @@
...
+    async () => {
+      const response = await axios.post<{ data: SubgraphQueryReturnTypes[E] }>(
+        subgraphUrl,
+        {
+          query: formatGraphQuery(opts),
+        },
+        { headers: { 'Content-Type': 'application/json' } },
+      )
+
+      let data = response.data?.data
```
peripheralist, line 43: let's remove this line and pass directly to `formatGraphResponse`?

src/hooks/SubgraphQuery.ts
```diff
@@ -0,0 +1,51 @@
...
+    // This should _only_ happen in development
+    throw new Error('env.REACT_APP_SUBGRAPH_URL is missing')
+  }
+  return useQuery<
+    SubgraphEntities[E][],
+    unknown, // Specific error type?
+    SubgraphEntities[E][],
+    readonly [string, GraphQueryOpts<E>]
+  >(
+    ['subgraph-query', opts],
```
peripheralist, line 33: would it make sense to be able to override `subgraph-query` tag here? not entirely sure what's happening with that first argument to `useQuery`

src/models/subgraph-entities/participant.ts
```diff
@@ -3,17 +3,15 @@ import { BigNumber } from 'ethers'
 export type Participant = Partial<{
```
peripheralist, line 3: What's the reasoning behind turning these subgraph entity types into interfaces with some optional args vs. using Partial<>?

For the record i don't feel too strongly about using types over interfaces or vice versa, but the important bit to note is that these entities aren't always returned with all properties. These represent the "full" schema, but in a graph query only the properties specified in the `opts.keys` are returned.

it's been a back-of-mind task for a while to find a tighter system for this so if you're onto something better than i'm down to use it, just trying to make sure i'm on your level.

peripheralist:
This is great. The two specific graph query cases you mentioned can definitely be left alone for now. I assume you're talking about the "download" feature for participants data, which can be treated separately and actually needs a rework anyway. The BalanceTimeline is also an edge case and can be left alone for now—would also be happy to see improvements there if you have ideas.

I took a crack at solving that typing issue you ran into but wasn't able to solve it. My guess is the best solution would be a discriminated union type https://www.typescriptlang.org/docs/handbook/2/narrowing.html#discriminated-unions, but it'd probably end up adding more bloat to the code than it's worth. I'm happy to let this be for now—the code is clear and your inline comments are awesome 👍 

Added a few comments in the code. Also on useSubgraphQuery, wdyt about adding an arg to override the `staleTime`? Defaulting to 60s will probably make sense in most places (it takes about that long for events to be reflected in the graph data anyway), but for example the projects query could be several minutes or more. Although in the case of _creating_ a project, we'd want to ensure a new query is made so you don't have to wait for the thing you just created to show up.

torvusbug:
@peripheralist Thanks for reviewing, I'll answer your code comments soon.

On your question:
>  ... Also on useSubgraphQuery, wdyt about adding an arg to override the staleTime? ...

I set up the hook such that you can pass a second argument to any usage of `useSubgraphQuery` entitled "reactQueryOptions." This is passed directly to React Query, potentially overwriting anything supplied as a default to React Query. If you wanted a `staleTime` of 5 minutes, an `onSuccess` callback to invalidate another query, and a `refetchInterval` every 5 minutes, you can pass those arguments to your usage of `useSubgraphQuery` and it'll overwrite the default 60 second `staleTime`.

src/hooks/SubgraphQuery.ts
torvusbug: @peripheralist Sure, I can explain.

The first argument to React Query is a key. This can either be a single string or an array of serializable values. These are how multiple `useQuery`s with the same key on the same page can share state. I've used this to my advantage countless times, mostly to create seemingly isolated components that end up sharing data, without having to create and setup (with the right `value` at the right time) new React Contexts.

The keys are diff'd deterministically (more on that [in their docs](https://react-query.tanstack.com/guides/query-keys)) I've actually never thought of allowing a customizable query key base and I don't think it's very useful. I tend to create query keys as: what's the organizational or business "type" of data, and what are the parameters that make _this_ one unique? So in the case of `useSubgraphQuery`, you're asking for a subgraph query; so the key base (index 0) is "subgraph-query", and the parameters that make _that_ instance unique is the query structure itself.

src/hooks/SubgraphQuery.ts
torvusbug: @peripheralist Good call out. I'll change that.

src/models/subgraph-entities/participant.ts
torvusbug: @peripheralist Yeah, I've been considering what to do here exactly, as I see the difficulty. 

I didn't think about `opts.keys` for optional keys, I was only looking at the parseJson for each respective DTO. I could try to find a better TypeScript solution... Maybe using `Pick<>` with a dynamic second argument.

I'll try a few things but in the meantime, it'll be prudent to revert these changes back to exclusively `Partial` types.

src/hooks/SubgraphQuery.ts
peripheralist: got it, thanks for the explanation. sounds like we can keep it as you have it for now 👍

src/models/subgraph-entities/participant.ts
peripheralist: Cool. Using some type strategy that depends on `keys` would be cool, but I wouldn't consider it essential for this PR. Unless you feel like diving into it now I'm happy to just come back to it later

torvusbug:
@peripheralist Pushing up some code that should solve our Partial<> and keying issue.

When you pass `keys` to `useSubgraphQuery`, it now returns an array of objects that contain _only_ those keys (and their respective value from the entity). This improved TypeScript actually caught 3 missing properties at build time: one from my code in this PR (a missing `id` key), and two from `TapEventsElem.tsx` and `ReservesEventElem.tsx` that were accessing the value `modPercent` but the `modPercent` `key` was never passed to the subgraph query (even in the code before my changes, when it was in a `useEffect`).

This _does_ have a limitation... The returned object is specifically _not_ the entity you asked for: it's a newly constructed object. So if you had a component that expected a full `Project` prop, you couldn't pass this to it without _all_ the keys.

I couldn't figure out how to approach the `keys` type when it can contain a different entity in an object with _its_ keys. I changed "keys" in that case to a `string[]`, losing some TypeScript validation. It's used only in one place, so I figure the trade-off is worth it.

I converted the other entities to interfaces and removed `Partial<>`.

src/utils/graph.ts
```diff
@@ -79,25 +96,31 @@ export type BlockConfig = {
...
+export type EntityKeys<E extends EntityKey> = keyof SubgraphEntities[E]
+
+export interface GraphQueryOpts<E extends EntityKey, K extends EntityKeys<E>> {
   entity: E
   first?: number
   skip?: number
   orderBy?: keyof SubgraphEntities[E]
   block?: BlockConfig
+
+  // `keys` can be a mix of the entity's keys or an entity specifier with its own keys
```
peripheralist, line 108: appreciate adding in comments like this 🙏

src/components/Dashboard/ProjectActivity/PaymentActivity.tsx
```diff
@@ -60,98 +59,88 @@ export function PaymentActivity({
     return override ? override : e.note
   }
 
-  useEffect(() => {
-    setLoading(true)
```
peripheralist, line 64: looks like this `setLoading(true)` was removed from the ProjectActivity components, which still needs to be there. this is a callback to set a `loading` state on the parent component

as an aside, this pattern has always felt a bit awkward to me, so if you have any ideas for streamlining it with this new `useSubgraphQuery` pattern then feel free to play around!

src/components/Dashboard/ProjectActivity/PaymentActivity.tsx
```diff
@@ -60,98 +59,88 @@ export function PaymentActivity({
...
+            key: 'project',
+            value: projectId.toString(),
+          }
+        : undefined,
+    },
+    {
+      onSuccess: data => {
         setLoading(false)
-        setCount(newEvents.length)
+        setCount(data?.length)
```
peripheralist, line 80: something unexpected is happening with the onSuccess callbacks. seems like `setCount` isn't being called every time.

If I tab from Pay to Redeem events then back to Pay, for example, it'll show the `count` of Redeem events under the Pay tab. It might be something to do with the component not being destroyed, so the hook isn't reinitialized.

peripheralist:
> This improved TypeScript actually caught 3 missing properties at build time: one from my code in this PR (a missing `id` key), and two from `TapEventsElem.tsx` and `ReservesEventElem.tsx` that were accessing the value `modPercent` but the `modPercent` `key` was never passed to the subgraph query (even in the code before my changes, when it was in a `useEffect`).

This is fantastic. I love good typing.

Incredible work with this PR so far, gonna be a huge improvement to the app 🙏. I left comments on some small things that aren't working right. Let's get those solved, then this will be ready to merge

torvusbug:
@peripheralist I just pushed up code that should address your concerns. I refactored that Payment Activity tabbed section in a way that makes each tab responsible for its own data fetching and representing its loading state more accurately (ie: if one tab has already been fetched, it won't go get it again unless it's stale).

I created a new hook that wraps React Query's `useInfiniteQuery` to handle pagination, it's right below the `useSubgraphQuery` and I called it `useInfiniteSubgraphQuery`. I had to do this because I noticed in my existing code that the "Load More" button ended up replacing the entire data set rather than adding to it. Using the infinite query, we can concatenate each page into a stream of results. Switching tabs now keeps that full paged data set in memory, which is pretty nifty.

peripheralist:
Hm, seems like there's a few new problems now:

- Pay tab will load events, but other tabs won't load any
- "Load more" on Pay tab reloads the first page and appends it to the list before loading other pages

torvusbug:
@peripheralist Pretty sure I fixed it. I figured out what was happening: I didn't configure the `skip` and `first` properties correctly, so it was in effect skipping the first `pageSize` number of elements, and subsequent page loads were actually the first result set. It should be good now.

peripheralist:
nice! this looks great to me 🥳 

since this is such a big diff, it'd be great to get more feedback from some others before we merge (cc @aeolianeth @jnoh @bonkrat @palevoo @odd-amphora) otherwise let's ship it in the next day or two

src/hooks/EthBalance.tsx
```diff
@@ -1,26 +1,19 @@
 import { readProvider } from 'constants/readProvider'
-import { BigNumber } from 'ethers'
-import { useState } from 'react'
-
-import { usePoller } from './Poller'
+import { useQuery } from 'react-query'
 
 export function useEthBalance(address: string | undefined) {
```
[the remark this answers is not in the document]
torvusbug: @jnoh 

1. I tend to prefer it wrapped in its own hook, that way you can't get the key wrong if hand-coded a `useQuery` in multiple places. To move quickly, I've certainly _not_ done that and haven't had any problems, but it's typically with simpler REST endpoints.
2. `useXyzQuery` is a good idea. I don't have a preference, since a simple inspection of the hook or the return type can show you what's going on. If you feel strongly they should end in `Query`, I can make that change.
3. I don't have a preference here either, since TypeScript's inference should save you most of the time. Something you end up doing with react-query is passing an additional options parameter, which requires it to be typed (see SubgraphQuery.ts:32), it can look quite verbose to see an options param and a return value typed very similarly. The biggest limitation to a weakly-typed react-query usage is a strongly-typed Error value and in-hook key validation. I don't usually have problems with the lack of Error typing as I tend to wrap any "error" interaction in a function that can reliably get me a strongly-typed Error object (since all `catch` statements now return an `unknown` error anyways).

aeolianeth:
@torvusbug Fantastic work with this PR 🙏 

I have been out of the React ecosystem for a few years (my day job is 100% Vue) so I'm not in a position to give a strong opinion. I _will_ say that my experience with Apollo as a GraphQL client + state manager has been pretty good; from my understanding, React Query looks less "magical" (which tbh is probably a good thing) and it enables a more flexible approach.

A couple of questions: 
1. @peripheralist Do we currently have any caching strategy in place? In any case, will React Query's cache mechanism lead to unforeseen consequences? For example, for something like gas prices I would imagine that we _wouldn't_ want to cache this (or at least set cache TTL to 30seconds or something).
2. What is the testing experience like with React Query? We don't do any frontend testing yet but I imagine we will eventually.

peripheralist:
> 1. @peripheralist Do we currently have any caching strategy in place? In any case, will React Query's cache mechanism lead to unforeseen consequences? For example, for something like gas prices I would imagine that we _wouldn't_ want to cache this (or at least set cache TTL to 30seconds or something).

The only cacheing used right now is a one-off solution for cacheing ENS names, in `FormattedAddress.tsx`. This PR doesn't seem to affect that, and we should consider replacing that pattern with react-query at some point if possible.

torvusbug:
@aeolianeth 

1. Adding to @peripheralist's comment, you're right to be wary of aggressive caching strategies that lead to presented information ending up detrimentally stale. In the case of gas prices, keeping with the existing implementation, react-query will poll for new data every 30 seconds, which is also the stale time. I didn't include a background fetching option, so it'll only poll if the user is active on the page. When the user returns to the page after 30 seconds (and the data is significantly stale), react-query will get the latest.
2. I've implemented testing in react-query a couple of ways, but I prefer this method: use [MSW](https://mswjs.io/) to mock your endpoints and set it up to intercept requests your test is expected to make. Then simply render your components, and react-query and your components should make those requests as they would in a production environment, except returning your test data instead. I prefer this to another common approach of mocking fetch/axios because you're also testing the entire end-to-end HTTP transaction rather than just making sure axios was called with the right params. In short, if you're building React UI tests properly (IMO), react-query or a custom fetching function shouldn't have any bearing on your testing capability or quality. FWIW, the other method of testing react-query is to manually hydrate the query client's cache with predetermined keys and data. This is only useful to me in niche situations.

aeolianeth:
@torvusbug thanks for those details, this all sounds great and I'm excited about this change!

aeolianeth (approved):

src/hooks/EthBalance.tsx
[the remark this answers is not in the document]
jnoh: If we want to truly wrap react-query, I'd suggest we do something like this:
``` 
export function useEthBalance(address: string | undefined): number {
  const { data: balance } =  useQuery(
      ['eth-balance', address],
      async () => {
        if (!address) {
          throw new Error("Can't fetch ETH balance, address is undefined")
        }
        return readProvider.getBalance(address)
      },
      {
        enabled: !!address,
        staleTime: 30000,
        refetchInterval: 30000,
      },
    )
  }

  return balance
}
```

If we're ok with not fully wrapping react-query and returning react-query types, then I think appending Query to the hook name would go along way for readability and understanding what kind of hook it is

--- 

Not super opinionated on either approach, but I do think we should pick one or the other.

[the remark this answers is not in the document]
jnoh: Also, please consider this comment non-blocking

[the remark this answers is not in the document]
torvusbug: @jnoh I've done this in the past multiple times and it has ended up becoming a pain. The main reason is you lose a few critical elements from returning _only_ `data`, namely: `isLoading`, `refetch`, and `error`. While it's not _always_ useful to use those and you may end up using only `data` most of the time, it's a good idea to at least expose the other properties in the event you _do_ want to show a loading spinner somewhere.

I'll rename these queries to include the `Query` suffix, I think that's reasonable.

peripheralist:
Feeling great about this. @torvusbug let's just solve these conflicts and we can merge!

torvusbug:
@peripheralist I think we're all good on this now, I renamed the hooks.

peripheralist:
> @peripheralist I think we're all good on this now, I renamed the hooks.

great! i just meant the merge conflicts. let's solve those and we can merge

torvusbug:
@peripheralist I'm not seeing any merge conflicts. I can merge it into an up-to-date `main` locally. I can't figure out what I'm missing...

peripheralist:
hmm. have you pulled the latest from main? github doesn't seem to like it, i'm seeing this: 
image

torvusbug:
@peripheralist I think it was caused by changing our PR->main merging strategy from "merge commit" to "squash and rebase." During development, I was running `git fetch --all && git merge upstream/main` rather than `git fetch --all && git rebase upstream/main`. I really dislike running `git push --force`, but I think these conflicts would only be solved by sort of "rewriting history" and force pushing. 

In our developer documentation, we should note that we prefer to "squash and rebase" rather than "merge commit," and warn users to upstream rebase instead of upstream merge.

peripheralist:
Got it, that's good to know! Adding an instruction to the readme is a good idea

Issue #91, opened 2021-08-18
Text entry input on number slider is difficult


peripheralist:
![image](https://user-images.githubusercontent.com/79433522/129926307-839d08c3-4508-4b74-9cb5-dc34e003f3ab.png)

Typing numbers is glitchy due to custom parsing/formatting and can emit the wrong value

nnnnicholas:
+1 

On Brave on Ubuntu 20.04 percentages entered in Spending and Reserved Tokens modals emit incorrect values most of the time. Fewer clicks between spawning the modal and confirming the number appears to work better.

Behavior: enter 25%, click confirm, 2% was emitted.
Expected behavior: enter 25%, click confirm, 25% emitted.

aeolianeth:
@peripheralist I'd like to take on this issue. I've reproduced using `@nnnnicholas`'s steps above

PR #230, opened 2021-11-25
Fix issue with percentage sliders

aeolianeth:
Fixes https://github.com/jbx-protocol/juice-interface/issues/91.

Using the `InputNumber` input of `NumberSlider` caused a bug where only the first digit of the input would be saved.

It turns out `Form.Item` intercepts `change` events from nested `input` elements. If the input element isn't the direct child of `Form.Item`, bad things happen! Even though we are handling `onChange` in `NumberSlider`, it appears `Form.Item` overrides this at some point.

Removing the `name` prop from `Form.Item` fixes the bug by delegating this responsibility to `NumberSlider`'s `onChange` event.

Accessibility is unaffected by this change; the `name` prop doesn't add `name` attribute to the input field.

### Demo

https://user-images.githubusercontent.com/94939382/143429149-10e4ac43-1098-48a6-b732-eaa9695c6e45.mp4

### DIFF
--- a/src/components/shared/formItems/ProjectPayoutMods.tsx
+++ b/src/components/shared/formItems/ProjectPayoutMods.tsx
@@ -455,7 +455,7 @@ export default function ProjectPayoutMods({
               }
             />
           ) : null}
-          <Form.Item label="Percent" rules={[{ required: true }]} shouldUpdate>
+          <Form.Item label="Percent" rules={[{ required: true }]}>
             <div style={{ display: 'flex', alignItems: 'center' }}>
               <span style={{ flex: 1, marginRight: 10 }}>
                 <NumberSlider
--- a/src/components/shared/formItems/ProjectPayoutMods.tsx
+++ b/src/components/shared/formItems/ProjectPayoutMods.tsx
@@ -455,12 +455,7 @@ export default function ProjectPayoutMods({
               }
             />
           ) : null}
-          <Form.Item
-            name="percent"
-            label="Percent"
-            rules={[{ required: true }]}
-            shouldUpdate
-          >
+          <Form.Item label="Percent" rules={[{ required: true }]} shouldUpdate>
             <div style={{ display: 'flex', alignItems: 'center' }}>
               <span style={{ flex: 1, marginRight: 10 }}>
                 <NumberSlider
--- a/src/components/shared/formItems/ProjectTicketMods.tsx
+++ b/src/components/shared/formItems/ProjectTicketMods.tsx
@@ -297,12 +297,7 @@ export default function ProjectTicketMods({
             }
           />
 
-          <Form.Item
-            name="percent"
-            label="Percent"
-            rules={[{ required: true }]}
-            shouldUpdate
-          >
+          <Form.Item label="Percent" rules={[{ required: true }]} shouldUpdate>
             <NumberSlider
               onChange={percent => form.setFieldsValue({ percent })}
               step={0.01}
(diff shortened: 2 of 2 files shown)


aeolianeth:
Hey there @peripheralist, could you please review this PR? 🙇

peripheralist:
Nice work figuring this out! happy to merge this as it is, but any chance you have a lead on how we can remove the dependency on `onChange` here?

Ideally, the form input should update the form value automatically, and should rerender without needing to rely on a state variable (this is what the Form.item `shouldUpdate` property is meant to be for, but it doesn't seem to work as expected).

An example of what I'm talking about is the `ProjectPayoutMods` file you edited. `onChange` manually updates the form value and a state variable, and the state variable is only used to ensure that the component re-renders on change

aeolianeth:
@peripheralist thanks for the review!

> Ideally, the form input should update the form value automatically, and should rerender without needing to rely on a state variable (this is what the Form.item shouldUpdate property is meant to be for, but it doesn't seem to work as expected).

As far as I can tell, `Form.Item` is intended be the direct parent of form fields. For example, this works:

```jsx
<Form.Item
  name="percent"
  label="Percent"
  rules={[{ required: true }]}
>
  <InputNumber
    formatter={(val?: string | number | undefined) => {
      let _val = val?.toString() ?? '0'
      const decimals = 2

      if (
        _val.includes('.') &&
        _val.split('.')[1].length > decimals
      ) {
        _val = parseFloat(_val).toFixed(decimals)
      }

      return `${_val ?? ''}%`
    }}
    parser={(val?: string) =>
      parseFloat(val?.replace('%', '') ?? '0')
    }
  />
  </Form.Item>
```

But, as soon as we wrap `InputNumber` in a `<span>`, it breaks:

```jsx
<Form.Item
  name="percent"
  label="Percent"
  rules={[{ required: true }]}
>
  <span>
    <InputNumber
      formatter={(val?: string | number | undefined) => {
        let _val = val?.toString() ?? '0'
        const decimals = 2

        if (
          _val.includes('.') &&
          _val.split('.')[1].length > decimals
        ) {
          _val = parseFloat(_val).toFixed(decimals)
        }

        return `${_val ?? ''}%`
      }}
      parser={(val?: string) =>
        parseFloat(val?.replace('%', '') ?? '0')
      }
    />
  </span>
</Form.Item>
```

So in this particular case, where we kinda can't have `NumberInput` be the direct child of `Form.Item`, I'm not sure we are able to remove our dependency on `onChange` 🤔 .

As for `shouldUpdate`, agreed it doesn't appear to be working as expected, but it's unclear to me why we need this. Is the intention to re-render the `NumberSlider` whenever another form field changes?

I only have 1 day of Antd experience though, so take my digging as a grain of salt 😊

peripheralist:
@aeolianeth Yep, that's my understanding too. Antd has lots of little nuanced rules, still figuring them out haha

In the `ProjectPayoutMods` form item, `shouldUpdate` is intended to allow the rendered percentage (inline with the input) to update when the form value updates. Although since we're using the `onChange` callback `shouldUpdate` becomes redundant (even if it worked as expected). Looks like it was added to `ProjectTicketMods` by mistake, doesn't serve any purpose there.

Can you go ahead and remove the `shouldUpdate` in both places? Then let's merge this in. We can work towards deprecating the `onChange` + state variable pattern later, but it works for now 🤷

aeolianeth:
> Can you go ahead and remove the shouldUpdate in both places? Then let's merge this in. We can work towards deprecating the onChange + state variable pattern later, but it works for now 🤷

@peripheralist sounds good to me!

`shouldUpdate` removed. Back to you 🏓

peripheralist (approved):

peripheralist:
nice, thanks! 🙏

Issue #175, opened 2021-10-29
[IDEA] Increase character limit of project description


peripheralist:
Project description is currently limited to 300 characters. Let's increase the limit to 1000 characters.

Displaying the project description on the project page should also be adapted to show more characters.

aeolianeth:
@peripheralist I'll take this one 🤚

PR #231, opened 2021-11-26
Increase project description limit to 1000 characters

aeolianeth:
Fixes https://github.com/jbx-protocol/juice-interface/issues/175

- Increase project description limit to 1000 characters.
- Improves layout of ProjectHeader for readability on small screens/mobile devices.

### Screenshots

| Before | After |
| ------ | ------ |
| Screen Shot 2021-11-26 at 10 14 15 pm | Screen Shot 2021-11-26 at 10 07 52 pm |
| Screen Shot 2021-11-26 at 10 14 46 pm | Screen Shot 2021-11-26 at 10 11 17 pm |
| Screen Shot 2021-11-26 at 10 15 07 pm | Screen Shot 2021-11-26 at 10 11 46 pm |

### With long descriptions

https://user-images.githubusercontent.com/94939382/143722568-06300d0d-b147-4c9f-a324-4de6896ba8ef.mp4

### DIFF
--- a/src/components/Dashboard/ProjectHeader.tsx
+++ b/src/components/Dashboard/ProjectHeader.tsx
@@ -73,18 +73,60 @@ export default function ProjectHeader() {
         </div>
 
         <div style={{ flex: 1 }}>
-          <h1
+          <div
             style={{
-              fontSize: '2.4rem',
-              lineHeight: '2.8rem',
-              margin: 0,
-              color: metadata?.name
-                ? colors.text.primary
-                : colors.text.placeholder,
+              display: 'flex',
+              justifyContent: 'space-between',
+              flexWrap: 'wrap',
+              alignItems: 'flex-start',
             }}
           >
-            {metadata?.name || 'Untitled project'}
-          </h1>
+            <h1
+              style={{
+                fontSize: '2.4rem',
+                lineHeight: '2.8rem',
+                margin: 0,
+                color: metadata?.name
+                  ? colors.text.primary
+                  : colors.text.placeholder,
+              }}
+            >
+              {metadata?.name || 'Untitled project'}
+            </h1>
+
+            {!isPreviewMode && (
+              <div
+                style={{
+                  display: 'flex',
+                  alignItems: 'center',
+                }}
+              >
+                <span
+                  style={{
+                    color: colors.text.tertiary,
+                    paddingRight: 10,
+                  }}
+                >
+                  ID: {projectId.toNumber()}
+                </span>
+
+                <div>
+                  <Button
+                    onClick={() => setToolDrawerVisible(true)}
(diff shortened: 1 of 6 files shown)


src/components/Dashboard/ProjectHeader.tsx
```diff
@@ -73,18 +73,60 @@ export default function ProjectHeader() {
         </div>
 
         <div style={{ flex: 1 }}>
-          <h1
+          <div
```
aeolianeth, line 76: Here, we're moving the `h1` and the project id under the same `div`. This way, we have more control over the project description layout

aeolianeth:
@peripheralist 👋 , mind taking a look at this one when you have a chance?

P.s. let me know how we should request reviews and who to ask 🙂

peripheralist:
@aeolianeth looking great, thanks for including the before/after pictures!

what do you think about also finding a way to show max 3-4 lines on the project page, with a "show more" option to expand it to show the rest? if a project does actually use a 1000 char description, it would be nice for it to not always be taking up so much space

aeolianeth:
@peripheralist totally agree! To keep the PR small, I'm happy to add this in a follow-up PR. WDYT?

peripheralist:
> @peripheralist totally agree! To keep the PR small, I'm happy to add this in a follow-up PR. WDYT?

Let's try to get this UI improvement included here. That way if anyone updates their description to be real long before we get another PR done, it'll look nice right off the bat 😄 

Looks like Antd includes a `Paragraph` component that supports this, I haven't used it before but it might only be a few lines of code to get a "read more" expand button in there https://ant.design/components/typography/

aeolianeth:
@peripheralist, totally, good call 🙏 

I've created a new component `ProjectDescription` to handle this (see **With long descriptions** section of PR description.

I had to update `RichNote` to use `span`s, but I checked all other uses of it and they look unaffected.

Let me know what you think!

peripheralist:
love it @aeolianeth ! this works great.

what do you think about moving the ProjectDescription component under shared/ as a more generic "Paragraph.tsx" component, where the maxLength could be passed in to `props`? I have a feeling we might wanna use this elsewhere 😊

also, i wouldn't mind leaving out the little arrow icon—i wonder if it draws more attention to the toggle button than it needs

aeolianeth:
@peripheralist 

> what do you think about moving the ProjectDescription component under shared/ as a more generic "Paragraph.tsx" component, where the maxLength could be passed in to props? I have a feeling we might wanna use this elsewhere 😊

Love this idea - done!

Here's a screenshot, (i didn't update the demo video in the description:)

Screen Shot 2021-11-29 at 3 41 58 pm

peripheralist:
yes! awesome work, thank you

peripheralist (approved):

Issue #189, opened 2021-11-13
[BUG] Should not show "handle already in use" warning when input value matches the project's current handle


peripheralist:
![image](https://user-images.githubusercontent.com/79433522/141600380-410354b9-b36d-4436-8749-5063e9839e4d.png)

jnoh:
Working on this here: https://github.com/jbx-protocol/juice-interface/pull/242

jnoh:
Addressed by https://github.com/jbx-protocol/juice-interface/pull/241

PR #241, opened 2021-11-29
Fix ProjectHandle input

aeolianeth:
Fixes #119, https://github.com/jbx-protocol/juice-interface/issues/189

When editing a payout, the project handle field wasn't populating automatically.

This PR fixes this by:
- Leveraging `Form.Item`'s automatic data binding correctly
- Adding a loading spinner to the project handle input

There's a couple of other unrelated fixes in here, too. I've made comments throughout the PR to highlight these 🙂 .

### Demo

https://user-images.githubusercontent.com/94939382/143811934-5942f732-3a38-43e7-91b3-81efb8f637f8.mp4

PR #242, opened 2021-11-30
Bugfix: should not show "handle already in use" warning when input value matches the project's current handle

jnoh:
Add initialValue to ProjectHandle to check if value is the same a project's current handle.

https://github.com/jbx-protocol/juice-interface/issues/189

### DIFF
--- a/src/components/shared/formItems/ProjectHandle.tsx
+++ b/src/components/shared/formItems/ProjectHandle.tsx
@@ -83,7 +83,7 @@ export default function ProjectHandle({
       
       return Promise.resolve()
     },
-    [handleExists, requireState],
+    [handleExists, initialValue, requireState],
   )
 
   let suffix: string | JSX.Element = ''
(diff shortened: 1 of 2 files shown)


aeolianeth:
@jnoh thanks for working on this, although it may be fixed with https://github.com/jbx-protocol/juice-interface/pull/241.

Will test now, just wanted to let you know ASAS 🙏

src/components/shared/formItems/ProjectHandle.tsx
```diff
@@ -71,13 +73,17 @@ export default function ProjectHandle({
 
   const checkHandle = useCallback(
     (rule: any, value: any) => {
+      // Input field is back at it's initial state
+      if (initialValue && initialValue === value)
+        return Promise.resolve()
```
jnoh, line 78: @aeolianeth just looked at your PR - you might need to add this little check, but your solution is in line  with and way more holistic than this one so I'll close this PR in favor of yours.

jnoh:
Closing this since @aeolianeth's pr will address this: https://github.com/jbx-protocol/juice-interface/pull/241

Issue #240, opened 2021-11-28
[BUG] Token info layout broken on mobile


peripheralist:
![8901CEAA-C4A0-418A-B73E-4C845869A6DC](https://user-images.githubusercontent.com/79433522/143729892-c1f18ce8-d885-4a86-853e-d49b2b671ded.png)

peripheralist:
fixed

PR #244, opened 2021-12-01
Fix mobile layout of Token Information section

torvusbug:
To fix the mobile layout of this section, I changed the layout of the antd component to always "horizontal" fixing much of the alignment issues.

I also made the "Holders" button outlined to match the "Manage" button directly below it.

Addresses: #240

### DIFF
--- a/src/components/Dashboard/Rewards.tsx
+++ b/src/components/Dashboard/Rewards.tsx
@@ -123,6 +123,10 @@ export default function Rewards({
     [],
   )
 
+  const labelStyle: CSSProperties = {
+    width: 128,
+  }
+
   return (
     <div>
       <Space direction="vertical" size="large">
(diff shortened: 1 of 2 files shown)


peripheralist:
Nice work! Good call restyling the Holders button.

What do you think of also left-aligning the token address? It may be clearer keeping content to the left and buttons to the right.

torvusbug:
I thought the same for mobile, so I'll give that a shot.

torvusbug:
@peripheralist I left aligned the address on that side by hardcoding the width of the labels.

I also added a new `useMobile` hook because I thought I'd right-align on desktop and left-align on mobile but didn't end up doing that. It's a pretty useful hook to keep around, I use almost the same thing in my day job all the time. I noticed a few spots where the current document's width is calculated manually (to determine mobile/desktop), rather than wrapped in something the browser can optimize (like media queries). The `useMobile` hook will also respond and update its state when resizing to/from mobile screen size.

peripheralist:
great stuff! thanks for adding that hook, should have been done a while ago. normally i'm hesitant to include anything preemptively that isn't implemented yet. but this will be getting used soon so, let's go ahead and merge this in so you don't have to fish it out later in a separate PR when someone else needs it.

peripheralist (approved):

PR #245, opened 2021-12-01
Fix mobile tooltips, including ETH addresses

torvusbug:
Setting "trigger" on the common use-cases for Ant's Tooltips to "click" and "hover" make mobile work very well. The addresses are still a little tricky to tap on mobile, but at least the full address and etherscan link _do_ appear now to mobile users.


peripheralist (approved):
Nice improvement! Didn't know about that `trigger` property.

PR #253, opened 2021-12-03
Next POC

jnoh:
An exploration of migrating to next so that we can get SSR and server-side niceties. 

Ignore Fleek builds


jnoh:
Going with cloudflare approach. See https://www.notion.so/juicebox/Tools-for-New-Frontends-3cea3153be4d46e7b0b78239cd6cc521

PR #254, opened 2021-12-04
Add clarity to modals `ok` button text

aeolianeth:
This PR changes the "okText" prop of most modals. This is an attempt to be as clear and explicit as possible about the action the user is taking.

I also ran `yarn lint:fix` to fix some left-over linting issues.


peripheralist (approved):
love it. thanks for looking out for little design improvements like this 👍

PR #260, opened 2021-12-07
Add infinite scroll loading to projects grid

torvusbug:
To solve for the ever-growing list of projects without compromising loading times due to querying all projects at once, implement a scrolling-based infinite loader.

I set the page size to 20 (which previously was erroneous) and load 20 more when scrolling within 200px of the bottom of the grid. I also added a backup "Load More" button in case either the browser doesn't support `IntersectionObserver` or something else went wrong.

### DIFF
--- a/src/components/Dashboard/index.tsx
+++ b/src/components/Dashboard/index.tsx
@@ -23,7 +23,7 @@ import { archivedProjectIds } from 'constants/archived-projects'
 
 import Loading from '../shared/Loading'
 import Project from './Project'
-import { useProjects } from '../../hooks/Projects'
+import { useProjectsQuery } from '../../hooks/Projects'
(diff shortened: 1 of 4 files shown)


peripheralist:
Stoked to play around with this! What do you think about pulling out the infinite scroll functionality to be extended to other parts of the app that show queried objects, like payment activity? (Though this may be more difficult for paged objects that are children of the scroll container)

torvusbug:
@peripheralist I think that's a good idea generally, but I'd wait to see how many of these types of sections we actually have before we abstract the current two. The two now have _just enough_ differences where I'd be hesitant to abstract them.

peripheralist (approved):
Sounds good, this looks great!

Issue #264, opened 2021-12-08
Handle `fundingCycle.metadata.payIsPaused`


peripheralist:
The new `FundingCycleMetadata` model in the V1.1 app will include a `payIsPaused` property.

**Reconfiguration**
The funding cycle reconfiguration form should include a boolean value for setting this to true or false.

**Pay form**
If `payIsPaused` is true for the current funding cycle, the Pay form should be disabled with a note that payments have been paused by the project owner.

Issue #263, opened 2021-12-08
Update contract call `printPreminedTickets` -> `printTickets`


peripheralist:
Ability to print tickets can be determined by `TerminalV1_1.ticketPrintingIsAllowed`

Issue #262, opened 2021-12-08
Update + rename FCMetadata interface -> FundingCycleMetadata


peripheralist:
For v1.1 app:

Rename `FCMetadata` -> `FundingCycleMetadata`

Update properties to:
![image](https://user-images.githubusercontent.com/79433522/145309298-0370061b-4c69-4468-931f-5dfdf1921f3d.png)

Funding cycle reconfiguration form should also support editing boolean values:
- `payIsPaused`
- `ticketPrintingIsAllowed`

Issue #261, opened 2021-12-08
Include form for migrating project from TerminalV1 to TerminalV1_1


peripheralist:
A button should be included in the current V1 app allowing projects to migrate to TerminalV1_1. This should include very clear information on what the migration entails.

Contract call docs: https://docs.juicebox.money/protocol-v1/terminalv1/write#migrate

PR #266, opened 2021-12-09
Add compatibility with new TerminalV1_1 contract

peripheralist:
This is a large update to make the app compatible with both TerminalV1 and TerminalV1_1 contracts.

## UX
- All new projects will be deployed using the new V1.1 terminal.
- Projects built on the V1 terminal can migrate to the new V1.1 terminal.
image

- V1.1 Projects can pause payments and manually set if tokens can be minted by reconfiguring funding cycles.
image

- All projects display a v1 or v1.1 tag on their dashboard, and on the project card
image
image

- Minting tickets has been moved to its own modal accessible through the tokens "manage" menu.
image

- Users can now search projects by handle from the projects page
image

- New subgraph data makes Participants data reliable (presumably) and allows showing user's staked balance alongside total (staked + unstaked) balance
image

## Code changes
- Updates FundingCycleMetadata property with new `payIsPaused` and `printingTicketsIsAllowed` properties
- Versioning support for terminal contracts and FundingCycleMetadata
- Allow setting `payIsPaused` and `printingTicketsIsAllowed` properties in the Create and Reconfigure forms, using new `RestrictedActionsForm`
- Includes `printingTicketsIsAllowed` in FundingCycleDetails element
- Move AdminFeePercent from ProjectContext to new useTerminalFee() hook
- Adds `terminal` property to ProjectContext including information on which terminal the project is using
- Instances of `useContractReader()` and `transactor()` use the terminal contract for the current project's terminal version
- New `ProjectSearch` subgraph entity
- Updated `Participant` subgraph entity

#261 #262 #263 #264


package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
-    "@jbx-protocol/contracts": "^0.0.2",
+    "@jbx-protocol/contracts-v1": "1.1.2",
```
odd-amphora, line 6: Don't forget to update this to 1.1.3.

peripheralist: done 👍

src/components/modals/MigrateV1Pt1Modal.tsx
```diff
@@ -0,0 +1,104 @@
...
+    <Modal
+      visible={visible}
+      onOk={migrate}
+      onCancel={onCancel}
+      okText="Migrate to V1.1"
+      okType="primary"
+      confirmLoading={loadingMigrate}
+      okButtonProps={{ disabled: needsBalance }}
+    >
+      <h2>Migrate to v1.1</h2>
```
aeolianeth, line 67: **suggestion**

```suggestion
      <h2>Migrate to Juicebox v1.1</h2>
```

src/components/modals/MigrateV1Pt1Modal.tsx
```diff
@@ -0,0 +1,104 @@
...
+      <p>
+        This project is currently using the Juicebox v1 terminal. New features
```
aeolianeth, line 69: **suggestion**: does the user typically know what `terminal` means? I wonder if it's okay to say something more general, like:

```suggestion
        This project is currently deployed on the Juicebox v1 smart contracts. New features
```

src/components/modals/MigrateV1Pt1Modal.tsx
```diff
@@ -0,0 +1,104 @@
...
+      },
+    )
+  }
+
+  return (
+    <Modal
+      visible={visible}
+      onOk={migrate}
+      onCancel={onCancel}
+      okText="Migrate to V1.1"
```
aeolianeth, line 62:
```suggestion
      okText="Migrate to v1.1"
```

src/components/modals/MigrateV1Pt1Modal.tsx
```diff
@@ -0,0 +1,104 @@
...
+        </a>
+      </p>
+
+      {needsBalance && (
+        <div>
+          <p>
+            <b>NOTE:</b> Projects cannot be migrated without a balance, and this
+            project currently has a balance of 0. To migrate this project, first
+            pay it or use the button below to deposit 1 gwei
+            (0.000000000000000001 ETH).
```
aeolianeth, line 93:
```suggestion
            <b>NOTE:</b> This project has a balance of 0. Projects cannot be migrated without a balance.
            To migrate this project, first
            pay it or use the button below to deposit 1 gwei
            (0.000000000000000001 ETH).
```

src/components/modals/PrintPreminedModal.tsx
```diff
@@ -0,0 +1,159 @@
...
+    setLoading(true)
+
+    await form.validateFields()
+
+    let terminalContract: Contract
+    let functionName: string
+    let args: any[]
+
+    switch (terminal.version) {
+      case '1':
```
aeolianeth, line 44: **suggestion (non-blocking)**: worth having shared constants for these?

src/components/modals/ReconfigureFCModal.tsx
```diff
@@ -131,10 +173,31 @@ export default function ReconfigureFCModal({
     ticketingForm.setFieldsValue({
       reserved: parseFloat(fromPerbicent(metadata.reservedRate)),
     })
-  }, [dispatch, fundingCycle, payoutMods, ticketMods, ticketingForm])
+
+    if (metadata.version === 1) {
```
aeolianeth, line 177: **question**: version in metadata is `int` (`1`), but previously I noticed it was a string (`'1'`). Is this correct?

src/components/shared/ProjectCard.tsx
```diff
@@ -72,8 +83,24 @@ export default function ProjectCard({
...
+              </span>
+              <span
+                style={{
+                  marginLeft: 10,
+                  color: colors.text.tertiary,
+                  fontSize: '0.7rem',
+                  fontWeight: 500,
+                }}
+              >
+                V{terminalVersion}
```
aeolianeth, line 98: **question**: should we use upper or lower-case v? (In the migrate modal, we're using lowercase)

src/components/shared/formItems/ProjectHandle/ProjectHandleInput.tsx
```diff
@@ -43,7 +43,6 @@ export function ProjectHandleInput({
   // initialValue can be a string (the human-readable handle)
   // or a project ID.
   useEffect(() => {
-    console.log('CHANGED', initialValue)
```
aeolianeth, line 46: **praise** 🙏

aeolianeth (commented):
@peripheralist incredible work here 👏. I left a few minor comments here and there, but overall, this PR is stellar.

One other question is whether we should be marking new strings for translation. In the interest of getting this shipped, I think it's fine to add these in a follow-up PR.

src/components/modals/ReconfigureFCModal.tsx
peripheralist: FundingCycleMetadata version is an `int` (maybe it used to be a string?) but you may be thinking of the Terminal contract version, which is a string `"1" | "1.1"`. My reasoning for those is that the FC metadata version is stored in the contracts as a number, whereas the terminal version is more like a name. Although we could treat 1 and 1.1 as numbers, there's nothing to prevent us from launching a `1.1.1`, or a `1.1-<string>`

src/components/shared/ProjectCard.tsx
peripheralist: the uppercase looked better on the project cards and in the project dashboard, though lowercase feels more appropriate somehow. if we had to pick one to unify, i think my vote is uppercase

src/components/modals/PrintPreminedModal.tsx
peripheralist: maybe, these are typed strings so an error will get thrown if you're using anything that isn't a version. i don't think it's as important here as it was with currency, where 0/1 does nothing to indicate ETH/USD

src/components/modals/MigrateV1Pt1Modal.tsx
peripheralist: ehh probably not. but it also may be misleading to say "v1 smart contracts", since it's really only the TerminalV1 contract being updated. i think i'd rather err on the side of clarity and encourage folks to ask "what's a terminal?"

peripheralist:
> One other question is whether we should be marking new strings for translation. In the interest of getting this shipped, I think it's fine to add these in a follow-up PR.

@aeolianeth great call. we should definitely be in the habit of adding new translations as we go, totally slipped my mind here. though since this diff is so big, i think it's best to break the rule here and minimize any last minute code changes to focus on review. i'll take responsibility for circling back to translations for these additions this week, unless someone else gets to it first

src/components/modals/PrintPreminedModal.tsx
aeolianeth: agreed. Given it's typed, should be fine 👍

src/components/shared/ProjectCard.tsx
aeolianeth: I don't have a strong opinion - uppercase SGTM 💯

src/components/modals/PrintPreminedModal.tsx
```diff
@@ -0,0 +1,159 @@
...
+    | { label: string; extra: string }
+    | undefined = useMemo(() => {
+    if (!terminal?.version) return
+
+    switch (terminal.version) {
+      case '1':
+        return {
+          label: 'Payment equivalent',
+          extra:
+            'The amount of tokens minted to the receiver will be calculated based on if they had paid this amount to the project in the current funding cycle.',
```
johnnyd-eth, line 90: **Question:** I'm a bit confused about this sentence. I may be understanding it wrong, but would it be better worded as:
'The amount of tokens minted to the receiver will be calculated based on the amount they paid to the project in the current funding cycle.' ?

src/components/modals/PrintPreminedModal.tsx
```diff
@@ -0,0 +1,159 @@
...
+      </div>
+
+      <Form layout="vertical" form={form}>
+        <Form.Item
+          label="Tokens receiver"
+          name="beneficary"
+          rules={[
+            {
+              required: true,
+              validator: (rule: any, value: any) => {
```
johnnyd-eth, line 123: No biggie, but in the future, I think we can get away with not passing these validator functions any arguments if they're not being used.

src/components/modals/PrintPreminedModal.tsx
```diff
@@ -0,0 +1,159 @@
...
+        <Form.Item label="Memo" name="memo">
+          <Input placeholder="Memo included on-chain (optional)" />
+        </Form.Item>
+        <Form.Item
+          name="preferUnstaked"
+          label="Mint as ERC-20"
+          valuePropName="checked"
+          extra={
+            erc20Issued
+              ? `Enabling this will mint ${tokenSymbol} ERC-20 tokens. Otherwise staked ${tokenSymbol} tokens will be minted, which can be claimed later as ERC-20 by the recever.`
```
johnnyd-eth, line 149: **Issue: ** Spelling mistake 'recever' -> 'receiver'

johnnyd-eth (commented):
Excellent PR @peripheralist. The key features here are essential additions. Just a few small comments from me. Aside from these, I would have also just mentioned the translations - I was planning to deliver a knockout blow to the remaining untranslated texts later this week, so would be happy to go over all these new ones as well.

src/components/modals/MigrateV1Pt1Modal.tsx
```diff
@@ -0,0 +1,104 @@
...
+        </a>
+      </p>
+
+      {needsBalance && (
+        <div>
+          <p>
+            <b>NOTE:</b> This project has a balance of 0. Projects cannot be migrated without a balance.
+            To migrate this project, first
+            pay it or use the button below to deposit 1 gwei
+            (0.000000000000000001 ETH).
```
johnnyd-eth, line 93: **Issue**: I think there are too many 0s here. Pretty sure it's just 0.000000001. And maybe we should show it as a power i.e. '10<sup>-9</sup> ETH', or both

aeolianeth: We could stringify 1gwei from `ethers`?

https://docs.ethers.io/v5/api/utils/constants/#constants-WeiPerEther

peripheralist: oh oops. the issue is actually that it says _gwei_ when it should say _wei_. this is intended to deposit the smallest possible amount. will fix

src/components/modals/PrintPreminedModal.tsx
peripheralist: good catch, will remove. probably copy and pasted this without thinking

src/components/modals/PrintPreminedModal.tsx
peripheralist: Lol it's a hard thing to word well. This is just terminalV1 and is more intuitive with v1.1.

the idea is to put in an amount that represents a hypothetical payment to the project. if you enter "5 ETH", the receiver will earn as many tokens as they would if they had just paid the project 5 ETH

PR #278, opened 2021-12-12
Fix: Toggling off Funding Cycle Duration now ensures the Funding Cycle is set to recurring

torvusbug:
Comment left in code: If toggling off Funding Cycle Duration when set to "one-time", revert it to "recurring"

### DIFF
--- a/src/components/shared/formItems/ProjectDuration.tsx
+++ b/src/components/shared/formItems/ProjectDuration.tsx
@@ -35,6 +35,11 @@ export default function ProjectDuration({
               onChange={checked => {
                 setShowDurationInput(checked)
                 onValueChange(checked ? '30' : '0')
+
+                // If toggling off Funding Cycle Duration when set to "one-time", revert it to "recurring"
+                if (!isRecurring && !checked) {
+                  onToggleRecurring()
+                }
               }}
             />
             <label>Set a funding cycle duration</label>


jnoh:
Adding additional content: UI issue caused a misconfiguration of the project where by the duration property was turned off and the recurring property was turned off locking the project owner from making any mods to project config. 

See [triage](https://discord.com/channels/775859454780244028/864240636277293106/919604404896796782) and [bug report](https://discord.com/channels/775859454780244028/866040669712678942/919608900943220747) threads.

src/components/shared/formItems/ProjectDuration.tsx
```diff
@@ -35,6 +35,11 @@ export default function ProjectDuration({
               onChange={checked => {
                 setShowDurationInput(checked)
                 onValueChange(checked ? '30' : '0')
+
+                // If toggling off Funding Cycle Duration when set to "one-time", revert it to "recurring"
+                if (!isRecurring && !checked) {
+                  onToggleRecurring()
```
jnoh, line 41: Noblocking nit: It might be clearer what's happening here if we switched `onToggleRecurring` to `setIsRecurring` and explicitly set the value. We can do toggle behavior in other places with `setIsRecurring(!isReucrring)`.

jnoh (approved):
One nit. Otherwise looks good.

peripheralist:
agree with the nit—there's probably a lot of room for improvement here. for now let's push this fix and come back to other changes in another PR

Issue #277, opened 2021-12-12
Add test for `src/components/App.test.tsx`


aeolianeth:
Let's write a basic test that asserts the App renders properly.

aeolianeth:
Closing this issue for now. The spec will be removed in https://github.com/jbx-protocol/juice-interface/pull/283.

PR #282, opened 2021-12-12
PoC: make `App.test.tsx` test relevant

aeolianeth:
This PR updates the `App.test.tsx` spec to check the App component renders as expected.

Previously, it was doing nothing.

** ❓ Question: should we merge this PR, or remove the file?**

Adding the spec is great but it has the potential to slow us down in some circumstances. I'd like to hear the opinions of the team before pursuing this further.

(note, if you `yarn test` locally the test passes, but there are console logs. These would be resolved if we were to continue with this PR.)

Fixes https://github.com/jbx-protocol/juice-interface/issues/277

cc @torvusbug @jnoh @peripheralist

### DIFF
--- a/src/components/App.test.tsx
+++ b/src/components/App.test.tsx
@@ -1,9 +1,48 @@
-import { render, screen } from '@testing-library/react'
+import { render, screen, act } from '@testing-library/react'
+import { Provider } from 'react-redux'
 
 import App from './App'
+import ProvideReactQuery from '../ReactQuery'
+import store from '../redux/store'
 
-test('renders learn react link', () => {
-  render(<App />)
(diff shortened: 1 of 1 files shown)


torvusbug:
TBH, I think we should remove it until we're willing to commit to writing tests consistently. I don't want others thinking they can build off this incomplete setup.

peripheralist:
i agree with @torvusbug . i'd love to start considering unit tests once we have the bandwidth to keep them updated without slowing down dev progress too much. but until then I'd rather ignore or remove the tests altogether.

aeolianeth:
Sounds great 👍 . Closing this PR, I'll open another one to remove it

Issue #136, opened 2021-09-13
Restrict using the same address multiple times for reserve tokens


odd-amphora:
I feel like in the frontend we should probably try to prevent this from happening?
<img width="591" alt="Screen Shot 2021-09-13 at 3 03 48 PM" src="https://user-images.githubusercontent.com/88144767/133141788-e46c5975-a5e3-4724-a469-dfa68030aa9e.png">

peripheralist:
yeah, in this case i think zeugh is the project owner, and is receiving the leftover 1% by default.

we can't restrict setting allocations to the project owner so this is a little tricky, but i think the UI also still allows duplicate allocations that should maybe be disallowed anyway.

aeolianeth:
> but i think the UI also still allows duplicate allocations that should maybe be disallowed anyway.

@johnnyd-eth 👋 did you end up resolving this issue with https://github.com/jbx-protocol/juice-interface/pull/293?

PR #293, opened 2021-12-18
Various fixes to reserved token form

johnnyd-eth:
- Modal displays edit-related text when editing
- Prevent null address being added as beneficiary
- Prevent duplicate beneficiaries being added

Went to fix issue #136 and realised none of the current validation (e.g. null address) was being applied. Fixed this and added the new one, but then was unable to edit existing beneficiaries (duplicate address validation breaking) - so added some extra logic in the validation. Then made the modal text edit-related for when a beneficiary is being edited. 

Video shows attempting to add a null address, edit-related modal text when editing, and then attempting to add the same address. There are preexisting issues with the error message styling which I intend to fix in a follow-up PR.
 
https://user-images.githubusercontent.com/96150256/146656758-774cfab1-41b1-45f2-b56e-c0a6dfe4c1d4.mp4

## Acceptance checklist

- [x] I have evaluated the
      [Approval Guidelines](../../CONTRIBUTING.md#approval-guidelines) for this
      PR.

### DIFF
--- a/src/components/shared/formItems/ProjectTicketMods.tsx
+++ b/src/components/shared/formItems/ProjectTicketMods.tsx
@@ -16,6 +16,10 @@ import FormattedAddress from '../FormattedAddress'
 import NumberSlider from '../inputs/NumberSlider'
 import { FormItemExt } from './formItemExt'
 
+const MODAL_MODE_EDIT = 'Edit'
+const MODAL_MODE_ADD = 'Add'
+type MODAL_MODE = 'Add' | 'Edit'
+
 export default function ProjectTicketMods({
   name,
   lockedMods,
@@ -32,7 +36,8 @@ export default function ProjectTicketMods({
     percent: number
     lockedUntil: moment.Moment
   }>()
-  const [editingModIndex, setEditingModIndex] = useState<number>()
+  const [editingModIndex, setEditingModIndex] = useState<number>() // index of the mod currently being edited (edit modal open)
+  const [modalMode, setModalMode] = useState<MODAL_MODE>() //either 'Add', 'Edit' or undefined
   const { owner } = useContext(ProjectContext)
 
   const {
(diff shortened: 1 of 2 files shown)


src/components/shared/formItems/ProjectTicketMods.tsx
```diff
@@ -200,10 +206,33 @@ export default function ProjectTicketMods({
...
+      modalMode === 'Edit' &&
+      address === mods[editingModIndex ?? 0]?.beneficiary
+    )
+      return Promise.resolve()
+    // if user edits an (already approved) address and doesn't change it, we accept
+    else if (!address || !utils.isAddress(address))
+      return Promise.reject('Address is required')
+    else if (address === constants.AddressZero)
+      return Promise.reject('Cannot use zero address.')
+    else if (mods.filter(mod => mod.beneficiary === address).length > 0)
```
peripheralist, line 231: nit pick: is mods.some() more performant here?

src/components/shared/formItems/ProjectTicketMods.tsx
```diff
@@ -16,6 +16,10 @@ import FormattedAddress from '../FormattedAddress'
 import NumberSlider from '../inputs/NumberSlider'
 import { FormItemExt } from './formItemExt'
 
+const MODAL_MODE_EDIT = 'Edit'
+const MODAL_MODE_ADD = 'Add'
+type MODAL_MODE = 'Add' | 'Edit'
```
peripheralist, line 21: since a type error would be thrown using any value other than "add" or "edit", I think we can get by without defining these values as variables

also nit pick: the app uses TitleCase for type declarations, so `ModalMode` instead of `MODAL_MODE` please 😊

peripheralist (commented):
Added a few nit picking comments, great work overall! Thanks for jumping on this

johnnyd-eth:
@peripheralist thanks for the review! I addressed your feedback

peripheralist (approved):
awesome!

PR #298, opened 2021-12-21
Fix error message styling on EthAddress inputs

johnnyd-eth:

Fix error message styling on EthAddress inputs. Previously, they were all over the place depending on two conditions: A, whether 'extra' prop is passed EthAddress, and B, whether the 'addressForENS' and its check symbol were present. I've accounted for these conditions by: A, giving FormItem.EthAddress a new class when 'extra' is passed to it and then making a separate CSS declaration for this class. And for B, moving the position of the 'addressForENS' in the code so the error message behaves the same whether it is present or not. 

OLD:
Screen Shot 2021-12-21 at 1 00 06 pm
Screen Shot 2021-12-21 at 1 00 16 pm
Screen Shot 2021-12-21 at 1 00 52 pm

NEW:
Screen Shot 2021-12-21 at 12 57 20 pm
Screen Shot 2021-12-21 at 12 55 51 pm
Screen Shot 2021-12-21 at 12 56 09 pm

## Acceptance checklist

- [x] I have evaluted the
      [Approval Guidelines](../../CONTRIBUTING.md#approval-guidelines) for this
      PR.

### DIFF
--- a/src/components/shared/formItems/EthAddress.tsx
+++ b/src/components/shared/formItems/EthAddress.tsx
@@ -74,9 +74,16 @@ export default function EthAddress({
 
     readENSName()
   }, [defaultValue])
-
+  console.log('formItemProps?.extra: ', formItemProps?.extra)
   return (
-    <Form.Item {...formItemProps}>
+    <Form.Item
+      {...formItemProps}
+      className={
+        formItemProps?.extra
+          ? 'ant-form-item-control-input-extra-and-error'
+          : ''
+      } // if we pass form.item an 'extra', need to set different padding for error message
+    >
       <Input
         placeholder={'juicebox.eth / ' + constants.AddressZero}
         type="string"
@@ -84,6 +91,11 @@ export default function EthAddress({
         onChange={e => onInputChange(e.target.value)}
         value={displayValue}
       />
+      {addressForENS?.length ? (
+        <div style={{ fontSize: '0.7rem', color: colors.text.secondary }}>
+          <CheckCircleFilled /> {addressForENS}
+        </div>
+      ) : null}
       <Form.Item
         name={name}
         style={{ height: 0, maxHeight: 0, margin: 0 }}
@@ -92,11 +104,6 @@ export default function EthAddress({
         {/* Hidden input allows for address value to be used in form, while visible input can display ENS name */}
         <Input hidden type="string" autoComplete="off" />
       </Form.Item>
-      {addressForENS?.length ? (
-        <div style={{ fontSize: '0.7rem', color: colors.text.secondary }}>
-          <CheckCircleFilled /> {addressForENS}
-        </div>
-      ) : null}
     </Form.Item>
   )
 }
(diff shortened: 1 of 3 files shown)


src/components/shared/formItems/EthAddress.tsx
```diff
@@ -74,16 +74,28 @@ export default function EthAddress({
...
   }, [defaultValue])
-
+  console.log('formItemProps?.extra: ', formItemProps?.extra)
   return (
-    <Form.Item {...formItemProps}>
+    <Form.Item
+      {...formItemProps}
+      className={
+        formItemProps?.extra
+          ? 'ant-form-item-control-input-extra-and-error'
```
peripheralist, line 83: i may be missing something with your CSS solution here, but should we just override this for all cases? as in, could we apply the CSS for `.ant-form-item-control-input-extra-and-error` to all form item elements?

src/components/shared/formItems/EthAddress.tsx
```diff
@@ -74,16 +74,28 @@ export default function EthAddress({
 
     readENSName()
   }, [defaultValue])
-
+  console.log('formItemProps?.extra: ', formItemProps?.extra)
```
peripheralist, line 77: 🥲

src/components/shared/formItems/ProjectPayoutMods.tsx
```diff
@@ -405,6 +405,7 @@ export default function ProjectPayoutMods({
               defaultValue={form.getFieldValue('beneficiary')}
               formItemProps={{
                 label: 'Address',
+                // extra: 'The address that should receive the tokens.',
```
peripheralist, line 408: is there a reason we're leaving this in + commented out, or should we get rid of it?

peripheralist:
nice catch with this bug! thanks for the fix. left some comments

src/components/shared/formItems/EthAddress.tsx
johnnyd-eth: @peripheralist oddly the FormItem places the error message in the same spot regardless of whether an 'extra' is present or not. So in cases where no 'extra' is present (such as 'Add a payout'), and we give it the class '.ant-form-item-control-input-extra-and-error', the error message appears far too low (see attached screenshot). So different CSS is required when we choose not to use the 'extra' prop. 

'Add a payout' (no extra) with the same CSS as other cases with an extra 
Screen Shot 2021-12-22 at 6 05 51 pm

src/components/shared/formItems/ProjectPayoutMods.tsx
johnnyd-eth: My bad there @peripheralist I did make a commit removing these but forgot to push!

src/components/shared/formItems/EthAddress.tsx
peripheralist: got it, thanks for clarifying. add this to the list of Antd funniness we should try to build around (cc @torvusbug)

peripheralist (approved):
thanks @johnnyd-eth !

PR #301, opened 2021-12-24
Optimize Distribution forms

johnnyd-eth:

This PR optimizes all Distribution forms, solving all points specified in section 2.1 and 3.1 of  ["Optimize the app's use of reactive forms"](https://www.notion.so/Optimize-the-app-s-use-of-reactive-forms-626b1ff4e1864d55973c6fe37930877a ). Was intending to cover section 3.1 (editing payouts from project page) in another PR, but there was so much dependant code I just included it in this. 

- Prevents submitting 0% payout in individual payouts
- Prevent submitting no address in individual payout
- Prevent duplicate payout addresses in individual payouts
- Prevent submitting >100% total payouts
- Show edit-related text when editing an individual payout (this was working from Project->Edit payouts, but not from create project)

optimize-distribution-form

## Acceptance checklist

### DIFF
--- a/src/components/shared/PayoutModsList.tsx
+++ b/src/components/shared/PayoutModsList.tsx
@@ -216,7 +216,7 @@ export default function PayoutModsList({
                   validator: () => {
                     if (totalPercentage > 100)
                       return Promise.reject(
-                        'Percentages must add up to less than 100%',
+                        'Percentages must add up to 100% or less',
                       )
 
                     return Promise.resolve()
--- a/src/components/shared/PayoutModsList.tsx
+++ b/src/components/shared/PayoutModsList.tsx
@@ -13,6 +13,8 @@ import { useContext, useLayoutEffect, useMemo, useState } from 'react'
 import { formatWad, fromPermyriad, fromWad } from 'utils/formatNumber'
 import { amountSubFee } from 'utils/math'
 
+import { getTotalPercentage } from './FormHelpers'
+
 import ProjectPayoutMods from './formItems/ProjectPayoutMods'
 
 export default function PayoutModsList({
@@ -96,11 +98,7 @@ export default function PayoutModsList({
 
   const hasEditPermission = useHasPermission(OperatorPermission.SetPayoutMods)
 
-  const totalPercentage =
-    editingMods?.reduce(
-      (acc, curr) => acc + parseFloat(fromPermyriad(curr.percent ?? '0')),
-      0,
-    ) ?? 0
+  const totalPercentage = getTotalPercentage(editingMods)
 
   if (!fundingCycle) return null
 
--- a/src/components/shared/PayoutModsList.tsx
+++ b/src/components/shared/PayoutModsList.tsx
@@ -96,7 +96,7 @@ export default function PayoutModsList({
 
   const hasEditPermission = useHasPermission(OperatorPermission.SetPayoutMods)
 
-  const total_percentage =
+  const totalPercentage =
     editingMods?.reduce(
       (acc, curr) => acc + parseFloat(fromPermyriad(curr.percent ?? '0')),
       0,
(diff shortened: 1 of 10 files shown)


mejango:
nice! if you're working on this flow, there are a few more bugs I've noticed, and we could maybe make resolving percents from inputted $$ amounts better.

johnnyd-eth:
> 

Sounds good @mejango. Let's arrange a chat next week

src/components/shared/PayoutModsList.tsx
```diff
@@ -91,6 +96,12 @@ export default function PayoutModsList({
 
   const hasEditPermission = useHasPermission(OperatorPermission.SetPayoutMods)
 
+  const total_percentage =
```
aeolianeth, line 99: **question (non-blocking)**: we're duplicating this reducer code from `PayModsForm`. Does it make sense to create a shared helper function somewhere and use it here and in `PayModsForm`?

aeolianeth (changes requested):
Awesome PR @johnnyd-eth, these improvements are important for sure.

I left a few minor comments here and there for you. Let me know what you think!

johnnyd-eth:
> Awesome PR @johnnyd-eth, these improvements are important for sure.
> 
> I left a few minor comments here and there for you. Let me know what you think!

Thanks for your feedback @aeolianeth. I've addressed your suggestions:
- used more specific types in all places you mentioned and others I could find 
- created helper file  (FormHelper) to put repeated code in a single place

johnnyd-eth:
@peripheralist @aeolianeth Do not merge this yet! Unfortunately, I've just noticed that my changes to the NumberSlider have caused the slider in the Reserved Tokens form to not update its corresponding input field properly. I couldn't find an obvious fix but will get onto it tomorrow.

johnnyd-eth:
Just resolved the issue I mentioned with other instances of the NumberSlider such as the reserved tokens form.

src/components/shared/PayoutModsList.tsx
```diff
@@ -167,40 +176,72 @@ export default function PayoutModsList({
...
-          okText="Save payouts"
-          onOk={() => setMods()}
-          onCancel={() => {
-            setEditingMods(mods)
-            setModalVisible(false)
+        <Form
+          form={form}
+          layout="vertical"
+          onKeyDown={e => {
+            if (e.key === 'Enter') setMods()
```
peripheralist, line 183: nice!

peripheralist:
@johnnyd-eth this is solid all around!

two small suggestions to make errors a bit clearer. what do you think about:

"Percentages must add up to less than 100%" -> "Percentages must add up to 100% or less"
"Address already in use." -> "A payout for this address already exists"

It would be awesome to modify this PR to add the same functionality to reserved token distributions as well, since the pattern should be pretty identical. Though I'm fine with doing that as a separate PR

@mejango if you think some other changes would be relevant here could you drop a list for reference? i'm down to add in more here if @johnnyd-eth wants to, but also happy to land these changes so far and open a separate PR for anything else to keep things well scoped.

johnnyd-eth:
> A payout for this address already exists

@peripheralist great ideas there, that wording makes much more sense. 

As for the reserved token stuff, I see you've now reviewed the other PR where I added checks for reserved token receivers. Is this all you were thinking needs to be done there? I was considering adding a 0% check on the overall reserved token allocation, but I figured we might want to just let users press save with that unset. Added screenshot below for clarification.


Screen Shot 2021-12-29 at 7 00 57 am


And yes happy to work keep working on it if you, @mejango, or anyone else can think of more improvements.

peripheralist (approved):
awesome @johnnyd-eth ! that other PR covers what I had in mind, forgot you had it open when I wrote that comment.

peripheralist:
@aeolianeth looks like your requested changes here are still open and blocking merge, was there something else in your comments that wasn't addressed?

aeolianeth (approved):
@johnnyd-eth @peripheralist apologies for blocking this merge. Happy to approve 🚀

Issue #182, opened 2021-11-10
[IDEA] Persist app state in localStorage to prevent data from being lost on refresh during edit/create flow


peripheralist:

peripheralist:
done!

PR #302, opened 2021-12-24
Save and resume the Design Project Form between sessions

torvusbug:

After making changes to the project form, the state will be saved to Local Storage. If you don't come back to the form within 7 days, the state will be erased. I like to do this as a safety measure in case the Redux state is sufficiently changed at some point and people could have an invalid state.

I also updated the Redux usage to take advantage of Immer

Linking #182 

https://user-images.githubusercontent.com/4457995/147359163-6ccbe6b8-6c0c-47b3-86d7-3c362a629957.mp4

## Acceptance checklist

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2677,12 +2677,12 @@
     pocket-js-core "0.0.3"
     web3-provider-engine "16.0.1"
 
-"@reduxjs/toolkit@^1.5.0":
-  version "1.6.1"
-  resolved "https://registry.yarnpkg.com/@reduxjs/toolkit/-/toolkit-1.6.1.tgz#7bc83b47352a663bf28db01e79d17ba54b98ade9"
-  integrity sha512-pa3nqclCJaZPAyBhruQtiRwtTjottRrVJqziVZcWzI73i6L3miLTtUyWfauwv08HWtiXLx1xGyGt+yLFfW/d0A==
+"@reduxjs/toolkit@^1.6.2":
+  version "1.6.2"
+  resolved "https://registry.yarnpkg.com/@reduxjs/toolkit/-/toolkit-1.6.2.tgz#2f2b5365df77dd6697da28fdf44f33501ed9ba37"
+  integrity sha512-HbfI/hOVrAcMGAYsMWxw3UJyIoAS9JTdwddsjlr5w3S50tXhWb+EMyhIw+IAvCVCLETkzdjgH91RjDSYZekVBA==
   dependencies:
-    immer "^9.0.1"
+    immer "^9.0.6"
     redux "^4.1.0"
     redux-thunk "^2.3.0"
     reselect "^4.0.0"
@@ -10306,7 +10306,7 @@ immer@8.0.1:
   resolved "https://registry.yarnpkg.com/immer/-/immer-8.0.1.tgz#9c73db683e2b3975c424fb0572af5889877ae656"
   integrity sha512-aqXhGP7//Gui2+UrEtvxZxSquQVXTpZ7KDxfCcKAF3Vysvw0CViVaW9RZ1j1xlIYqaaaipBoqdqeibkc18PNvA==
 
-immer@^9.0.1:
+immer@^9.0.6:
   version "9.0.6"
   resolved "https://registry.yarnpkg.com/immer/-/immer-9.0.6.tgz#7a96bf2674d06c8143e327cbf73539388ddf1a73"
   integrity sha512-G95ivKpy+EvVAnAab4fVa4YGYn24J1SpEktnJX7JJ45Bd7xqME/SCplFzYFmTbrkwZbQ4xJK1xMTUYBkN6pWsQ==
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
     "@jbx-protocol/contracts-v1": "^1.1.5",
     "@lingui/detect-locale": "^3.13.0",
     "@lingui/react": "^3.13.0",
-    "@reduxjs/toolkit": "^1.5.0",
+    "@reduxjs/toolkit": "^1.6.2",
     "@testing-library/jest-dom": "^5.11.4",
     "@testing-library/react": "^11.1.0",
     "@testing-library/user-event": "^12.1.10",
--- a/src/components/Create/index.tsx
+++ b/src/components/Create/index.tsx
@@ -24,14 +24,7 @@ import { FundingCycleMetadata } from 'models/funding-cycle-metadata'
 import { FCProperties } from 'models/funding-cycle-properties'
 import { PayoutMod, TicketMod } from 'models/mods'
 import { TerminalVersion } from 'models/terminal-version'
-import {
-  useCallback,
-  useContext,
-  useEffect,
-  useLayoutEffect,
-  useMemo,
-  useState,
-} from 'react'
+import { useCallback, useContext, useEffect, useMemo, useState } from 'react'
 import { editingProjectActions } from 'redux/slices/editingProject'
 import { fromPerbicent, fromPermille, fromWad } from 'utils/formatNumber'
 import {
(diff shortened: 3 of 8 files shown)


peripheralist:
awesome, that reducer file is so much cleaner.

wdyt about adding a version to the state? definitely an edge case, but if we were to change the editingProject state model (like we're doing with v1.1 with the new fundingCycleMetadata properties) while a user had the previous model cached, it could create some issues

aeolianeth:
Awesome stuff @torvusbug! Been craving this feature. Approving from my perspective, but would echo what Peri mentioned in https://github.com/jbx-protocol/juice-interface/pull/302#issuecomment-1001701006 re: versioning.

aeolianeth (approved):

torvusbug:
That's a good idea, I'll implement that and remove the datetime check.

torvusbug:
@peripheralist @aeolianeth Updated with the version check.

peripheralist:
awesome! one last thing @torvusbug, can we also add a step here https://github.com/torvusbug/juice-interface/blob/72ed8a9d8280b378e6faaa8f13647b5359e0c1dc/src/components/Create/index.tsx#L233 that resets the state once a project has been completed?

torvusbug:
@peripheralist Good idea, also is it worth adding a full Reset button somewhere?

peripheralist:
> @peripheralist Good idea, also is it worth adding a full Reset button somewhere?

That's also a good idea. If any changes have been made to the initial EditingProject state (maybe we could add an `editingProject.isChanged` property for convenience?) we could show a button below the "Deploy" button saying "Start over". In that case we should also show a modal when clicked saying "All changes you've made will be lost, Confirm/Cancel". Wdyt?

peripheralist:
something else to think about:

the Reconfigure modal also makes use of the editingProject state. if the user has an in-progress project state saved, that shouldn't interfere with using the Reconfigure modal. not too sure what the best way is to make that work

torvusbug:
@peripheralist I came up with a wacky solution for the Reconfigure Funding modal issue...

I created a new component that wraps the modal trigger button and the modal itself that wraps the Reconfigure Funding modal with a local version of the Redux state to avoid any conflicts with persistence. The store is initialized only when clicking the button to open the modal.

It works as far as I can tell.

(I also added a "Start Over" button that asks for confirmation first)

peripheralist:
haha i love the creativity here. thanks @torvusbug , i'll dive into your latest this weekend!

peripheralist:
i'm wondering if this would be a good time to consider a redesign of the `EditingProject` state entirely, and maybe allow solving this in a more straightforward way?

that redux state was introduced early in this codebase to try to manage re-rendering, but i've never felt like it was very elegantly designed. any thoughts @aeolianeth @johnnyd-eth @jnoh ?

johnnyd-eth:
> i'm wondering if this would be a good time to consider a redesign of the `EditingProject` state entirely, and maybe allow solving this in a more straightforward way?
> 
> that redux state was introduced early in this codebase to try to manage re-rendering, but i've never felt like it was very elegantly designed. any thoughts @aeolianeth @johnnyd-eth @jnoh ?

I see the problem but can't think of an elegant solution. All I can think is making a redux state for each instance we edit/create a project?

torvusbug:
Eh, Redux isn't the problem tbh. If it's not Redux, it'd be some other state management, like React Context with a `useReducer` passed as `value={{ state, dispatch }}` or some form manager like Fielder or Formik.

I see two problems: 

1. Form reusability across disparate parts of the experience, where each use case has a potentially different context. In this case, sub-forms are not abstracted away from their containing models. This is often very okay to do, unless...
2. Using global state management in a reusable form. When I use Redux for UX, I try to very carefully limit the entry points into that experience or use `meta` properties on objects to sort-of route the action into the right place (like a keying strategy).

I've used Formik to solve a _similar_ type of problem in the past, but it requires more "all-in" on Formik than I'd like, but this is just an example of an abstraction:

1. Create your top-level form, let's say it's the entire Editing Project form. It's a component that wraps Formik's initialization, submission, validation, React Context, and the `<form>`, ie:
```tsx
const EditingProjectForm: React.FC<Props> = () => {
  return (
    <Formik
      initialValues={{ meta: { name: "" } }}
      onSubmit={(values) => {}}
      validationSchema={schema}
    >
      {/* Formik's special Form, handles onSubmit */}
      <Form>{children}</Form>
    </Formik>
  )
}
```

2. Wrap base design system elements in a Formik-compatible one:
```tsx
interface Props {
  name: string;
}

const FormikFormattedNumberInput: React.FC<Props> = ({ name, ...props }) => {
  const [field, meta, helpers] = useField(name)
  return (
    <FormattedNumberInput
      name={name}
      {...props}
      errorText={meta.error}
      value={field.value}
      onChange={field.onChange}
    />
  )
}
```

3. Then create separate form field components that can wrap business logic and design:
```tsx
const FundingCycleFormFields: React.FC<Props> = () => {
  return (
    <Column>
      <FormikFormattedNumberInput
        label="Duration"
        name="fundingCycle.duration"
      />
      <FormikFormattedNumberInput label="Target" name="fundingCycle.target" />
    </Column>
  )
}
```

4. You can reuse the `FundingCycleFormFields` within a _different_ top-level Formik Form that may have different goals but requires the same elements. I use this pattern sometimes where a form has the same fields for creation _and_ editing/updating a data model.

This is just an example and some thoughts, but overall on this, I think initializing a new Redux `<Provider>` is _okay_ to do, albeit tech debt. The UX wins from saving/resuming projects may be worth it at the moment, but happy to be proven wrong.

peripheralist:
thanks @torvusbug I'm not familiar with Formik so that's a helpful start.

IMO saving a project will be nice to have but not critical. it shouldn't take someone more than a couple minutes to define a project start to finish, and starting over isn't the worst thing (to be honest, i'd rather optimize for folks being triple-sure of the decisions they're making when creating a project, than for easy UX)

still, the added tech debt doesn't seem very significant, and i'm down to move forward with this if everyone supports it

yarn.lock
```diff
@@ -2475,12 +2475,12 @@
     pocket-js-core "0.0.3"
     web3-provider-engine "16.0.1"
 
-"@reduxjs/toolkit@^1.5.0":
-  version "1.6.1"
-  resolved "https://registry.yarnpkg.com/@reduxjs/toolkit/-/toolkit-1.6.1.tgz#7bc83b47352a663bf28db01e79d17ba54b98ade9"
-  integrity sha512-pa3nqclCJaZPAyBhruQtiRwtTjottRrVJqziVZcWzI73i6L3miLTtUyWfauwv08HWtiXLx1xGyGt+yLFfW/d0A==
+"@reduxjs/toolkit@^1.6.2":
```
aeolianeth, line 2478: **question** was this upgrade necessary?

aeolianeth (approved):
@torvusbug I think the local copy of redux actually ends up being elegant and, strangely, makes this _more_ maintainable.

I've used Formik before and it's solid, but given the amount of forms in Juicebox, I'd be hesitant to adopt it without some more though. Definitely not ruling it out though - it was pretty amazing in my experience.

Left one question about the redux toolkit dep upgrade, but other than that, happy to approve 🚀

aeolianeth:
> still, the added tech debt doesn't seem very significant, and i'm down to move forward with this if everyone supports it

@peripheralist I'm supportive of this change. It ends up being a fairly [2-way door decision](https://shit.management/one-way-and-two-way-door-decisions/) - given @torvusbug's architectural choice here, we can always bail on the localstorage sync if there are issues.

One thing that might be nice here in the interest of iteration is to split this PR up into two - 1 for refactoring the form to use the local Redux state, and 1 to add localstorage. This way, we have a chance to identify bugs in the first PR, and can also easily revert the second PR if we decide to remove this functionality.

Just my 2 gwei - happy to proceed with this PR as-is :)

yarn.lock
torvusbug: Not necessary, but I try to keep this package up-to-date when I mess around in Redux-land.

peripheralist:
let's try it as is! thanks @torvusbug , and thanks for the input @aeolianeth

Issue #258, opened 2021-12-06
[IDEA] Disclose a token's AMM price(s)


aeolianeth:
How might we disclose a Token’s AMM price(s) on the dashboard so users can make the most informed decisions when considering a contribution? Perhaps this is something addable in a project’s metadata?

Raised in: https://discord.com/channels/775859454780244028/786985806396784690/917154545560330340

peripheralist:
docs for fetching pair price on Uniswap: https://docs.uniswap.org/sdk/guides/fetching-prices

for pair price on Sushi, looks like we can do `SushiV2Factory.getPair()` https://dev.sushi.com/sushiswap/contracts

i'd be down to support both exchanges, and show one or both prices depending on if a pair exists on that exchange

aeolianeth:
Sweet thanks for the links. I'm keen to take this one 👍

peripheralist:
hey @aeolianeth, i have a WIP prototype here https://github.com/jbx-protocol/juice-interface/pull/267 for showing redeem + pay as tabbed UI elements in the same section. But to do it well there needs to be some pretty significant reorganization of components, so to avoid messiness with merge conflicts I'm delaying finishing this until the v1.1 goes live. But feel free to build off this branch! I think it'd be great to solve both of these things together.

aeolianeth:
> hey @aeolianeth, i have a WIP prototype here #267 for showing redeem + pay as tabbed UI elements in the same section. But to do it well there needs to be some pretty significant reorganization of components, so to avoid messiness with merge conflicts I'm delaying finishing this until the v1.1 goes live. But feel free to build off this branch! I think it'd be great to solve both of these things together.

Oh nice @peripheralist, looks great. Hope you don't mind me dropping some screenshots in #267's description for my future reference https://github.com/jbx-protocol/juice-interface/pull/267#issue-798403194.

aeolianeth:
I think the main challenge here is the UX. As @mejango has mentioned, we don't want to blatantly advertise a particular AMM; our goal is simply to reduce information asymmetry. I'm hoping to have some time in the next few days to get some clear ideas down in mockups 👍

peripheralist:
I don't mind specifically supporting Uniswap/Sushiswap/both. Our choice of supported AMMs can be arbitrary to start off with—if the community were to pressure us to support other AMMs, we'd be open to it.

aeolianeth:
Closing this issue!

[...]

PR #311, opened 2021-12-31
Add token's Uniswap price to project page

aeolianeth:

This PR adds the Uniswap price for a project's ERC20 token to the project page.

The main additions are as follows:
- `TokenAMMPriceBadge.tsx` - this is the UI component for the price badge.
- `ERC20UniswapPrice.tsx` - exposes a hook `useUniswapPriceQuery` to fetch the exchange rate of `$TOKEN`/`WETH` on Uniswap.
- 2 dependencies: 
   -  `"@uniswap/sdk": "^3.0.3"`
   - `"@uniswap/sdk-core": "^3.0.1"`

The big diff is in part due to the lockfile changes, the uniswap logo, and the translation updates: do not be afraid! ☺️ 

| light | dark |
| --- | --- |
| Screen Shot 2021-12-31 at 1 38 44 pm | Screen Shot 2021-12-31 at 1 38 32 pm |

### Loading state

Uses the loading spinner:

Screen Shot 2021-12-31 at 1 38 59 pm

### When Uniswap has no liquidity pool for the token

Screen Shot 2021-12-31 at 2 04 16 pm

### When project hasn't issued an ERC20

Badge isn't shown (same as the current state).

## Acceptance checklist

https://github.com/jbx-protocol/juice-interface/issues/258


aeolianeth:
Hey @johnnyd-eth @peripheralist, I'd like to get some eyes on this sooner rather than later. There's a few little optimizations to make here and there, but I'd love some feedback on the implementation and UI 🙏 .

One thing that's confusing me at the moment is that there's a slight discrepancy between the price on app.uniswap.org and Juicebox. IMO it's not a huge issue as the user will end up on app.uniswap.org and see the "real" price anyway, but it would be nice to uncover what's causing the discrepancy.

Screen Shot 2021-12-31 at 7 50 44 pm

src/components/Dashboard/Pay.tsx
```diff
@@ -106,6 +113,16 @@ export default function Pay() {
 
   return (
     <div>
+      <div style={{ marginBottom: '0.5rem' }}>
+        {tokenSymbol && tokenAddress && (
+          <TokenAMMPriceBadge
```
aeolianeth, line 118: The idea would be to add a token badge for each DEX/AMM we want to add (next would probs be sushiswap)

src/hooks/ERC20UniswapPrice.ts
```diff
@@ -0,0 +1,155 @@
...
+}
+
+// https://docs.uniswap.org/protocol/reference/deployments
+const UNISWAP_V3_FACTORY_ADDRESS = '0x1F98431c8aD98523631AE4a59f267346ea31F984'
+const WETH_ADDRESS = '0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2'
+const UNISWAP_FEE = 10000
+const networkId = readNetwork.chainId
+const wadPrecision = 18
+
+export function useUniswapPriceQuery({ tokenSymbol, tokenAddress }: Props) {
```
aeolianeth, line 44: Most of this code comes from https://docs.uniswap.org/sdk/guides/fetching-prices

There's a few caching optimizations we can probably do if we're happy with this approach.

johnnyd-eth:
@aeolianeth fantastic and well-styled feature which will be very useful for users. 

My first thoughts around the UI are: 
- The 'U' in 'Uniswap' should always be capitalized 
- Not sure how it would look, but I'm wondering if having the tooltips left-aligned with the objects they're popping up from would look better?
- Other than that I think it looks great.

As for the code, do you think we should be using translation functions for all the new pieces of text added?

aeolianeth:
@johnnyd-eth thanks for the great review!

> * I think the 'U' in 'Uniswap' should always be capitalized

Yes, totally. I noticed this after taking the screenshots and fixed it: https://github.com/jbx-protocol/juice-interface/pull/311/files#diff-591484954f82f0218166243ccb6c44bc986c63988b53b3a17e304e8af0826f79R119

> * Not sure how it would look, but I'm wondering if having the tooltips left-aligned, rather than center-aligned, with the objects they're popping up from would look better?

Here's a comparison of placements. I think I actually prefer `topLeft` - WDYT? I don't have a strong opinion here.

| top | left | topLeft | 
| --- | --- | --- | 
| Screen Shot 2021-12-31 at 1 38 44 pm | Screen Shot 2022-01-01 at 8 26 44 am |  Screen Shot 2022-01-01 at 8 27 55 am |

> As for the code, do you think we should be using translation functions for all the new pieces of text added?

Yes, such a good point, thanks for raising this. I'll mark all the additional strings in the PR for translation.

johnnyd-eth:
> > * Not sure how it would look, but I'm wondering if having the tooltips left-aligned, rather than center-aligned, with the objects they're popping up from would look better?
> 
> Here's a comparison of placements. I think I actually prefer `topLeft` - WDYT? I don't have a strong opinion here.

@aeolianeth  Yep I also prefer topLeft

src/constants/tokens/dai.ts
```diff
@@ -1 +0,0 @@
-export const DAI = '0x6B175474E89094C44Da98b954EedeAC495271d0F'
```
aeolianeth, line 1: Noticed that this wasn't being used.

src/components/Dashboard/Pay.tsx
```diff
@@ -106,6 +113,16 @@ export default function Pay() {
 
   return (
     <div>
+      <div style={{ marginBottom: '0.5rem' }}>
+        {tokenSymbol && tokenAddress && (
```
peripheralist, line 117: nit pick—could we move line 117 outside of the parent div, so we aren't creating any new empty containers if there's no token?

src/components/Dashboard/Pay.tsx
peripheralist: love it 👏

src/components/shared/TokenAMMPriceBadge.tsx
```diff
@@ -0,0 +1,88 @@
...
+export default function TokenAMMPriceBadge({
+  exchangeName,
+  tokenSymbol,
+  tokenAddress,
+  exchangeLink,
+}: Props) {
+  const {
+    theme: { colors },
+  } = useContext(ThemeContext)
+  const Logo = LOGOS[exchangeName]
```
peripheralist, line 36: other nit: the convention we've been using is lowercase for variable names and Titlecase for Types/Components, so let's make `Logo` > `logo` and `LOGOS` -> `logos`?

love this pattern for making this component compatible with multiple exchanges! can't wait to see it with Sushi etc

src/hooks/ERC20UniswapPrice.ts
peripheralist: want to include this link as a comment in this file?

src/components/shared/TokenAMMPriceBadge.tsx
```diff
@@ -0,0 +1,88 @@
+import { LoadingOutlined } from '@ant-design/icons'
+import { useContext } from 'react'
+import { ThemeContext } from 'contexts/themeContext'
+import { useUniswapPriceQuery } from 'hooks/ERC20UniswapPrice'
+import { formattedNum } from 'utils/formatNumber'
+import UniswapLogo from 'components/icons/Uniswap'
+import { Tooltip } from 'antd'
+import { t } from '@lingui/macro'
+
+type exchangeName = 'Uniswap'
```
peripheralist, line 10: types are TitleCase

peripheralist (commented):
@aeolianeth this is looking SO good! Sorry I didn't get to this yesterday when I was catching up on recent PRs. Layout/placement/logo all feel great. And this will be easy to iterate on as we continue to play with the layout on this page and the Pay component.

Left a few nit pick comments, but everything is looking strong. My main concern is the extensibility of the TokenAMMPriceBadge component—do you feel confident that we'll be able to display the price for any exchange using a single Props type (exchangeName, exchangeLink, etc)? 

Loose thought: another option might be to leave the query functionality open ended for each AMM, so we could use a contract call, subgraph query, or api call as needed on a case-by-case basis. If we went that route, the TokenAMMPriceBadge component's job could just be to unify the styling and accept props like `(price: string, exchangeName: ExchangeName, isLoading: boolean)`. Then perhaps all AMM badges could be in a parent component AMMPrices.tsx responsible for all the hooks used to get AMM prices.

src/components/shared/TokenAMMPriceBadge.tsx
aeolianeth: ahh 🙈 . Thanks for catching it. See https://github.com/jbx-protocol/juice-interface/pull/323 😘

src/hooks/ERC20UniswapPrice.ts
aeolianeth: Totally, good call.

src/components/Dashboard/Pay.tsx
aeolianeth: Lovely suggestion, thanks!

src/hooks/ERC20UniswapPrice.ts
aeolianeth: https://github.com/jbx-protocol/juice-interface/pull/311/commits/32821cbcbc0f50a9eb0c8fe14740b46185b00df0

src/components/Dashboard/Pay.tsx
aeolianeth: https://github.com/jbx-protocol/juice-interface/pull/311/commits/32821cbcbc0f50a9eb0c8fe14740b46185b00df0

src/components/shared/TokenAMMPriceBadge.tsx
aeolianeth: https://github.com/jbx-protocol/juice-interface/pull/311/commits/32821cbcbc0f50a9eb0c8fe14740b46185b00df0

aeolianeth:
@peripheralist thanks for the great review!

I really like your suggestion of making `TokenAMMPriceBadge` a purely presentational component. I think the initial approach still allows for arbitrary token price query. But as as I type this out, I'm convinced about your idea of keeping `TokenAMMPriceBadge` presentational and wrapping it in a `AMMPrices` which fetches all the necessary data and renders `TokenAMMPriceBadge`s.

The downside would be that we wouldn't be able to slap the `TokenAMMPriceBadge` anywhere in the app as easily. But, that feels like over-optimization territory.

I've made the change and it feels great. There's still some features we could add to AMMPrices.tsx, like being able to filter which AMMs to display, but these will be simple future additions. https://github.com/jbx-protocol/juice-interface/pull/311/commits/b6fa8c5b621497ba519397bb09267791e997e067

Let me know what you think 🏓

src/components/Dashboard/Pay.tsx
```diff
@@ -138,8 +152,7 @@ export default function Pay() {
                   (payIn === 0 ? parseEther('1') : converter.usdToWei('1')) ??
                     BigNumber.from(0),
                 )}{' '}
-                {tokenSymbol ?? 'tokens'}/
-                <CurrencySymbol currency={payIn} />
+                {tokenSymbol ?? 'tokens'}/{currencyName(payIn)}
```
aeolianeth, line 155: btw, this changes the exchange rate to so token/ETH rather than token/Ξ. I think it's more consistent this way.

src/components/shared/TokenAMMPriceBadge.tsx
aeolianeth: Might be personal pref. but I had it capitalized because it's a component. I renamed it to `LogoComponent` - perhaps that's more clear. Don't have a strong opinion though - happy to change to `logo` 👍

peripheralist: ahhh, true, could go either way here. okay, i'll take the L.

peripheralist:
Another option might be to create unique components for each exchange, like UniswapAMMPrice.tsx which is responsible for any hooks needed to get the price, and for rendering a TokenAMMPriceBadge.tsx. That might be the best option for keeping hooks sorted, and for making it easy to show badges elsewhere around the app as needed.

Other than that this is feeling good to me! Before we merge @aeolianeth can you share this in the Discord to give folks an idea of what's coming, and to see if there's any feedback? This will be a pretty prominent change (which i'm stoked for)

src/components/shared/TokenAMMPriceBadge.tsx
aeolianeth: >  i'll take the L
LoL

aeolianeth:
@peripheralist 
> Another option might be to create unique components for each exchange, like UniswapAMMPrice.tsx which is responsible for any hooks needed to get the price, and for rendering a TokenAMMPriceBadge.tsx. That might be the best option for keeping hooks sorted, and for making it easy to show badges elsewhere around the app as needed.

Yeah that's another good possible approach. IMO let's wait and see how we might end up using the component/s in the future. It will be a fairly simply change that we can make when/if we need it 👍 

> I'll share this in the Discord

Sweet, thanks for sharing it!

peripheralist (approved):
huge upgrade for the app, well done 👏

PR #312, opened 2021-12-31
Add My Projects page

0xzachvai:
## What does this PR do and why?

Adds a new top level page for viewing a person's joined projects, as mentioned [on notion](https://juicebox.notion.site/Add-My-Projects-page-that-lists-projects-you-ve-bought-into-8083f44673304c07aa341f69bece117e) and issue #256 .

I setup the navbar link to only display if the user has a wallet connected, and to send them to the Projects page if they disconnect while on My Projects.

On Notion, there's also requests for including additional information for each project. I wasn't sure what designs for that might look like, so I left them off this PR for now.

## Acceptance checklist

- [x] I have evaluted the [Approval Guidelines](../../CONTRIBUTING.md#approval-guidelines) for this PR.

PR #315, opened 2022-01-01
formatGraphQuery Tests

0xzachvai:

While working on #312 , I noticed a couple oddities with the graph queries created by the `formatGraphQuery` util function, in particular that they would often include a redundant "id" key, e.g. `project { id id }`.

While the redundant key doesn't break things, it was an easy fix to make while still keeping the current behavior of always including at least `id`, so I did that. I also took this as an opportunity to write a couple of first tests, using Jest since it's already bundled in with react-scripts.

In the process of writing those tests, I noticed little quirks with formatting, so I addressed those as well. Example of a change:
```
# before:
projects(where: { id: "1" ,creator: "123"  }) { id }
                          ^               ^

# after:
projects(where: { id: "1", creator: "123" }) { id }
```

**One extra thing**, slightly unrelated to the above (if you'd prefer I split this lmk, I included it since it was a 2 line change in the same file): I updated `WhereConfig` to have context of the type it's filtering on, so that `key` is type aware, and only supports keys that exist on the entity being queried.

Not applicable.

## Acceptance checklist

### DIFF
--- a/src/utils/graph.ts
+++ b/src/utils/graph.ts
@@ -77,8 +77,8 @@ export interface SubgraphError {
 
 export type OrderDirection = 'asc' | 'desc'
 
-export type WhereConfig = {
-  key: string
+export type WhereConfig<E extends EntityKey> = {
+  key: EntityKeys<E>
   value: string | number | boolean
   operator?:
     | 'not'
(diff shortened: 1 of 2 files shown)


src/utils/graph.ts
```diff
@@ -136,6 +136,8 @@ export const formatGraphQuery = <E extends EntityKey, K extends EntityKeys<E>>(
     args += (args.length ? ', ' : '') + `${name}: ` + value
   }
 
+  if (!opts.keys.includes('id' as K)) opts.keys.push('id' as K)
```
peripheralist, line 139: i can't remember why `id` was ever hard-coded into the query string, and i don't think it needs to be. can we just remove this line for simplicity?

peripheralist:
great fixes! thanks for paying close attention.

peripheralist (approved):
thanks!

PR #326, opened 2022-01-05
Reorganize src directory

aeolianeth:
This PR relates to https://www.notion.so/juicebox/Establish-clearer-organization-and-naming-conventions-for-app-files-directories-e72d12f9b773416d92a913c5514f75bc.

This PR does the following:

- create `providers/` directory and move relevant components there.
- move `App.tsx` to `src/` directory.
- remove `CatchallRedirect.tsx` file, move it's component definition to `Router.tsx`.

No user-facing changes should arise from this PR.

## Acceptance checklist


src/Router.tsx
```diff
@@ -1,10 +1,14 @@
...
+import { Redirect, useParams } from 'react-router'
+import Dashboard from 'pages/Dashboard'
+import Landing from 'pages/Landing'
+import Create from 'pages/Create'
+import Projects from 'pages/Projects'
 
-import CatchallRedirect from './CatchallRedirect'
-import Create from './Create'
-import Projects from './Projects'
+function CatchallRedirect() {
```
aeolianeth, line 8: CatchallRedirect is only used in this file - seemed like a good opportunity to remove a file.

peripheralist:
i like where this is going! the directory structure could definitely use improving and this feels like a step in the right direction. actually @0xzachvai recently added a few tests for formatting graph queries—not much but it's a start. let's leave the test script in for now

aeolianeth:
@peripheralist 
> actually @0xzachvai recently added a few tests for formatting graph queries—not much but it's a start. let's leave the test script in for now

Ohh no way! That's amazing, very happy to hear that. Will re-add the test stuff!! ❤️

aeolianeth:
@peripheralist test deps and scripts re-added. 🙏 

Perhaps we should leave this PR open for another day to give @jnoh @johnnyd-eth @torvusbug a chance to comment?

torvusbug:
This feels right to me. It's akin to the NextJS organizational structure, just without the filesystem routing.

johnnyd-eth:
Awesome contribution here @aeolianeth this will be super handy for new devs. There's improvement to be made in this area and this is a great step in the right direction.

aeolianeth:
Thanks for taking a look @torvusbug @johnnyd-eth! 🙏 

@peripheralist, think it's safe to send this back your way 🏓

peripheralist:
after playing with this a bit more @aeolianeth , it feels a bit odd to me that the files under Components/ are grouped into subfolders named after Pages, but the page components themselves are in a separate dir. It feels like if we wanted to separate Pages from smaller/shareable components, they should be organized by functionality rather than where they're intended to be used.

I might be overthinking this or be just too married to the logic behind the current structure: folders organized by pages, with subcomponents grouped into their parent folder, and _shared_ components separated entirely in a Shared folder.

aeolianeth:
> It feels like if we wanted to separate Pages from smaller/shareable components, they should be organized by functionality rather than where they're intended to be used.

@peripheralist definitely hear what you're saying. I was torn about this, too. Personally, I like having a clear idea about what is a page and what is not. But I'm agreeing with your statement above - having the Page directory mirrored in Components feels wrong.

I wonder if this is a reasonable step in the right direction:
- move each page component directory to `pages/` (including each page's related components)
- move everything in `components/shared/*` to `components/` (i.e. remove `shared/` dir)
   - the implication at this point is that everything in `components/` is considered `shared/`

As an alternative, I'm happy to omit this change from the PR. Perhaps we are trying to fix something that isn't broken 🤔 

cc @torvusbug @johnnyd-eth

johnnyd-eth:
> > It feels like if we wanted to separate Pages from smaller/shareable components, they should be organized by functionality rather than where they're intended to be used.
> 
> @peripheralist definitely hear what you're saying. I was torn about this, too. Personally, I like having a clear idea about what is a page and what is not. But I'm agreeing with your statement above - having the Page directory mirrored in`components/` feels wrong.
> 
> I wonder if this is a reasonable step in the right direction:
> 
> * move each page component directory to `pages/` (including each page's related components)
> * move everything in `components/shared/*` to `components/` (i.e. remove `shared/` dir)
>   
>   * the implication at this point is that everything in `components/` is considered `shared/`
> 
> As an alternative, I'm happy to omit this change from the PR. Perhaps we are trying to fix something that isn't broken 🤔
> 
> cc @torvusbug @johnnyd-eth

I agree with @peripheralist that shared components should be outside of the `pages` directories and organized in terms of functionality rather than their location in the app. But I still think making a `pages` directory with all top-level, unique components (and any of their sub-components not used elsewhere, such as the new navbar ones) would be useful in making the code easier to navigate.

src/index.tsx
```diff
@@ -1,34 +1,35 @@
...
-import reportWebVitals from './reportWebVitals'
-import ProvideReactQuery from './ReactQuery'
-import { LanguageProvider } from './i18n'
+import reportWebVitals from './utils/reportWebVitals'
+import ReactQueryProvier from './providers/ReactQueryProvider'
 
 ReactDOM.render(
   <React.StrictMode>
-    <ProvideReactQuery>
+    <ReactQueryProvier>
```
aeolianeth, line 20: This diff is renaming the components to be consistent (with `Provider` suffix)

src/providers/LanguageProvider.tsx
```diff
@@ -32,12 +32,16 @@ const getLocale = (): string => {
...
 
 async function dynamicActivate(locale: string) {
-  const { messages } = await import(`./locales/${locale}/messages`)
+  const { messages } = await import(`../locales/${locale}/messages`)
   i18n.load(locale, messages)
   i18n.activate(locale)
 }
 
-export function LanguageProvider({ children }: { children: ReactNode }) {
+export default function LanguageProvider({
```
aeolianeth, line 40: Changed to default export to be consistent with other providers.

aeolianeth:
Hey @peripheralist, in the interest of getting something merged here, I've removed the addition of the `pages/` directory for the meantime.

I did some investigation and it seems like the `pages/` vs `components/` discussion mostly comes down to personal preference. I think I was initially trying to fix something that wasn't broken.

This PR is quite small now, but still cleans the `src/` directory up nicely. Let me know what you think 🏓

peripheralist (commented):
good stuff @aeolianeth ! feels better to me, let's go with it.

just a few non-functional nits: want to also rename the actual exports from `UserProvider`, `NetworkProvider`, `ReactQueryProvider`, and `ThemeProvider` to match the filenames?

aeolianeth:
> good stuff @aeolianeth ! feels better to me, let's go with it.
> 
> just a few non-functional nits: want to also rename the actual exports from `UserProvider`, `NetworkProvider`, `ReactQueryProvider`, and `ThemeProvider` to match the filenames?

@peripheralist Ah, good catch! Done => https://github.com/jbx-protocol/juice-interface/pull/326/commits/1fd32d54542c26de9ec515ccdde64e91477d05ed. Also renamed the PR to be more accurate 🏓

peripheralist (approved):

PR #321, opened 2022-01-04
Add tooltips to funding progress bar

aeolianeth:
## What does this PR do and why?

Adds 2 basic tooltips to the funding progress bar on the project page. The aim is to help folks understand the progress bar easier.

## Screenshots or screen recordings

| target tooltip | overflow tooltip |
| --- | --- |
| Screen Shot 2022-01-04 at 9 42 25 pm | Screen Shot 2022-01-04 at 9 42 19 pm |

## Acceptance checklist

- [x] I have evaluted the [Approval Guidelines](../../CONTRIBUTING.md#approval-guidelines) for this PR.

PR #334, opened 2022-01-07
Add constants for currencies

aeolianeth:

When working on https://github.com/jbx-protocol/juice-interface/pull/321 I found it difficult to remember which currency was `0` and which was `1`.

This PR creates constants for these values ([in this file](https://github.com/jbx-protocol/juice-interface/pull/334/files#diff-59da7d045d0a68f4f5dd5002596eaa1277948140874bea0dd143ee1eb7d68018)), and uses them everywhere where we previously used `0` or `1`.

No changes.

## Acceptance checklist

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -10,5 +10,5 @@ changes._
 
 ## Acceptance checklist
 
-- [ ] I have evaluted the [Approval Guidelines](../../CONTRIBUTING.md#approval-guidelines) for this PR.
+- [ ] I have evaluated the [Approval Guidelines](../../CONTRIBUTING.md#approval-guidelines) for this PR.
 - [ ] I have tested this PR in [all supported browsers](../../CONTRIBUTING.md#browser-support).
--- a/src/components/Dashboard/BalanceTimeline.tsx
+++ b/src/components/Dashboard/BalanceTimeline.tsx
@@ -30,6 +30,7 @@ import { querySubgraph } from 'utils/graph'
 import { readProvider } from 'constants/readProvider'
 
 import SectionHeader from './SectionHeader'
+import { CURRENCY_ETH } from 'constants/currency'
 
 const now = moment.now() - 5 * 60 * 1000 // 5 min ago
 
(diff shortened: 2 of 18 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -10,4 +10,4 @@ changes._
 
 ## Acceptance checklist
 
-- [ ] I have evaluted the [Approval Guidelines](../../CONTRIBUTING.md#approval-guidelines) for this PR.
+- [ ] I have evaluated the [Approval Guidelines](../../CONTRIBUTING.md#approval-guidelines) for this PR.
```
aeolianeth, line 13: A passing typo fix

aeolianeth:
@johnnyd-eth 👋 , could you please review this PR?

peripheralist:
i dig this! really appreciate your help lately with the fundamental stuff 🙏 . @johnnyd-eth , @torvusbug, @jnoh any thoughts?

johnnyd-eth:
Another great cleanup here @aeolianeth. Your contributions are making for a much nicer dev experience, thank you!

Non-blocking thought: I'm wondering if there is any place in the UI for other currencies like Chinese RMB? In which case it would be an addition to your new `currency.ts` file.

aeolianeth:
> I'm wondering if there is any place in the UI for other currencies like Chinese RMB?

@johnnyd-eth glad you brought this up. I'd imagine not for now, as a project is "based" in either USD or ETH. Although we could allow user to view the site in their local currency one day 🤔 .

A related thought I had: it's not entirely clear that the dollar values are in USD currency; many currencies (like AUD) use the `$` symbol for values. I think we should find a way to be more explicit about this.

torvusbug:
@aeolianeth Definitely a good idea to make these consistent. I think this is a good solution.

peripheralist:
Agreed with @aeolianeth, the ETH/USD options are just derived from what the contracts support so adding in other currencies would be complicated (we could probably manage it with conversions on the frontend, but might not be worth the complexity).

Also great feedback on the use of $. I'm not sure the best way to clarify it without using USD everywhere, which could feel awkward. Open to suggestions

aeolianeth:
@peripheralist thanks! Conflicts fixed and issue created => https://github.com/jbx-protocol/juice-interface/issues/341

Back to you! 🏓

PR #338, opened 2022-01-08
Add mdx support

carlbarrdahl:

Support for MDX (and Markdown) in Project descriptions.
This allows for embedding of components like:

```md
# Project heading 1
- this is
- a list

<YouTube youtubeId="GoltwBHXCx8" />
```

Even more advanced components could be added:
```md
<Poll questions={["define questions here"]} />
```

https://juicebox.notion.site/Allow-video-embeds-in-project-3470d259ab26418eafd9195acbc7885d

Sugggestion to use MDX
https://discord.com/channels/775859454780244028/786985806396784690/929072517434454027

## Acceptance checklist


aeolianeth:
@carlbarrdahl this is awesome! Thanks for this contribution 🙏 . Love me some markdown.

I think we may need to consider the design of the project page a little more closely before we proceed with this. Because the user has a lot more control of their project description, we will lose consistency in the design across project pages. As it stands now, things like headings, bullet-points etc. might look a bit funky.

I definitely see the value in allowing the user to provide a more details description though. There's been some thinking around [how we might improve the project page with a tab-based layout](https://www.notion.so/juicebox/Improve-UI-for-individual-project-page-9d7d7dcf10c84c088d5a3cb7e21946c2). As an example, I an imagine an `Overview` tab where the users' MDX-defined description would be rendered.

Keen to hear others thoughts about this PR though. The solution you've proposed looks perfect - I just think we need to consider the UI implications more closely 👍

carlbarrdahl:
Thank you @aeolianeth!
I totally understand the considerations for consistency. Let me know if I can help out in any way.

Just want to mention that it's possible to define how each of the markdown rendered components should look like here:
https://github.com/jbx-protocol/juice-interface/blob/83f14d60469fc2aac34dc6a3e4f744d9ff900558/src/components/shared/MDXViewer.tsx#L10-L14

For example with antd components:
```tsx
import { Image, Typography } from 'antd';

const components: MDXComponents | any = { 
  YouTube,
  h1: props => <Typography.Title level="h1" {...props} />,
  img: props => <Image {...props} />
  ul: props => ...,
}
```

This makes it fairly straightforward styling the html components rendered from the markdown.

aeolianeth:
@carlbarrdahl is there a way to restrict MDX such that only `<Youtube />` (or another video element) is supported?

We'll need more time to consider a full-blown markdown description in the context of the project page UI, but I think we can aim to get video embed shipped ASAP. What do you think?

carlbarrdahl:
@aeolianeth Fixed!

Project description:

Rendered:


https://github.com/jbx-protocol/juice-interface/blob/515f921c29f227d35dbfc6ed89975301a2f880a6/src/components/shared/MDXViewer.tsx#L42-L47

carlbarrdahl:
I have given this a bit more thought.

One concern is how to communicate this convention to users of adding a video with: `<YouTube youTubeId="<id>" />` in the project description. 

It would be a lot better UX to just paste the link in the description: https://www.youtube.com/watch?v=<id>. The text could then be regexped for urls matching youtube (and other providers) and then embed.

Here are two that does this with remark:
https://github.com/remark-embedder/transformer-oembed
https://github.com/sergioramos/remark-oembed

However they are requesting cors resources (https://oembed.com/providers.json and https://www.youtube.com/oembed?url=https://www.youtube.com/watch?v=<id>) so it doesn't work in browsers.

Another option would be to not use MDX at all for now and just regexp the project description and render an iframe:
```html
<iframe
  width="853"
  height="480"
  src="https://www.youtube.com/embed/<id>"
  frameBorder="0"
  allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture"
  allowFullScreen
/>
```

I guess it very much depends on what kind of direction you want to take for the future.
The MDX route could enable some really cool widgets for project pages but it might also confuse the users unless communicated in a clear way, probably with lots of examples).

What are your thoughts?

aeolianeth:
@carlbarrdahl sorry for the delay here! Will get back to you tomorrow (UTC+10)

aeolianeth:
@carlbarrdahl thanks for your patience!

Here's my thoughts on MDX: I can definitely see this being in the product one day, but as it stands I feel that it might not be suitable. The Project page isn't flexible enough to accommodate much stylistic variety in descriptions, so I worry that the project-to-project experience may suffer. As you point out, there is also an educational piece that comes along with it. There is a Notion task to improve the project page, so I would encourage you to drop a link to this proof-of-concept in there - I think we should definitely keep it in mind: https://www.notion.so/juicebox/Improve-UI-for-individual-project-page-9d7d7dcf10c84c088d5a3cb7e21946c2

As for what to do about video embeds in the short-term, I think your Iframe idea is much more tenable. There are a few considerations here that we'd still have to iron out - will we build the iframe ourself and get the video link for the user, or will we let the user drop a HTML snippet in (and how do we do this securely), for example - but these are answerable questions!

cc @peripheralist @JayZeeDesign for additional thoughts

carlbarrdahl:
@aeolianeth Thanks for your input! I added the link here to the notion page.

Perhaps a better use-case would be a full-blown landing page for the project that acts more as a promotional information page rather than interacting with the juicebox smart contracts (could do that as well of course).

A powerful combination might be to use system-UIs theme specification for the visual appereance of the page with MDX for describing content and widgets. 

I would be interested in building such proof of concept if it would be of interest to the project.

Here's a post I made in the #dev channel about Custom layout themes
https://discord.com/channels/775859454780244028/786985806396784690/929705828393553940
https://www.notion.so/juicebox/Custom-layout-themes-48e54aa6f7eb48679fd37461add1928f

Relevant links
- https://github.com/system-ui/theme-specification
- https://chakra-ui.com/docs/theming/theme
- https://reflexjs.org

aeolianeth:
Niiice @carlbarrdahl! Really liking the sound of everything you've said. the Design channel in Discord would be the best place to start a discussion around landing pages 🚀

aeolianeth:
Thanks for the summary @carlbarrdahl. Let's close this PR for now and revisit in the future!

PR #313, opened 2022-01-01
Improvements to funding and distribution forms

johnnyd-eth:
## What does this PR do and why?

This PR improves how the funding target is set/changed, as well as how individual distributions are set/changed:

1. Allows user to input **funding target after Juicebox fee** and have it set the **funding target**
2. Allows user to set distribution payout as an **amount** rather than just a **percentage** of the funding target

These were ideas sent to me from @mejango 

## Screenshots or screen recordings

Overview:

https://user-images.githubusercontent.com/96150256/147845351-9d85070c-81a8-494e-b45d-fd5019bbdd49.mp4

Rounds **amount** to floor of 2 d.p. **percentage**:

https://user-images.githubusercontent.com/96150256/147865936-1341550a-464b-4456-a3a2-3ab5d44e6651.mp4

Fix bug where subsequent payout of a project payout gets assigned that project (before/after):

https://user-images.githubusercontent.com/96150256/147891846-af99cd86-35ad-477c-b6dc-778c73e9c2a2.mp4

Address @peripheralist's 5 points of feedback:

https://user-images.githubusercontent.com/96150256/148140402-f241a3c4-8caf-4fc3-a70d-f7f9555ac2d2.mp4

## Acceptance checklist

- [x] I have evaluted the [Approval Guidelines](../../CONTRIBUTING.md#approval-guidelines) for this PR.

PR #342, opened 2022-01-09
Use new library to better handle number inputs with decimals

torvusbug:

I found it slightly difficult to use some of the number inputs. To fix it, I'm using a new library [react-number-format](https://www.npmjs.com/package/react-number-format) to handle it.

https://user-images.githubusercontent.com/4457995/148684073-41079b20-2383-4f42-999f-b56738427ed9.mp4

## Acceptance checklist

- [x] I have tested this PR in [all supported browsers](../../CONTRIBUTING.md#browser-support).

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
     "ethers": "^5.1.0",
     "react": "^17",
     "react-dom": "^17.0.1",
+    "react-number-format": "^4.9.1",
     "react-query": "^3.33.2",
     "react-redux": "^7.2.2",
     "react-router-dom": "^5.2.0",
--- a/src/components/Create/BudgetForm.tsx
+++ b/src/components/Create/BudgetForm.tsx
@@ -40,7 +40,7 @@ export default function BudgetForm({
   const [currency, setCurrency] = useState<CurrencyOption>(0)
   const [target, setTarget] = useState<string>('0')
   const [targetSubFee, setTargetSubFee] = useState<string>('0')
-  const [duration, setDuration] = useState<string>('0')
+  const [duration, setDuration] = useState<string | undefined>('0')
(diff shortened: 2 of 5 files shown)


peripheralist:
Great stuff. No idea why I didn't start out using a library like this 😅

Only issue I see is that the input breaks when its value is just `.`—i.e. if typing `.5` without a leading zero it breaks after typing the decimal. Looks solid otherwise!

aeolianeth:
@torvusbug ohh, wow, this is awesome <3.

We may need to wait for https://github.com/jbx-protocol/juice-interface/pull/313 to merge as I believe it touches these number inputs.

Also curious about the weight the new dep adds to our bundle - want to run a `yarn analyze`?

peripheralist:
are we feeling good about adding this new dependency @aeolianeth @torvusbug ?

also, could we be replacing other instances of the `formattedNum()` util with this? (can be a separate PR)

torvusbug:
@peripheralist I'm trying to fix your discovered bug where ".5" creates a runtime error. There are other number-only inputs I've tried to replace with this but ran into state issues. I'll try to solve some of these and see how far I can get before moving this PR forward.

aeolianeth:
> are we feeling good about adding this new dependency @aeolianeth @torvusbug ?
> 
> also, could we be replacing other instances of the `formattedNum()` util with this? (can be a separate PR)

Yep, feeling real good about this, thanks again @torvusbug. Still curious about the added bundle size, but can't imagine it would be a blocker.

torvusbug:
@aeolianeth I'll close this for now and reopen when I think it's ready. There's still a runtime bug when typing ".5" instead of "0.5". Something to do with how one of the libraries parses ".5".

PR #373, opened 2022-01-14
Fix typo in Incentives section

aeolianeth:


peripheralist:
@aeolianeth good catch! when we make changes like this, do we also need to be updating the `msgid` in a .po file somewhere?

johnnyd-eth:
> @aeolianeth good catch! when we make changes like this, do we also need to be updating the `msgid` in a .po file somewhere? cc @johnnyd-eth

@peripheralist Now that @aeolianeth got the source file deleting and starting from scratch when `i18n:extract`ing, the old one should just be replaced by the correct one. I'm also 90% sure this will be reflected in Crowdin, but will take a look after this is merged.

aeolianeth:
@johnnyd-eth I think it might still make sense to commit the changes to the english .po file to err on the side of caution.

@peripheralist If we don't do it here, it may still happen when Crowdin creates a PR. But let's not leave that to chance.

aeolianeth:
@peripheralist @johnnyd-eth to test our assumption, I manually ran the Crowdin download action to create this PR: https://github.com/jbx-protocol/juice-interface/pull/376/files. It has the correct changes to our source file: `src/locales/en/messages.po`.

The main problem with deferring the extraction to the Crowdin download action is lack of continuity from the codebase to Crowdin. The source file should be the single source of truth. By not updating the source file in each PR, the source file is temporarily out of date while we wait for the Crowdin download to run.

It's not immediately clear what the implications of that are, but I feel that best practice would be to keep the source file as tightly in-sync with the codebase as possible at all times. Fortunately, LinguiJS still renders text that doesn't correspond to a `msgid`, so we're safe there.

src/locales/en/messages.po
```diff
@@ -95,7 +136,7 @@ msgid "Address"
 msgstr ""
 
 #: src/components/Create/index.tsx
-msgid "Adjust incentivizes for paying your project."
+msgid "Adjust incentives for paying your project."
```
aeolianeth, line 139: The main concern of this PR

aeolianeth:
@peripheralist I ran the extraction script to generate this commit for this PR: https://github.com/jbx-protocol/juice-interface/pull/373/commits/5ad451718b5402bf1da89e02c673db4f532575d1

It contains all the strings that have landed in `main` because we haven't merged a Crowdin download PR in a few days.

In future, let's run `yarn i18n:extract` manually for all PRs to generate a fresh copy of the source file `src/locales/en/messages.po`.

cc @johnnyd-eth

johnnyd-eth:
> @peripheralist I ran the extraction script to generate this commit for this PR: [5ad4517](https://github.com/jbx-protocol/juice-interface/commit/5ad451718b5402bf1da89e02c673db4f532575d1)
> 
> It contains all the strings that have landed in `main` because we haven't merged a Crowdin download PR in a few days.
> 
> In future, let's run `yarn i18n:extract` manually for all PRs to generate a fresh copy of the source file `src/locales/en/messages.po`.
> 
> cc @johnnyd-eth

Yeah I get you @aeolianeth. SGTM

package.json
```diff
@@ -53,7 +53,7 @@
     "lint:fix": "yarn lint --fix",
     "lint-staged": "lint-staged",
     "prepare": "husky install",
-    "i18n:extract": "rm -f ./src/locales/en/messages.po && lingui extract --locale en",
+    "i18n:extract": "lingui extract --clean --locale en",
```
aeolianeth, line 56: TIL that lingui has a `--clean` option which achieves a similar result.

.linguirc.json
```diff
@@ -13,5 +13,6 @@
   "fallbackLocales": {
     "default": "en"
   },
-  "locales": ["en", "zh", "ru", "tr"]
+  "locales": ["en", "zh", "ru", "tr"],
+  "sourceLocale": "en"
```
aeolianeth, line 17: Added sourceLocale

aeolianeth:
@peripheralist @johnnyd-eth sorry for the confusion here. I sat down and properly thought about how this might break if we _aren't_ committing the updated source translation file in each PR. It turns out that there's no risk in not doing this. Our translations might be described as **eventually consistent**.

**tl;dr: we don't need to commit changes to `.po` files manually.**

Below is the scenario I was worried about, but after sketching it out I don't see it being a problem.

> Assumptions:
> - that we _never_ commit the source translation file to main (in other words, no human ever touches the `.po` files).
> - commits **A** and **B** contain string updates in the source code. 
> Note:
> - In the Crowdin upload, we are re-compiling the source translation file, meaning it's always up-to-date when we push to Crowdin.

crowdin

In this example, there is a period of time where the source code will be at **B**, but the translations will be at **A**. While this seems like it would break something, it is actually fine. Say in **A** we uploaded the string **Overflow**, and we had translations for this word. Now, say **B** changed this word to **Overflow.** (with a period). In the example, we download the string **Overflow** but not **Overflow.**. **Lingui doesn't break** - if a translation doesn't exist for a string, lingui renders the text as-is, without substitution.

Eventually, **B** will be downloaded, and the string **Overflow** will be replaced with **Overflow.** in the source translation file; we achieve eventual consistency.

peripheralist (approved):
nice! thanks @aeolianeth and @johnnyd-eth for all the effort in mapping this out. all the up-front effort will be well worth it once we can set and forget the crowdin integration

PR #374, opened 2022-01-14
React & Contract usage testing

torvusbug:

Create reusable foundation on testing within React and the underlying usage of contract readers/writers (`useContractReader`, `readProvider`, `useTransactor`, etc.)

None at the moment, but will provide imagery of running tests.

## Acceptance checklist

- [x] I have tested this PR in [all supported browsers](../../CONTRIBUTING.md#browser-support).

### DIFF
--- a/src/components/shared/formItems/__tests__/EthAddress.test.tsx
+++ b/src/components/shared/formItems/__tests__/EthAddress.test.tsx
@@ -8,14 +8,9 @@ import { Form } from 'antd'
 
 import { render } from '../../../../../test/test-utils'
 import EthAddress from '../EthAddress'
+import { readProvider } from '../../../../constants/readProvider'
 
-const lookupAddress = jest.fn()
-const resolveName = jest.fn()
-
-jest.mock('../../../../constants/readProvider', () => ({
-  lookupAddress,
-  resolveName,
-}))
+jest.mock('../../../../constants/readProvider')
 
 function TestableEthAddress() {
   const [testForm] = useForm<{ address: string }>()
@@ -26,7 +21,6 @@ function TestableEthAddress() {
     <Form form={testForm}>
       <Form.Item name="address">
         <EthAddress
-          name="address"
           onAddressChange={a => setAddress(a)}
           defaultValue={undefined}
         />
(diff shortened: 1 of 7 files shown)


src/components/shared/formItems/__tests__/EthAddress.test.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+import { useForm } from 'antd/lib/form/Form'
+
+import { Form } from 'antd'
+
+import { render } from '../../../../../test/test-utils'
+import EthAddress from '../EthAddress'
+import { readProvider } from '../../../../constants/readProvider'
+
+jest.mock('../../../../constants/readProvider')
```
torvusbug, line 13: I don't love this, but it can work for us in a limited capacity until there's a sort-of mocked chain we can lean on. It can be spun up in `src/setupTests.ts` in `beforeEach` (or `beforeAll` if it can be completely reset within `beforeEach`)

src/components/shared/formItems/__tests__/EthAddress.test.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+function TestableEthAddress() {
+  const [testForm] = useForm<{ address: string }>()
+
+  const [, setAddress] = useState('')
+
+  return (
+    <Form form={testForm}>
+      <Form.Item name="address">
+        <EthAddress
```
aeolianeth, line 23: **question**: to make this test as tightly-scoped as possible, couldn't we instead pass a mocked function to `onAddressChange`, and assert that it was called with the new address? I think we can get away with testing `EthAddress` in isolation, without the need for the wrapping `Form`

src/setupTests.ts
```diff
@@ -3,3 +3,35 @@
 // expect(element).toHaveTextContent(/react/i)
```
aeolianeth, line 3: **praise**: nice job figuring out this!

src/components/shared/formItems/__tests__/EthAddress.test.tsx
aeolianeth: I think this is totally valid; indeed, ideal. From the perspective of this component, `readProvider`'s behavior is irrelevant. As long as `EthAddress` is calling the correct things, `readProvider`'s behavior can be independently tested.

aeolianeth:
Hey @torvusbug @peripheralist, based on the [corrective actions identified](https://www.notion.so/juicebox/RCA-production-downtime-2022-01-29-ca703708b20543b3825f8dbaf0c1db25#74d633d0e4634d2e82fe73eba09237a5) from the recent downtime, I'd really like to add a basic test to ensure the app renders correctly.

I think we can use most of the code in this PR, except perhaps for simplicity we:
- remove the `EthAddress` spec (for now).
- add a spec for `App.tsx` or `index.ts`.

I'm happy to drive this effort. Let me know what ya'll think ✌️

aeolianeth:
Going to close this for now but let's revisit it soon.

PR #375, opened 2022-01-14
formatNumberConfig decimals > precision

peripheralist:

Previously the `formatNumber` util assumed the number to format used 18 decimal places. Although this is standard, not all ERC20s use 18 decimal places, and formatting non-standard token balances requires specifying the number of decimal places to use.

The `decimals` property has been renamed to `precision`, and a new `decimals` property is now used to specify the decimal length of the number to be formatted.

For consistency, the `WAD_PRECISION` constant is now named `WAD_DECIMALS`.

Currently the only place this new `decimals` property is used is in the `ERC20TokenBalance` component, which reads the decimal places from the ERC20 contract.

## Acceptance checklist

- [ ] I have tested this PR in [all supported browsers](../../CONTRIBUTING.md#browser-support).

### DIFF
--- a/src/utils/formatNumber.ts
+++ b/src/utils/formatNumber.ts
@@ -119,7 +119,7 @@ export const formattedNum = (
     if (post === '0') return formatNearZero(pre)
 
     const formattedPost = post
-      .substring(0, config?.precision ?? 18)
+      .substring(0, config?.precision ?? WAD_DECIMALS)
       .padEnd(config?.padEnd ? config?.precision ?? 0 : 0, '0')
 
     // If we can ignore postDecimal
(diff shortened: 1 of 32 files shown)


torvusbug (approved):

src/utils/formatNumber.ts
```diff
@@ -113,11 +119,11 @@ export const formattedNum = (
     if (post === '0') return formatNearZero(pre)
 
     const formattedPost = post
-      .substr(0, config?.decimals ?? 18)
-      .padEnd(config?.padEnd ? config?.decimals ?? 0 : 0, '0')
+      .substring(0, config?.precision ?? 18)
```
aeolianeth, line 122: **question (non-blocking)**: is falling back to a precision of `18` meaningfully related to `WAD_DECIMALS` (which is also `18`)?

aeolianeth:
@peripheralist great PR, thanks for adding this clarity. Left one question.

src/utils/formatNumber.ts
peripheralist: definitely meaningfully related! it's probably better to use `WAD_DECIMALS` here instead `18`.

aeolianeth (approved):

Issue #343, opened 2022-01-09
[IDEA] Wrap `transactor()` hook instances into unique hooks


peripheralist:
## Problem to solve

transactor() hook instances lengthen files due to their complexity, and are occasionally reused.

## Proposal

Specific transactor() uses could be wrapped in their own hooks, which would:
- keep code organized
- prevent code duplication
- make it easier to add tests for creating transactions

## More details

![image](https://user-images.githubusercontent.com/79433522/148697206-97e7a468-91be-4fb8-9e62-539964fa5878.png)
Could become `usePay(note: string, preferUnstaked: boolean, value: BigNumber)` with other args derived from `ProjectContext`/`UserContext`

aeolianeth:
Love this idea. I'm interested in getting more familiar with this code, so will assign myself if you're not already working on it.

peripheralist:
done!

PR #382, opened 2022-01-16
Replace uses of transactor() with unique hooks

peripheralist:

Addresses #343 

This creates unique hooks for all current instances of `transactor()`, using a new `TransactorInstance<T>` type, where the arguments passed to the transactor are an object of type `T`. These hooks are named with the convention `use<transaction-name>Tx`.

This allows for:
- Typed arguments when creating transactions from components
- More logical separation of concern
- Less code repetition

## Acceptance checklist

### DIFF
--- a/src/components/modals/ReconfigureFCModal.tsx
+++ b/src/components/modals/ReconfigureFCModal.tsx
@@ -14,15 +14,13 @@ import TicketModsList from 'components/shared/TicketModsList'
 
 import { ProjectContext } from 'contexts/projectContext'
 import { ThemeContext } from 'contexts/themeContext'
-import { UserContext } from 'contexts/userContext'
 import { constants } from 'ethers'
 import { useAppDispatch } from 'hooks/AppDispatch'
 import { useEditingFundingCycleSelector } from 'hooks/AppSelector'
 import { useTerminalFee } from 'hooks/TerminalFee'
+import { useConfigureProjectTx } from 'hooks/transactor/ConfigureProjectTx'
 import { CurrencyOption } from 'models/currency-option'
-import { FundingCycle } from 'models/funding-cycle'
 import { FundingCycleMetadata } from 'models/funding-cycle-metadata'
-import { FCProperties } from 'models/funding-cycle-properties'
 import { PayoutMod, TicketMod } from 'models/mods'
 import { useCallback, useContext, useLayoutEffect, useState } from 'react'
 import { editingProjectActions } from 'redux/slices/editingProject'
@@ -49,21 +47,12 @@ import PayModsForm from '../Create/PayModsForm'
 import TicketingForm, { TicketingFormFields } from '../Create/TicketingForm'
 
 export default function ReconfigureFCModal({
-  fundingCycle,
-  projectId,
   visible,
   onDone,
-  payoutMods,
-  ticketMods,
 }: {
   visible?: boolean
-  fundingCycle: FundingCycle | undefined
-  projectId: BigNumber | undefined
-  payoutMods: PayoutMod[] | undefined
-  ticketMods: TicketMod[] | undefined
   onDone?: VoidFunction
 }) {
-  const { transactor, contracts } = useContext(UserContext)
   const { colors, radii } = useContext(ThemeContext).theme
   const [currentStep, setCurrentStep] = useState<number>()
   const [payModsModalVisible, setPayModsFormModalVisible] =
@@ -87,9 +76,19 @@ export default function ReconfigureFCModal({
   const [editingPayoutMods, setEditingPayoutMods] = useState<PayoutMod[]>([])
   const [editingTicketMods, setEditingTicketMods] = useState<TicketMod[]>([])
   const dispatch = useAppDispatch()
-  const { currentFC, terminal, isPreviewMode } = useContext(ProjectContext)
+  const {
+    queuedFC,
+    currentFC,
+    terminal,
+    isPreviewMode,
+    queuedPayoutMods,
+    currentPayoutMods,
+    queuedTicketMods,
+    currentTicketMods,
+  } = useContext(ProjectContext)
   const editingFC = useEditingFundingCycleSelector()
   const terminalFee = useTerminalFee(terminal?.version)
+  const configureProjectTx = useConfigureProjectTx()
 
   const resetTicketingForm = () =>
     ticketingForm.setFieldsValue({
@@ -149,6 +148,14 @@ export default function ReconfigureFCModal({
   }
 
   useLayoutEffect(() => {
+    const fundingCycle = queuedFC?.number.gt(0) ? queuedFC : currentFC
+    const payoutMods = queuedFC?.number.gt(0)
+      ? queuedPayoutMods
+      : currentPayoutMods
+    const ticketMods = queuedFC?.number.gt(0)
+      ? queuedTicketMods
+      : currentTicketMods
+
     if (
       !visible ||
       isPreviewMode ||
@@ -184,10 +191,13 @@ export default function ReconfigureFCModal({
       })
     }
   }, [
+    currentFC,
+    queuedFC,
+    currentPayoutMods,
+    queuedPayoutMods,
+    currentTicketMods,
+    queuedTicketMods,
     dispatch,
-    fundingCycle,
-    payoutMods,
-    ticketMods,
     ticketingForm,
     restrictedActionsForm,
     isPreviewMode,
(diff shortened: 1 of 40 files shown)


src/hooks/transactor/PayProjectTx.ts
```diff
@@ -0,0 +1,42 @@
...
+import { TransactorInstance } from './Transactor'
+
+export function usePayProjectTx(): TransactorInstance<{
+  note: string
+  preferUnstaked: boolean
+  value: BigNumber
+}> {
+  const { transactor, contracts } = useContext(UserContext)
+  const { terminal, projectId } = useContext(ProjectContext)
+  const { userAddress } = useContext(NetworkContext)
```
peripheralist, line 16: Note: For simplicity many `-Tx()` hooks rely on contexts to derive `projectId`, `terminal`, and `userAddress`. This works for all current uses in the app, but will need to be considered if ever allowing the user to create these transactions from outside of the appropriate context.

johnnyd-eth:
Great idea to separate these out @peripheralist, way cleaner! Nothing stood out to me in need of change looking through the code and testing some transactions. LGTM

johnnyd-eth (approved):
Great idea to separate these out @peripheralist, way cleaner! Nothing stood out to me in need of change looking through the code and testing some transactions. LGTM

src/hooks/transactor/AddToBalanceTx.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export function useAddToBalanceTx(): TransactorInstance<{
+  value: BigNumber
+}> {
+  const { transactor, contracts } = useContext(UserContext)
+  const { projectId } = useContext(ProjectContext)
+
+  return ({ value }, txOpts) => {
+    if (!transactor || !projectId || !contracts?.TicketBooth) {
+      if (txOpts?.onDone) txOpts.onDone()
```
torvusbug, line 16: `txOpts?.onDone?.()` works just as well.

src/hooks/transactor/PayProjectTx.ts
torvusbug: @peripheralist My general solution to this is to create wrapper `useContext` hooks. One could look like this:

```typescript
export function useUserContext() {
  const userContext = useContext(UserContext)
  if (typeof userContext === 'undefined') {
    throw new Error('Attempting to use a UserContext outside of a UserContext provider')
  }
  
  return userContext
}
```

Despite the potentially catastrophic `throw new Error`, it should really only happen during development while testing your local changes. That code snippet won't work _exactly_ as-is in the codebase due to the default value currently setup in each Context. You'd have to create a module export for the default value and reference it in the `useUserContext` function (even if it's an empty object)

I take issue with a pattern I'm seeing negatively affect this code quality. It's out of scope of this PR but something I want anyone reading to actively think about. By setting every Context's default value to `{}` you've covered your bases erroneously. This default value creates unusual assumptions later, as seen in lines 19-25 with all these narrowing `||` combinations where you can't actually determine if the data is present or not. You also don't bail out of this function with a meaningful statement, like a console log (`tiny-invariant` is cool for this stuff, it also does type narrowing!) or `throw new Error`. While missing any pieces of this data may be incredibly unlikely, it's still not guaranteed even by the type system.

The only reasonable solution to me is to pass more data to the hook so it becomes _more_ of a pure function. Global session or user hooks tend to be unavoidable and guaranteed to be global; but a Project Context with bounds to a particular UX isn't. IMO, abstractions like this should be _given_ the data they need rather than _access_ the data they need, such as a project ID or the project's terminal version. In my own codebases, often I'd have a context sit above a UX but even single-use hooks receive an ID or an entire payload, just so it's a true abstraction of the UX flow.

This is just thinking out loud without much actionable in this PR's scope. I'm quite happy with how this is turning out, but I would encourage at _least_ a `console.error` before these `return Promise.reject(false)` statements.

aeolianeth:
Will take a look at this tomorrow morning (for me, ~9 hours from now) 🙂

peripheralist:
> My general solution to this is to create wrapper useContext hooks. One could look like this:

@torvusbug brilliant suggestion here, totally see what you're getting at. It is out of scope for this PR but I'll create an issue for addressing this separately

src/hooks/transactor/AddToBalanceTx.ts
peripheralist: updated. somehow after years of typescript i never knew that `fn.()` notation was a thing

peripheralist:
@aeolianeth have you had a chance to go through this yet?

src/components/modals/RedeemModal.tsx
```diff
@@ -83,27 +83,16 @@ export default function RedeemModal({
...
-        projectId.toHexString(),
-        redeemWad.toHexString(),
+    redeemTokensTx(
+      {
+        redeemAmount: parseWad(redeemAmount),
         minAmount,
-        userAddress,
-        false, // TODO preferconverted
-      ],
+        preferConverted: false, // TODO support in UI
```
aeolianeth, line 94: **question**: what does this do of interest?

src/components/modals/ReconfigureFCModal.tsx
```diff
@@ -184,70 +191,44 @@ export default function ReconfigureFCModal({
...
-        })),
-      ],
+    configureProjectTx(
+      {
+        fcProperties: {
+          target: editingFC.target,
+          currency: editingFC.currency,
+          duration: editingFC.duration,
+          discountRate: editingFC.discountRate,
+          cycleLimit: BigNumber.from(0), // TODO support in UI
```
aeolianeth, line 217: **question**: what does `cycleLimit` do?

aeolianeth:
> @aeolianeth have you had a chance to go through this yet?

@peripheralist apologies for the delay!

Really like the approach here. At a minimum, it improves grok-ability of the codebase substantially. I have been on the fence about context vs props, but seeing a PR like this convinces me that context is totally desirable in many cases 😂 .

Approving!

aeolianeth (approved):

src/hooks/transactor/PayProjectTx.ts
aeolianeth: Thanks for pointing this out @torvusbug - as someone still getting up to speed with React best practices, this is massively helpful and has prompted me to look into this more! ❤️ 

Personally I wouldn't feel safe throwing an error in a hook and not handling the exception. While it may only happen in development, I don't like leaving things like that open-ended 😅 . Might be projecting my paranoia!

I wonder what you think about the following approach that I seem to see a lot in react-world. Essentially we'd extend your example by returning an `error` property from the hook if context doesn't exist. Below is untested, just a loose thought:

```diff
export function useUserContext() {
  const userContext = useContext(UserContext)
  if (typeof userContext === 'undefined') {
-    throw new Error('Attempting to use a UserContext outside of a UserContext provider')
+    return { error: new Error('Attempting to use a UserContext outside of a UserContext provider') }
  }
  
  return userContext
}
```

Then, in each component, we check for the existence of `error`, and handle it appropriately.

cc @peripheralist @johnnyd-eth

src/components/modals/ReconfigureFCModal.tsx
peripheralist: it's one of the funding cycle properties that the UI has never made use of https://docs.juicebox.money/protocol-v1/ticketbooth/funding-cycles#write

src/components/modals/RedeemModal.tsx
peripheralist: converted is another word for "claimed/erc20". `preferConverted = true` means that if a user has both erc20 and unclaimed tokens, and redeems less than their entire token balance, the erc20 tokens will be redeemed first.

src/components/modals/ReconfigureFCModal.tsx
aeolianeth: Oh cool I see, so like a time-bound config that will auto-revert. I like.

src/components/modals/RedeemModal.tsx
aeolianeth: oh nice I see, thanks for the explanation. What does the behavior when `false`? Is it just the opposite (unclaimed is first-out)?

aeolianeth: Answered my own question: **yes** https://docs.juicebox.money/protocol/specifications/contracts/jbtokenstore/write/burnfrom

src/hooks/transactor/PayProjectTx.ts
torvusbug: @aeolianeth I disagree but not without caveats to my disagreement. Throwing inside a hook or component is a totally fine thing to do if we have error boundaries setup properly. Of course it's risky, but throwing inside a hook is not _always_ a scary thing to do (React Suspense actually works by throwing a Promise) . In this _particular_ case, I'm not afraid of throwing here (even in prod) because an error will be caught almost immediately by a developer testing their changes locally. Of course that's error prone, but I've actually never been bit in the ass by this.

Error boundary example placements: an error boundary above every modal/drawer (optimally, the Modal is an abstraction that contains an error boundary), an error boundary above every page and/or critical area. The fallback property can contain a "Retry" button that re-renders `children`. They're cheap to place around and will be leaned on more in React's future so it's probably a good idea to start placing them.

That said, I'm happy to move forward with your suggestion since adding error boundaries is out of scope of this PR.

aeolianeth: @torvusbug totally on board with error boundaries! Love me some. After working on https://github.com/jbx-protocol/juice-interface/pull/410 (V2 support PoC), I must say: when I was moving stuff around, I would have liked to have exceptions thrown when I was using things outside of context!

Without error boundaries, though, I'm still not sure it's the best approach, because we can't handle exceptions thrown from hooks (AFAIK). Not sure...

PR #384, opened 2022-01-17
Re-add Chinese to Selector and fix Lingui not translating FAQ's

johnnyd-eth:

1. 60% of Chinese now completed on Crowdin, so re-added it to the selector. 
2. Also fixes Lingui not translating the FAQ's section reliably. 
3. Fixes the resources dropdown being spaced differently to the other nav items 

## Acceptance checklist


peripheralist:
Nice fix, though it's not obvious to me why this works and the other way doesn't. Could you add a comment in the QAs file explaining why these are separated the way they are? Otherwise someone may just forget and restructure them into something breaking later down the road. It might also be helpful in case we see this same problem elsewhere

johnnyd-eth:
> Nice fix @johnnyd-eth , though it's not obvious to me why this works and the other way doesn't. Could you add a comment in the QAs file explaining why these are separated the way they are? Otherwise someone may just forget and restructure them into something breaking later down the road. It might also be helpful in case we see this same problem elsewhere

I added a note on why I did it @peripheralist, but I'm not sure the mechanics of why it fixed the issue. I guess the take-away is, if you're storing a list of t`` strings in a list, make sure you're not rendering it from the same file.

peripheralist (approved):
Got it, thanks for the added clarification!

PR #391, opened 2022-01-18
Add Spanish to language selector

johnnyd-eth:

- Adds Spanish to languages selector and lingui.js configuration
- Fixes bug when user's localStorage is setting to something not in the language selector. Was causing an error when opening language selector but now it defaults to English 
- Update Contribution.MD

## Acceptance checklist

### DIFF
--- a/src/constants/locale.ts
+++ b/src/constants/locale.ts
@@ -1,4 +1,4 @@
-export const SUPPORTED_LOCALES = ['en', 'zh', 'ru', 'tr', 'pt']
+export const SUPPORTED_LOCALES = ['en', 'zh', 'ru', 'tr', 'es']
 export type SupportedLocale = typeof SUPPORTED_LOCALES[number]
 
 export const DEFAULT_LOCALE: SupportedLocale = 'en'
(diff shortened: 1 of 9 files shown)


src/constants/locale.ts
```diff
@@ -1,4 +1,4 @@
-export const SUPPORTED_LOCALES = ['en', 'zh', 'ru', 'tr']
+export const SUPPORTED_LOCALES = ['en', 'zh', 'ru', 'tr', 'pt']
```
peripheralist, line 1: @johnnyd-eth looks like `pt` is commented out elsewhere and that this should be adding `es` instead of `pt`, but i see the branch is called add-portuguese lol. so i'm not sure if i know what's going on here

peripheralist (changes requested):

src/constants/locale.ts
johnnyd-eth: Oh damn my bad! Yeah I originally thought it was the Portuguese translations that were ready but it was actually the Spanish, so had to re-update these configurations to es

peripheralist (approved):

PR #396, opened 2022-01-18
Add Sushiswap price to project page

aeolianeth:

As per title.

| state | image|
| --- | --- |
| has sushi market | Screen Shot 2022-01-19 at 7 30 12 am |
| no sushi market | Screen Shot 2022-01-20 at 7 39 12 am |

## Acceptance checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,7 @@
     "@lingui/detect-locale": "^3.13.0",
     "@lingui/react": "^3.13.0",
     "@reduxjs/toolkit": "^1.6.2",
+    "@sushiswap/sdk": "5.0.0-canary.116",
     "@testing-library/jest-dom": "^5.11.4",
     "@testing-library/react": "^11.1.0",
     "@testing-library/user-event": "^12.1.10",
--- a/src/components/shared/AMMPrices/TokenAMMPriceRow.tsx
+++ b/src/components/shared/AMMPrices/TokenAMMPriceRow.tsx
@@ -7,7 +7,6 @@ import { CSSProperties } from 'react'
 import { formattedNum } from 'utils/formatNumber'
 import UniswapLogo from 'components/icons/Uniswap'
 import { t, Trans } from '@lingui/macro'
-import { Price, Token } from '@uniswap/sdk-core'
 import SushiswapLogo from 'components/icons/Sushiswap'
 import { Tooltip } from 'antd'
 
@@ -22,7 +21,7 @@ type Props = {
   exchangeName: ExchangeName
   tokenSymbol: string
   exchangeLink?: string
-  WETHPrice?: Price<Token, Token>
+  WETHPrice?: string
   loading?: boolean
   exchangeSupported?: boolean
   style?: CSSProperties
@@ -97,7 +96,7 @@ export default function TokenAMMPriceRow({
               target="_blank"
               style={{ fontWeight: 400 }}
             >
-              {`${formattedNum(WETHPrice.toFixed(0))} ${tokenSymbol}/ETH`}
+              {`${formattedNum(WETHPrice)} ${tokenSymbol}/ETH`}
               <LinkOutlined style={{ marginLeft: '0.2rem' }} />
             </a>
           </Tooltip>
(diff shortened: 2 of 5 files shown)


aeolianeth:
Hey @johnnyd-eth , still in draft, but would you mind reviewing this if you have a chance?

johnnyd-eth:
Nice @aeolianeth love these two features. Code all looks good to me. 

My only thoughts are: maybe something like 'Not supported', 'Unavailable', etc. would be easier to understand for most users than 'No liquidity'

johnnyd-eth (approved):
Nice @aeolianeth love these two features. Code all looks good to me.

My only thoughts (non-blocking) are: maybe something like 'Not supported', 'Unavailable', etc. would be easier to understand for most users than 'No liquidity'

aeolianeth:
> Nice @aeolianeth love these two features. Code all looks good to me.
> 
> My only thoughts (non-blocking) are: maybe something like 'Not supported', 'Unavailable', etc. would be easier to understand for most users than 'No liquidity'

@johnnyd-eth thanks for the review! Appreciate the UX feedback, too. `Unavailable` could be the go, actually. Will update 👍

aeolianeth:
...and with that, @peripheralist, this is ready for your review 🙏

package.json
```diff
@@ -7,6 +7,7 @@
     "@lingui/detect-locale": "^3.13.0",
     "@lingui/react": "^3.13.0",
     "@reduxjs/toolkit": "^1.6.2",
+    "@sushiswap/sdk": "^5.0.0-canary.116",
```
peripheralist, line 10: @aeolianeth are we sure we need to be using a canary build?

src/components/shared/AMMPrices/TokenAMMPriceRow.tsx
```diff
@@ -57,7 +56,7 @@ export default function TokenAMMPriceRow({
           {!exchangeSupported ? (
```
peripheralist, line 56: since we're no longer using `exchangeSupported` do you think we could remove it? i'm fine with leaving it in if we expect to use it again, but it felt like more of a patch to address the concerns from folks in the community who wanted Sushi specifically.

peripheralist:
awesome work @aeolianeth. super clean. couple easy questions first, but this is ready for main

src/components/shared/AMMPrices/TokenAMMPriceRow.tsx
aeolianeth: Yeah, good point, can remove 👍

package.json
aeolianeth: @peripheralist that's a great catch, I was meaning to check if there was a stable version.

Looks like there's a `5.0.0`. so I'll give it a whirl. https://www.npmjs.com/package/@sushiswap/sdk/v/5.0.0

aeolianeth: @peripheralist Ok, very strange... the `5.0.0` doesn't have any code in it 🙃 

`https://registry.yarnpkg.com/@sushiswap/sdk/-/sdk-5.0.0.tgz#7e33337eef7adba5a13716aba0bcdcb6a48bae22`

| 5.0.0 | 5.0.0-canary.116 |
| --- | --- |
| Screen Shot 2022-01-20 at 2 56 30 pm | Screen Shot 2022-01-20 at 2 56 38 pm |

The latest non-canary release was [almost a year ago](https://www.npmjs.com/package/@sushiswap/sdk/v/4.1.1) (`v4.1.1`), which also makes me a bit hesitant. Upon trying it, it's not quite compatible with the changes in this PR (breaking changes, makes sense), but given there's no docs specific to this version (or any version, really), it might take some time to re-implement with `v4.1.1`.

I've fixed the version at `5.0.0-canary.116` for now. Curiously, the `https://github.com/sushiswap/sushiswap-sdk` repo is archived now (must have only happened in the last few days, unless I missed it originally) which is... also strange.

I'd be happy to re-implement with `v4.1.1` if you feel strongly. Let me know 😊

peripheralist: strange... looking at their version history seems like they have a lot of canary versions, and those seem to have the most downloads. so maybe just odd versioning practice, or the sushi team just never does any testing and doesn't wanna be held accountable lol.

since it seems unlikely that bugs here would cause big issues in the app, let's just roll with the canary build for now

aeolianeth: @peripheralist  yeah lol the version history is quite confusing... I reached out in the Sushi Discord but nothing has come from that yet. Will keep an eye on it and make changes if necessary 👍

PR #392, opened 2022-01-18
Add Tally integration

0xzoz:
## What does this PR do and why?

Add support for [Tally](https://tally.cash/) wallet. Tally is an open-source browser extension similar to Metamask.

## Screenshots or screen recordings

N/A

## Acceptance checklist

- [x ] I have evaluated the [Approval Guidelines](../CONTRIBUTING.md#approval-guidelines) for this PR.
- [ x] I have tested this PR in [all supported browsers](../CONTRIBUTING.md#browser-support).

PR #397, opened 2022-01-19
Update lock file for bnc-onboard v1.36.0

aeolianeth:

https://github.com/jbx-protocol/juice-interface/pull/392/files upgraded the `bnc-onboard` dependency but didn't include updates to the `yarn.lock` file. This is causing a broken `main` (the Crowdin job is failing).

I ran `yarn install` in the main branch to generate this diff.

## Acceptance checklist

- [-] I have tested this PR in [all supported browsers](../CONTRIBUTING.md#browser-support).

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -89,7 +89,7 @@
     semver "^5.4.1"
     source-map "^0.5.0"
 
-"@babel/core@^7.1.0", "@babel/core@^7.12.3", "@babel/core@^7.4.4", "@babel/core@^7.7.5", "@babel/core@^7.8.4", "@babel/core@^7.9.0":
+"@babel/core@^7.1.0", "@babel/core@^7.12.3", "@babel/core@^7.7.5", "@babel/core@^7.8.4", "@babel/core@^7.9.0":
   version "7.15.5"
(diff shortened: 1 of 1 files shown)


aeolianeth:
Sweeping lock file changes like this are scary and can introduce security vulnerabilities. Might be worth investigating how we can verify such changes.

In the meantime, I encourage anyone to run `yarn install` from `main` and compare it to this diff to "verify" correctness.

peripheralist:
thanks for catching this @aeolianeth. big lockfile diffs are scary indeed. there shouldn't be such a big diff for only updating one package, but even running `yarn add bnc-onboard@1.36.0`seems to create the same diff, so i'm guessing we've somehow missed some lockfile updates with previous package bumps.

peripheralist (approved):
this seems okay to me.

aeolianeth:
@peripheralist Yeah, it's a weird one. I think the lockfile was in-sync previously, or else every `yarn install` would be trying to generate a new lockfile, which hasn't been the case for me at least.

It might be because the version is fixed at `1.36.0`, which means that any dependencies of `bnc-onboard` get updated throughout the lockfile.

In any case, I think if we add `yarn install --frozen-lockfile` in the CI somewhere for PRs, it will fail if there's a mismatch.

PR #401, opened 2022-01-19
Improve project card HTML semantics

johnnyd-eth:

Changed wrapping <div> on project cards to an <a> so we use it like a link (e.g. see the URL in bottom-left on hover, open in new tab, etc.)

**Vid shows two clicks:** first click (while holding Command/Control key) opens in new tab, second click without any keys down. Also see URL in bottom-right on hover

https://user-images.githubusercontent.com/96150256/150235381-bc5f4340-00af-4360-9704-7481da83a7a2.mp4

## Acceptance checklist

### DIFF
--- a/src/components/shared/ProjectCard.tsx
+++ b/src/components/shared/ProjectCard.tsx
@@ -46,8 +46,7 @@ export default function ProjectCard({
         overflow: 'hidden',
       }}
       key={project?.handle}
-      href={`/p/${project.handle}`}
-      onClick={() => (window.location.hash = '/p/' + project.handle)}
+      href={`/#/p/${project.handle}`}
     >
       {metadata ? (
         <div
--- a/src/components/shared/ProjectCard.tsx
+++ b/src/components/shared/ProjectCard.tsx
@@ -39,14 +39,14 @@ export default function ProjectCard({
   const terminalVersion = getTerminalVersion(project.terminal)
 
   return (
-    <div
+    <a
       style={{
         borderRadius: radii.lg,
         cursor: 'pointer',
         overflow: 'hidden',
       }}
-      className="clickable-border"
       key={project?.handle}
+      href={`/p/${project.handle}`}
       onClick={() => (window.location.hash = '/p/' + project.handle)}
     >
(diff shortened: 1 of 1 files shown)


src/components/shared/ProjectCard.tsx
```diff
@@ -39,14 +39,14 @@ export default function ProjectCard({
...
-    <div
+    <a
       style={{
         borderRadius: radii.lg,
         cursor: 'pointer',
         overflow: 'hidden',
       }}
-      className="clickable-border"
       key={project?.handle}
+      href={`/p/${project.handle}`}
```
aeolianeth, line 49: **suggestion**: to be safe, I think we need to include the `#`:

```suggestion
      href={`/#/p/${project.handle}`}
```

**suggestion (non-blocking)**: it might be nice to define a shared helper function to build a project path. This way we can use it in other places.

src/components/shared/ProjectCard.tsx
```diff
@@ -39,14 +39,14 @@ export default function ProjectCard({
...
       onClick={() => (window.location.hash = '/p/' + project.handle)}
```
aeolianeth, line 50: **issue**: I guess we can remove this line:
```suggestion
```

aeolianeth (changes requested):
@johnnyd-eth been praying for this change 😁 

I left a few comments. Could you please also upload a screenshot and demo video so we can ensure it works as expected? 🏓

peripheralist:
thank you!! also been wanting this change every time i cmd + click a project. not sure why they were ever built as <div/>s and not <a/>s.

looks good to me, i think aeolian's comments are the only things that need attention

src/components/shared/ProjectCard.tsx
johnnyd-eth: Thanks good call

johnnyd-eth:
Thanks @aeolianeth and @peripheralist I've addressed the comments 💯

aeolianeth (approved):

peripheralist (approved):
yes!

Issue #411, opened 2022-01-22
[BUG] Reserved token config state isn't cleared when clicking "Start Over"


aeolianeth:
## Summary

Reserved tokens slider (and possibly more) isn't cleared when clicking "start over" or deploying project.

## Steps to reproduce

1. On the create project form, set a reserved token allocation percentage.
2. Click "save"
3. Reload the page to confirm the state is persisted (to localstorage)
4. Click "start over"
5. Observe the reserved tokens slider. It remains set.
6. Refresh page and observe. Same.

## What is the _current_ bug behavior?

Reserved tokens slider (and possibly more) isn't cleared when clicking "start over" or deploying project

## What is the expected _correct_ behavior?

Everything should be cleared when clicking "start over" or deploying project.

## Relevant logs and/or screenshots

https://user-images.githubusercontent.com/94939382/150618807-b8030030-ce05-4c51-adbf-4870de17cc79.mp4

PR #421, opened 2022-01-26
Refactor redux localstorage

aeolianeth:

Fixes https://github.com/jbx-protocol/juice-interface/issues/411

Previously we were updating localStorage in a redux middleware. The problem with this approach is that redux middleware is invoked _before_ the reducer (between the `dispatch` call and the reducer), so the state in the middleware is a tick behind.

Using a redux subscription appears to be the preferred approach. This is invoked after the reducer, so localStorage is always in sync with the latest state.

This illustrates the difference:

```diff
- dispatch(resetState()) --> middleware (with old state) --> localstorage is set (with old state) --> reducer (state updated)
+ dispatch(resetState()) --> middleware (with old state) --> reducer (state updated) --> subscriber (with new state) --> localstorage is set (with new state)
```

This approach is described here: https://stackoverflow.com/questions/35305661/where-to-write-to-localstorage-in-a-redux-app.

https://user-images.githubusercontent.com/94939382/151093697-87347857-5a2f-481c-9f81-81be0127d783.mp4

## Acceptance checklist

### DIFF
--- a/src/redux/localStoragePreload.ts
+++ b/src/redux/localStoragePreload.ts
@@ -1,33 +1,14 @@
-import { Middleware } from '@reduxjs/toolkit'
-
 import { defaultProjectState } from './slices/editingProject'
 
-import { RootState } from './store'
-
-const KEY = 'jb_redux_preloadedState'
+import { RootState, REDUX_STATE_LOCALSTORAGE_KEY } from './store'
 
 interface PreloadedState {
   reduxState: RootState
 }
 
-export const localStoragePreloadMiddleware: Middleware =
-  store => next => action => {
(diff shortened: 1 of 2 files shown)


torvusbug (approved):
@aeolianeth I'm fine with this solution. I'd prefer it kept within the middleware but I don't feel strongly about that.

aeolianeth:
> @aeolianeth I'm fine with this solution. I'd prefer it kept within the middleware but I don't feel strongly about that.

awesome thank you!

@peripheralist over to you 🏀

peripheralist (approved):
love this! awesome work, and thanks for the clear PR description

PR #427, opened 2022-01-27
Improve 'Reconfigure funding' process

johnnyd-eth:

1. Changes Reconfigure funding button text from 'Reconfigure funding' to 'Reconfigure upcoming'
2. Moves button to top of the section, in place of tabs. Moves tabs below
3. Adds tooltip to button explaining that the current FC cannot be edited
4. Changes title of reconfigure FC modal to 'Reconfigure upcoming funding'

https://user-images.githubusercontent.com/96150256/151281961-75e71d1c-0023-4594-9497-d16069049062.mp4

## Acceptance checklist


src/components/modals/ReconfigureFCModal.tsx
```diff
@@ -308,16 +313,22 @@ export default function ReconfigureFCModal({
...
-            </Trans>
-          </p>
+        {currentFC?.duration.gt(0) ? (
+          <React.Fragment>
+            <h1 style={{ marginBottom: 20 }}>
+              <Trans>Reconfigure upcoming funding</Trans>
+            </h1>
+            <p>
+              <Trans>
+                All changes will be applied to the <b>upcoming</b> funding
```
aeolianeth, line 323: **suggestion (non-blocking)** - to ever so slightly improve HTML semtanics here, we could use a `<strong>` instead of `<b>`. [Here's a decent explanation why](https://stackoverflow.com/questions/271743/whats-the-difference-between-b-and-strong-i-and-em#:~:text=b%20or%20i%20means%20you,might%20use%20a%20different%20mapping.)

```suggestion
                All changes will be applied to the <strong>upcoming</strong> funding
```

aeolianeth (approved):
nice @johnnyd-eth LGTM! left a non-blocking suggestion, so approving 😊

src/components/Dashboard/FundingCycles.tsx
```diff
@@ -95,18 +95,21 @@ export default function FundingCycles({
           text={t`Funding cycle`}
           tip={t`A project's lifetime is defined in funding cycles. If a funding target is set, the project can withdraw no more than the target for the duration of the cycle.`}
           style={{
-            marginBottom: 12,
+            marginBottom: 30,
```
peripheralist, line 98: i think we talked about this in the Discord some but it may have gotten lost. this margin feels a bit too big to me, especially when the button isn't visible (for non-project owners). and for 99% of folks this button won't ever be on screen (note: in local development buttons like these are always visible for testing purposes, maybe we should change that)

peripheralist (changes requested):
looking really good! couple small thoughts

johnnyd-eth:
> looking really good! couple small thoughts

Thanks @peripheralist addressed and resolved :D

peripheralist (approved):
awesome!

PR #445, opened 2022-01-30
Improve i18n robustness

aeolianeth:

This PR fulfills some of the corrective actions revealed in the root-cause analysis of our recent downtime: https://www.notion.so/juicebox/RCA-production-downtime-2022-01-29-ca703708b20543b3825f8dbaf0c1db25

### Main changes
- **always import the default locale `locale/en/messages`.** The app will fail at compile-time if this file doesn't exist, so we will catch it in development/in the Fleek CI, instead of at run-time 😉 
- **adds exception handling to the `dynamicActivate` function.** If a `messages` file for a given locale doesn't exist, we will always fall back to the default locale, meaning the app will always render in English in the worst case.
- **update the `i18n:compile` npm script to _only_ run `lingui compile`** (and _not_ `lingui extract`). Extract generates the `en/messages.po` file, which we don't want to run locally; we want that to be the responsibility of the Crowdin GitHub action. `lingui compile` generates the `{lang}/messages.js` files, _which the app depends on_, and which we definitely need to run post-install (or pre-build).

### Supplementary changes
- adds missing `key` attributes to kill some lingering console errors.
- make exported function name `NavLanguageSelector` match its filename.

## Acceptance checklist

### DIFF
--- a/src/components/Landing/Payments.tsx
+++ b/src/components/Landing/Payments.tsx
@@ -24,6 +24,7 @@ export default function Payments() {
       'beneficiary',
       'note',
       'timestamp',
+      'id',
       { entity: 'project', keys: ['id'] },
     ],
     first: 20,
--- a/.github/workflows/crowdin-upload.yml
+++ b/.github/workflows/crowdin-upload.yml
@@ -21,7 +21,7 @@ jobs:
         run: yarn install --frozen-lockfile
 
       - name: Extract translations
-        run: 'yarn i18n:compile'
+        run: 'yarn i18n:extract'
 
       - name: Upload to Crowdin
         uses: crowdin/github-action@1.4.4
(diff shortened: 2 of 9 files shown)


src/components/Landing/Payments.tsx
```diff
@@ -24,6 +24,7 @@ export default function Payments() {
       'beneficiary',
       'note',
       'timestamp',
+      'id',
```
aeolianeth, line 27: we use `e.id` below as a `key`, but `id` wasn't defined previously.

johnnyd-eth (approved):
Beautiful detective work @aeolianeth, thanks for looking into this! All makes sense to me and looks fantastic

peripheralist (approved):
woooo really solid! more robust indeed, and thanks for the good explanation

PR #447, opened 2022-01-31
[BUG] Prompt wallet connect on pay project if no wallet connected

johnnyd-eth:

Prompts wallet connect on pay project if no wallet connected. 

Still not ideal, would be nice to not have to click the pay button twice, but I had trouble getting the usePayProjectTx() hook async so just went with this. 

https://user-images.githubusercontent.com/96150256/151794110-7524c6c3-8ae5-4ab1-a82b-41e220f3a4ba.mp4

## Acceptance checklist

### DIFF
--- a/src/hooks/transactor/PayProjectTx.ts
+++ b/src/hooks/transactor/PayProjectTx.ts
@@ -13,19 +13,26 @@ export function usePayProjectTx(): TransactorInstance<{
 }> {
   const { transactor, contracts } = useContext(UserContext)
   const { terminal, projectId } = useContext(ProjectContext)
-  const { userAddress } = useContext(NetworkContext)
+  const { userAddress, onSelectWallet } = useContext(NetworkContext)
+
+  const checkWalletConnected = () => {
+    if (!userAddress && onSelectWallet) {
+      onSelectWallet()
+    }
+  }
 
   return ({ note, preferUnstaked, value }, txOpts) => {
     if (
       !transactor ||
       !projectId ||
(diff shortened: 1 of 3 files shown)


src/hooks/transactor/PayProjectTx.ts
```diff
@@ -13,20 +13,27 @@ export function usePayProjectTx(): TransactorInstance<{
 }> {
   const { transactor, contracts } = useContext(UserContext)
   const { terminal, projectId } = useContext(ProjectContext)
-  const { userAddress } = useContext(NetworkContext)
+  const { userAddress, onSelectWallet } = useContext(NetworkContext)
+
+  const checkWalletConnected = () => {
```
peripheralist, line 18: @johnnyd-eth do you think we could just add this flow into the useTransactor hook itself? since any transaction in the app will need to have a connected wallet

johnnyd-eth: > @johnnyd-eth do you think we could just add this flow into the useTransactor hook itself? since any transaction in the app will need to have a connected wallet

Yep good idea, I'll do that

johnnyd-eth: @peripheralist Moved it into the useTransactor hook 👍

peripheralist (approved):
perfect! nice @johnnyd-eth !

i have some ideas for improving that conditional `if (userAddress) setLoading` pattern, but gonna address it in a different PR

PR #458, opened 2022-02-04
Subgraph hooks refactoring

peripheralist:

Better code organization, functionality unchanged.

Because subgraph queries can only return a maximum of 1000 items, getting _all_ items requires a sequence of queries. This adds a new `querySubgraphExhaustive` util which will run sequential queries until it gets back less than a full page size. It's currently used in DownloadParticipantsModal and BalanceTimeline.

Also consolidates duplicate logic from /hooks/SubgraphQuery.ts:`useSubgraphQuery` and /utils/graph.ts:`querySubgraph` into a single `querySubgraph` function in /utils/graph.ts

## Acceptance checklist

### DIFF
--- a/src/components/Dashboard/BalanceTimeline.tsx
+++ b/src/components/Dashboard/BalanceTimeline.tsx
@@ -5,8 +5,7 @@ import CurrencySymbol from 'components/shared/CurrencySymbol'
 import { V1ProjectContext } from 'contexts/v1/projectContext'
 import { ThemeContext } from 'contexts/themeContext'
 import EthDater from 'ethereum-block-by-date'
-import { parseProjectJson, Project } from 'models/subgraph-entities/project'
-import { parseTapEventJson } from 'models/subgraph-entities/tap-event'
+import { Project } from 'models/subgraph-entities/project'
 import moment from 'moment'
 import {
   CSSProperties,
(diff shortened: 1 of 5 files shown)


src/components/modals/DownloadParticipantsModal.tsx
```diff
@@ -156,11 +133,12 @@ export default function DownloadParticipantsModal({
           <Trans>Download CSV of {tokenSymbol || t`token`} holders</Trans>
         </h4>
 
-        <p style={{ padding: 10, background: colors.background.l1 }}>
-          {erc20IsUntracked && (
+        {erc20IsUntracked && (
```
peripheralist, line 136: little UI cleanup

src/components/Dashboard/BalanceTimeline.tsx
```diff
@@ -216,41 +212,35 @@ export default function BalanceTimeline({ height }: { height: number }) {
...
-                    ),
-                    timestamp: event.timestamp ?? 0,
-                  }
-                }),
-              )
-            }
-          },
-        )
+        await querySubgraph({
+          entity: 'tapEvent',
```
aeolianeth, line 216: **though (non-blocking)**: not related to this PR at all but something worth considering eventually:
I feel we should aim to move towards defining actual GraphQL queries rather than using a bespoke query builder. I think it'll help make our GraphQL queries easier to maintain and update.

Just something to keep in mind for one day 😊 

Example: https://github.com/tannerlinsley/react-query/blob/master/examples/basic-graphql-request/src/index.js#L48

src/components/modals/DownloadParticipantsModal.tsx
```diff
@@ -54,83 +52,62 @@ export default function DownloadParticipantsModal({
...
       )
-    }
+      document.body.appendChild(link)
 
-    query()
+      link.click()
+
+      setLoading(false)
+    } catch (e) {
+      console.log('Error downloading participants', e)
```
aeolianeth, line 108: **suggestion**: what's your opinion on `console.error` instead?

Side note: it might be worth considering some exception tracking tool like Sentry going forward.

src/utils/graph.ts
```diff
@@ -303,6 +281,71 @@ export function formatGraphResponse<E extends EntityKey>(
   return []
 }
 
+export async function querySubgraph<
```
aeolianeth, line 284: **praise**: returning a promise makes using this fn a lot nicer - sweet improvement 💯

aeolianeth (approved):
@peripheralist niiice improvement, I left a few side-notes here and there but will gladly approve this 😁 ✅

src/components/Dashboard/BalanceTimeline.tsx
peripheralist: nice idea, i'm open to this!

src/components/modals/DownloadParticipantsModal.tsx
peripheralist: i'm not sure why this error would ever actually get thrown, but down to use console.error

also down for considering sentry at some point

PR #460, opened 2022-02-04
Add trending projects

peripheralist:
# What does this PR do and why?

- New Trending tab on the Projects page
- New Trending section on the Landing page
- Replaces the ProjectsGrid component with a more generic Grid, layout functionality is unchanged

# Screenshots or screen recordings

image
image

# Acceptance checklist

- [x] I have evaluated the [Approval Guidelines](https://github.com/jbx-protocol/juice-interface/CONTRIBUTING.md#approval-guidelines) for this PR.
- [x] I have tested this PR in [all supported browsers](https://github.com/jbx-protocol/juice-interface/CONTRIBUTING.md#browser-support).


src/components/Dashboard/Rewards.tsx
```diff
@@ -96,7 +96,8 @@ export default function Rewards() {
               text={tokenSymbol ? tokenSymbol + ' ' + t`tokens` : t`Tokens`}
               tip={
                 `${tokenSymbol ? tokenSymbol + ' ' + t`ERC20` : t`Tokens`}` +
```
aeolianeth, line 98: **suggestion**: let's wrap this whole thing in a `t`

src/components/Landing/index.tsx
```diff
@@ -63,6 +80,32 @@ export default function Landing() {
     margin: '0 auto',
   }
 
+  const statsRowStyles: CSSProperties = {
+    borderTop: `1px solid var(--stroke-tertiary)`,
```
aeolianeth, line 84: **question**: can we use colors from `ThemeContext`? Might be worth grepping this PR for `var(` to find other areas we can use ThemeContext,too.

src/components/Landing/index.tsx
```diff
@@ -190,9 +233,20 @@ export default function Landing() {
               />
             </Col>
           </Row>
+          <Row
+            gutter={40}
+            align="middle"
+            style={{ ...statsRowStyles, display: 'none' }}
+          >
+            {statsCol('292', t`Projects powered by Juicebox`)}
+            {statsCol(`$94,137,431`, t`Raised on Juicebox`)}
```
aeolianeth, line 242: **question**: should these numbers be hard coded?

src/components/Landing/TrendingSection.tsx
```diff
@@ -0,0 +1,159 @@
...
+            className="hide-mobile hide-tablet"
+            style={{ height: 550, marginTop: 46 }}
+            src="/assets/green_orange.png"
+            alt="Green orange singing"
+          />
+        </Col>
+        <Col xs={24} lg={15}>
+          <div style={{ paddingBottom: 20 }}>
+            <h3 style={headingStyles}>
+              {/* Only way I could get the title wrapping properly on mobile was making each word an element
```
aeolianeth, line 135: **question**: how are we wanting it to wrap? It would be great to avoid this somehow 😅

src/components/Projects/index.tsx
```diff
@@ -141,148 +143,137 @@ export default function Projects() {
               and project configurations can vary widely. There are risks
               associated with interacting with all projects on the protocol.
               Projects built on the protocol are not endorsed or vetted by
-              JuiceboxDAO, so you should do your own research and understand the
-              risks before committing your funds.
+              JuiceboxDAO. Do your own research and understand the risks before
```
aeolianeth, line 146: yesss

src/components/Projects/ProjectsFilterAndSort.tsx
```diff
@@ -0,0 +1,153 @@
...
+        disabled={disabled}
+      />
+      <Trans>{label}</Trans>
+    </div>
+  )
+
+  // If includeArchived is true, we set active back to true
+  // to avoid both being unselected
+  const toggleArchived = () => {
+    if (includeArchived) {
```
aeolianeth, line 82: **suggestion**: could we simplify this block to:

```js
      setIncludeActive(includeArchived)
```

src/components/Projects/ProjectsFilterAndSort.tsx
```diff
@@ -0,0 +1,153 @@
...
+                e.stopPropagation()
+              }}
+            >
+              <FilterOutlined /> <label>Filter</label>
+            </span>
+          }
+        >
+          {/* onClick: Do not close collapse when clicking its items*/}
+          <div onClick={e => e.stopPropagation()}>
+            {filterCheckboxItem('V1', includeV1, setIncludeV1)}
```
aeolianeth, line 128: **suggestion**: I think as a general rule, components should be rendered with JSX and not called as a function like this: https://kentcdodds.com/blog/dont-call-a-react-function-component

```suggestion
            <FilterCheckboxItem label='V1' checked={includeV1} onChange={setIncludeV1} />
```

To do this, the argument for `FilterCheckboxItem` needs to be a single object: https://github.com/jbx-protocol/juice-interface/pull/460/files#diff-3428316285ddea1141ee3a5a4378d2df3d2a0f6d92d8b72e7a6a47ab029e57d5R56

aeolianeth (changes requested):
@johnnyd-eth @peripheralist sooo goood. I left a few comments here and there after a first pass, I'll probably take another look later tonight.

I also opened https://github.com/jbx-protocol/juice-interface/pull/463 (to be merged into this PR), which improves URL syncing with tabs/search. Let me know what you think :)

src/components/Dashboard/Rewards.tsx
johnnyd-eth: Remember the problem we were having with `t`s within `t`s ? Not a great solution but since I couldn't figure a way around this I left it like that

Screen Shot 2022-02-05 at 8 22 37 pm

src/components/Landing/index.tsx
johnnyd-eth: Isn't being shown at the moment. Just leaving it there for when we get the subgraph queries done.

src/components/Projects/ProjectsFilterAndSort.tsx
johnnyd-eth: Good call. Componentised it

src/components/Projects/ProjectsFilterAndSort.tsx
johnnyd-eth: Lmao what was I doing here

src/components/Landing/TrendingSection.tsx
johnnyd-eth: Only way I could think to avoid this happening:

https://user-images.githubusercontent.com/96150256/152640442-ed49f80a-c850-4a1a-9403-d4e5f3e9b813.mp4

This is how it works wrapping each word in its own element:

https://user-images.githubusercontent.com/96150256/152640506-fd8b1632-07b3-4201-bd06-136eac01eb3b.mp4

johnnyd-eth:
Screen Shot 2022-02-05 at 8 56 11 pm

src/components/Projects/ProjectsFilterAndSort.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+  // If includeArchived is true, we set active back to true
+  // to avoid both being unselected
+  const toggleArchived = () => {
+    setIncludeActive(includeArchived)
+    setIncludeArchived(!includeArchived)
+  }
+
+  // Close collapse when clicking anywhere in the window except the collapse items
+  window.addEventListener('click', () => setActiveKey(undefined), false)
```
torvusbug, line 58: This is _somewhat_ fine. Personally, I prefer something like a `useOutsideClick` hook that takes a ref to an element and binds a click event to the window where the click event checks the event's `path`. That way you don't have to worry about click event priority where clicking _within_ the opened menu actually closes it _before_ the menu item's click event. In _this particular_ case, React kinda saves you from this based on its synthetic click event priorities and your use of `stopPropagation`, but consider this unstable.

Another issue I have with this is React will continually bind a new event listener to the window _every time this component renders_. It's not a massive problem, but I consider it a bad practice. This code should look like this instead:

```suggestion
  // Close collapse when clicking anywhere in the window except the collapse items
  useEffect(() => {
    function handleClick() {
      setActiveKey(undefined)
    }

    window.addEventListener('click', handleClick)
    return () => window.removeEventListener('click', handleClick)
  }, [])
```

src/components/Projects/FilterCheckboxItem.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+export default function FilterCheckboxItem({
+  label,
+  checked,
+  onChange,
+  disabled,
+}: {
+  label: string
+  checked: boolean
+  onChange: Function
```
torvusbug, line 14: FWIW, I don't like typing functions as just a `Function`, especially when it's not _just_ a function; it has a parameter. I would prefer typing this as `(checked: boolean) => void`, that way callers know what to expect from this prop.

src/components/Projects/TrendingProjectCard.tsx
```diff
@@ -0,0 +1,213 @@
...
+            <div
+              style={{
+                fontWeight: 400,
+                color: colors.text.secondary,
+                fontSize: 13,
+                marginTop: 2,
+              }}
+            >
+              {formattedNum(project.trendingPaymentsCount)} payment
+              {project.trendingPaymentsCount > 1 ? 's' : ''}
```
torvusbug, line 193: Not super important, but remember 0 values are plural too. In all my React codebases, I like to keep a little function handy:

```typescript
function pluralize<S, P>(count: number, singular: S, plural: P) {
  if (count == 1) {
    return singular
  }

  return plural
}
```

That way, this code could look like:

```suggestion
              {formattedNum(project.trendingPaymentsCount)}
              {pluralize(
                project.trendingPaymentsCount,
                ' payment',
                ' payments',
              )}
```

src/components/Landing/index.tsx
```diff
@@ -211,13 +265,18 @@ export default function Landing() {
               {smallHeader(t`Projects using Juicebox`)}
               <div style={{ marginTop: 20 }}>
                 {previewProjects ? (
-                  <ProjectsGrid projects={previewProjects} list />
+                  <Grid
+                    children={previewProjects.map(p => (
+                      <ProjectCard project={p} />
+                    ))}
+                    list
+                  />
```
torvusbug, line 273: If you're going to use `children`, you should use it syntactically correct, ie:

```suggestion
                  <Grid list>
                    {previewProjects.map(p => (
                      <ProjectCard key={p.uri} project={p} />
                    ))}
                  </Grid>
```

src/components/Projects/MyProjects.tsx
```diff
@@ -0,0 +1,44 @@
...
+    theme: { colors },
+  } = useContext(ThemeContext)
+
+  return (
+    <React.Fragment>
+      {projects && projects.length ? (
+        <React.Fragment>
+          <Grid
+            children={projects?.map(p => <ProjectCard project={p} />) ?? <></>}
+          />
```
torvusbug, line 25: Same deal with `children` here. You also wouldn't need the ` ?? <></>`

torvusbug (changes requested):
Overall this looks pretty good. I have a handful of code changes.

src/constants/styles/layouts.ts
```diff
@@ -15,3 +15,5 @@ export const layouts: Record<string, CSSProperties> = {
     padding: padding.app,
   },
 }
+
+export const isMobile = window.innerWidth <= 900
```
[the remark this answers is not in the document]
johnnyd-eth: Ahh yep nice, didn't know about this one. Thanks!

src/components/Projects/TrendingProjectCard.tsx
johnnyd-eth: Great suggestion. Added

src/components/Landing/index.tsx
johnnyd-eth: For sure 👍

src/components/Projects/ProjectsFilterAndSort.tsx
johnnyd-eth: Awesome @torvusbug thanks for putting the change together and giving such a detailed explanation  🙌

src/components/Projects/FilterCheckboxItem.tsx
johnnyd-eth: Yep good call!

johnnyd-eth:
COPY FROM [THREAD](https://discord.com/channels/775859454780244028/935118596101779578/939497740721868870): I noticed a little bit of sketchiness with the load state when changing the trending window after @aeolianeth 's reload fix. We talked it over and agreed we just remove the option for now and merge with 7 days only.

johnnyd-eth:
> Overall this looks pretty good. I have a handful of code changes.

Thanks for your feedback @torvusbug ! Super helpful

src/components/Landing/index.tsx
```diff
@@ -63,6 +82,32 @@ export default function Landing() {
...
+  }
+
+  const statsColStyles: CSSProperties = {
+    textAlign: 'center',
+    marginBottom: 40,
+  }
+
+  const statsCol = (header: string, subHeader: string) => (
+    <Col xs={24} md={8} style={statsColStyles}>
+      {bigSubHeader(header)}
```
aeolianeth, line 97: **suggestion**: could we render this with JSX?

```suggestion
      <BigSubHeader header={header} />
```

src/components/Landing/index.tsx
aeolianeth: Ah k cool, in that case, I'd suggest we delete it for now to avoid the risk of it leaking out and rendering unexpectedly. Can always add it later 🤷

aeolianeth (approved):
@johnnyd-eth thanks for addressing those comments!

I left a couple more minor comments but happy to pre-approve for you 🚀

src/components/Projects/TrendingProjectCard.tsx
aeolianeth: @torvusbug @johnnyd-eth ideally we'd leverage the lingui JS `plural` helper for plurals, this way we ensure we're generating nice translatable plurals: https://lingui.js.org/guides/plurals.html

johnnyd-eth: Yeah great idea @aeolianeth. I've changed the new `pluralize` function to incorporate i18n plural functionality

torvusbug (approved):

PR #461, opened 2022-02-04
Project-specific metadata

lobovkin:

1. Updated Router, it uses project url "/p/projectname" and redirects from "/#/p/projectname".
2. Added metadata generator for top 100 (to be changed, 40 sec for 100) projects (saved in "/p/projectname" html file).
3. Added node-fetch dev dependency for metadata generator.
4. ipfs-404.html for not pre-generated projects to let url "/p/projectname" works.

Example: view-source:https://bafybeihyrsao4dbvnl4f5q7g5qroddzawbzr46hvnyt4u3nhqnlwwpw7su.on.fleek.co/p/assangedao

metadata example

## Acceptance checklist

- [*] I have evaluated the [Approval Guidelines](../CONTRIBUTING.md#approval-guidelines) for this PR.
- [*] I have tested this PR in [all supported browsers](../CONTRIBUTING.md#browser-support).


johnnyd-eth:
This is awesome man 🔥

generateMetadata.js
```diff
@@ -0,0 +1,70 @@
...
+      }
+    }
+    metadata.handle = handle;
+    let data = template;
+    for (let key in metadata) {
+      data = data.replaceAll('%' + key + '%', metadata[key]);
+    }
+    try {
+      console.log('Write ' + handle + ' metadata');
+      await fs.writeFile('./build/p/' + handle, data);
```
aeolianeth, line 64: **question**: should these files have a `.html` extension?

generateMetadata.js
```diff
@@ -0,0 +1,70 @@
...
+    <script>location.replace('/#' + location.pathname)</script>
+  </body>
+</html>`;
+
+(async () => {
+  let loaded = false;
+  let projects;
+  while (!loaded) {
+    try {
+      projects = (await (await fetch('https://gateway.thegraph.com/api/6a7675cd9c288a7b9571d5c9e78d5aff/deployments/id/QmNxBy8UnUsQr3aeBgFvdFyWbTSMiTGpJbkAJzSm5m6vYf', {
```
aeolianeth, line 34: We could add `dotenv` as a dependency here and use `REACT_APP_SUBGRAPH_URL`:

1. `yarn add -D dotenv`.
2. Add `require('dotenv').config()` to the top of this file.

```suggestion
      projects = (await (await projects = (await (await fetch(process.env.REACT_APP_SUBGRAPH_URL, {
```

aeolianeth (changes requested):
@lobovkin thanks so much for this contribution! I had no idea this approach was possible 🤯 .

There are some improvements I think we can make to speed the script up and a few clean ups here and there, but I don't want to block the progress of this PR - we can handle them in a follow-up :)

I'd probably like to see the `REACT_APP_SUBGRAPH_URL` environment variable leveraged. This way, we can sanity-check that the app still works on Rinkeby with the script generating metadata for Rinkeby projects.

Again, super thankful for your efforts here! 🚀

generateMetadata.js
lobovkin: Fleek works good without .html extension and serve as html file.
With .html extension url "/p/assangedao" doesn't serve "/p/assangedao.html" and project urls need to be changed to "/p/assangedao.html".
So it is better to keep files without .html extension

peripheralist:
Huge! Thanks @lobovkin !

Im afk for most of today but will check this out later tonight or tomorrow

generateMetadata.js
lobovkin: I added these changes, but there is "fleek/build — Build timeout", but I can't see Details, I have no access.
It looks like this REACT_APP_SUBGRAPH_URL variable is not available to build for rinkeby and the script is endlessly trying to get a list of projects.

peripheralist:
this is brilliant @lobovkin, i love it. couple thoughts:

1) I'm so curious how the project page can render without the /#/ in the url. Is this using some Fleek magic I don't know about?

2) I'd love to formalize the generation script a bit more. The ideal scenario would be to regenerate the html only for projects that have been edited (sent a `setURI` transaction), or created (sent a `deploy` transaction). 

The ideal design might be something like a chron job that listens for those on-chain events, runs the generation script for those created/updated projects, then submits a PR to add the generated html that we could merge frequently like we do with the current Crowdin projects. Though at a minimum, we might at least want to build html for all projects—could seem unfair for some projects to not have rich share links.

generateMetadata.js
peripheralist: Looks like this timeout was just due to some concurrent builds hogging our bots. that .env variable is available for rinkeby!

lobovkin:
> 1. I'm so curious how the project page can render without the /#/ in the url. Is this using some Fleek magic I don't know about?

I modified Router.tsx to render project pages without /#!/. Then user loads /p/projectname, Fleek serves a pre-generated page /p/projectname or default ipfs-404.html. This page redirects to index /#!/projectname and then Router remove /#!/ in address bar and render the page.

lobovkin:
> 2\. I'd love to formalize the generation script a bit more. The ideal scenario would be to regenerate the html only for projects that have been edited (sent a `setURI` transaction), or created (sent a `deploy` transaction).

As far as I can see, Fleek doesn't keep previous files and builds from a scratch, so it needs to re-generate all pages.

peripheralist:
> As far as I can see, Fleek doesn't keep previous files and builds from a scratch, so it needs to re-generate all pages.

Fleek doesn’t, though we could still git track the generated html no?

I’m down to get a version of this shipped without worrying about that, though obviously at some point running the script for all projects will take a prohibitively long time (and may also get costly). So it’d be nice to at least have an idea of how that could be solved later.

Also doing top 100 projects was awesome for a POC, but I think we should to be running this for all projects so no one gets excluded

wdyt @johnnyd-eth @aeolianeth @torvusbug ?

torvusbug (changes requested):
This looks really promising. I looked for a solution like this last year but the `ipfs-404.html` usage isn't in their docs, only some release notes awhile ago. I have a few gripes, but the most important one is removing the infinite `while` loops.

generateMetadata.js
```diff
@@ -0,0 +1,75 @@
...
+        headers: {'Content-Type': 'application/json'},
+      })).json()).data.projects;
+      loaded = true;
+    } catch (e) {
+      console.log('Retry loading projects in 1 sec');
+      await new Promise(resolve => setTimeout(resolve, 1000));
+    }
+  }
+  console.log(projects.length);
+  try { await mkdir(targetDir, {recursive: true}); } catch(e) {}
```
torvusbug, line 49: Didn't catch this upon first review, but `mkdir` isn't imported from anywhere, I think you meant `fs.mkdir`.

generateMetadata.js
```diff
@@ -0,0 +1,75 @@
...
+      data = data.replaceAll('%' + key + '%', metadata[key]);
+    }
+    try {
+      console.log('Write ' + handle + ' metadata');
+      await fs.writeFile(targetDir + handle, data);
+    } catch(e) {
+      console.log(e);
+      console.log('Write file ' + handle + ' eror')
+    }
+  }
```
torvusbug, line 74: @lobovkin If the above while loop fails to retrieve metadata, this will surely throw. Probably a good idea to add this whole section into `if (metadata != null)`

generateMetadata.js
```diff
@@ -0,0 +1,75 @@
...
+    <meta name="twitter:site" content="@%twitter%" />
+    <meta name="twitter:creator" content="@%twitter%" />
+    <meta name="twitter:title" content="%name%" />
+    <meta name="twitter:description" content="%description%" />
+    <meta name="twitter:image" content="%logoUri%" />
+    <title>%name%</title>
+  </head>
+  <body>
+    <noscript>You need to enable JavaScript to run this app.</noscript>
+    <script>location.replace('/#' + location.pathname)</script>
```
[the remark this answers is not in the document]
lobovkin: This is necessary to redirect the user from the metadata page to the application page (to index.html). There is no React in it, just stub pages for displaying metadata.

src/Router.tsx
```diff
@@ -10,29 +18,48 @@ function CatchallRedirect() {
...
-        <Route path="/p/:handle">
-          <Dashboard />
-        </Route>
-        <Route path="/:route">
-          <CatchallRedirect />
-        </Route>
-      </Switch>
-    </HashRouter>
+    <>
+      <ReactRouter history={history}>
```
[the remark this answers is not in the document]
lobovkin: This is taken into account in the code, the old links with the hash continue to work.
Non-hash router is only used for project links, there is still hash router for other links:
```
<Route exact path="/">
            <HashRouter>
```
And this allows old links to projects with a hash to work:
```
function CatchallRedirect() {
  const route = useParams<{ route: string }>()['route']
  return <Redirect to={'/p/' + route} />
}
```

generateMetadata.js
lobovkin: Yes, thank you.

generateMetadata.js
```diff
@@ -0,0 +1,75 @@
...
+  </body>
+</html>`;
+const targetDir = './build/p/';
+
+(async () => {
+  let loaded = false;
+  let projects;
+  while (!loaded) {
+    try {
+      process.env.REACT_APP_SUBGRAPH_URL = 'https://gateway.thegraph.com/api/6a7675cd9c288a7b9571d5c9e78d5aff/deployments/id/QmNxBy8UnUsQr3aeBgFvdFyWbTSMiTGpJbkAJzSm5m6vYf';
```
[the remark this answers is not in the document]
lobovkin: Thanks, I forgot to delete it.

generateMetadata.js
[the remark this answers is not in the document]
torvusbug: @lobovkin I see, that makes sense to me.

public/p/ipfs-404.html
```diff
@@ -0,0 +1,39 @@
+<!DOCTYPE html>
```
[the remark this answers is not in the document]
torvusbug: @lobovkin I take this back. I didn't fully understand what was going on and I see the difference between these files now.

peripheralist:
closing this for now due to concerns over scaling.

PR #480, opened 2022-02-09
Update subgraph entities

### LINKED ISSUE #470: [IDEA] Update subgraph entities with new data available with jb subgraph v2.0.1
## Problem to solve

With the launch of the Juicebox v2.0.1 subgraph, some entities were added/updated. Schema available here https://thegraph.com/explorer/subgraph?id=0x63a2368f4b509438ca90186cb1c15156713d5834-0&view=Playground

## Proposal

Update the subgraph models to include this new data.

peripheralist:

- Adds previously unsupported subgraph entities `DeployedERC20Event` and `PrintPremineEvent`, and adds some missing properties to existing entities
- Makes type definitions for entities more uniform by removing conditional? properties and using `Partial<>`s

## Acceptance checklist

Fixes #470

### DIFF
--- a/src/hooks/v1/Projects.ts
+++ b/src/hooks/v1/Projects.ts
@@ -178,7 +178,7 @@ export function useTrendingProjects(count: number, days: number) {
       setProjectStats(
         (payments ?? []).reduce(
           (acc, curr) => {
-            const projectId = curr.project?.toString()
+            const projectId = curr.project.id?.toString()
 
             return projectId
               ? {
@@ -289,17 +289,14 @@ export function useHoldingsProjectsQuery(wallet: string | undefined) {
 
       // Reduce list of paid project ids
       setProjectIds(
-        participants?.reduce(
-          (acc, curr) => [
+        participants?.reduce((acc, curr) => {
+          const projectId = curr?.project?.id?.toString()
+
+          return [
             ...acc,
-            ...(curr.project
-              ? acc.includes(curr.project.toString())
-                ? []
-                : [curr.project.toString()]
-              : []),
-          ],
-          [] as string[],
-        ),
+            ...(projectId ? (acc.includes(projectId) ? [] : [projectId]) : []),
+          ]
+        }, [] as string[]),
       )
 
       setLoadingParticipants(false)
--- a/src/models/subgraph-entities/distribute-to-payout-mod-event.ts
+++ b/src/models/subgraph-entities/distribute-to-payout-mod-event.ts
@@ -1,34 +1,39 @@
 import { BigNumber } from 'ethers'
 
+import { parseProjectJson } from 'models/subgraph-entities/project'
+
+import { Project, ProjectJson } from './project'
+
 export interface DistributeToPayoutModEvent {
   id: string
-  project?: BigNumber
-  fundingCycleId?: BigNumber
-  projectId?: BigNumber
+  project: Partial<Project>
+  fundingCycleId: BigNumber
+  projectId: BigNumber
   modBeneficiary: string
   modPreferUnstaked: boolean
-  modProjectId?: BigNumber
+  modProjectId: BigNumber
   modAllocator: string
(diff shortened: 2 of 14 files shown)


src/components/Dashboard/ProjectActivity/ReservesEventElem.tsx
```diff
@@ -14,7 +14,18 @@ import { smallHeaderStyle } from '../styles'
 export default function ReservesEventElem({
   printReservesEvent,
 }: {
-  printReservesEvent: PrintReservesEvent | undefined
+  printReservesEvent:
```
peripheralist, line 17: stricter type checking should hopefully make it easier to catch missing properties

src/hooks/v1/Projects.ts
```diff
@@ -178,7 +178,7 @@ export function useTrendingProjects(count: number, days: number) {
       setProjectStats(
         (payments ?? []).reduce(
           (acc, curr) => {
-            const projectId = curr.project?.toString()
+            const projectId = curr.project.id?.toString()
```
peripheralist, line 181: now we can get the whole `project` object if we like, not just the `id`

src/models/subgraph-entities/distribute-to-payout-mod-event.ts
```diff
@@ -1,34 +1,39 @@
...
 export const parseDistributeToPayoutModEvent = (
   json: DistributeToPayoutModEventJson,
-): DistributeToPayoutModEvent => ({
+): Partial<DistributeToPayoutModEvent> => ({
   ...json,
   fundingCycleId: json.fundingCycleId
     ? BigNumber.from(json.fundingCycleId)
     : undefined,
-  project: json.project ? BigNumber.from(json.project) : undefined,
+  project: json.project ? parseProjectJson(json.project) : undefined,
```
peripheralist, line 36: now we can get the whole `project` object if we like, not just the id

peripheralist:
okay, leaving the other subgraph entity typing alone for now.

torvusbug (approved):
This looks pretty good. I know we can't solve the typing issue which is why you're leaning on `Partial`.

Out of scope here, but long-term I wonder if we can go with something perhaps radical... Like using [a GraphQL schema generator](https://github.com/dotansimha/graphql-code-generator) to create our interfaces with the subgraph. Completely dev-only TypeScript definitions (I suppose we could use a build process to validate graph usage, but don't commit the .d.ts). I really don't like code generation, but it might be better than misleading types.

johnnyd-eth (approved):
These changes look great Peri love it, thanks!! LGTM

peripheralist:
> This looks pretty good. I know we can't solve the typing issue which is why you're leaning on `Partial`.
> 
> Out of scope here, but long-term I wonder if we can go with something perhaps radical... Like using [a GraphQL schema generator](https://github.com/dotansimha/graphql-code-generator) to create our interfaces with the subgraph. Completely dev-only TypeScript definitions (I suppose we could use a build process to validate graph usage, but don't commit the .d.ts). I really don't like code generation, but it might be better than misleading types. They have a [live demo view](https://www.graphql-code-generator.com/#live-demo)

Cool idea, and I think @aeolianeth was talking about something similar as well?

Right now the app uses some runtime-computed graph queries for things like projects search and the volume/balance graph. Would this work with those?

aeolianeth:
Nice improvement @peripheralist , thank you!

We could look at type generation from our graphql schema. I will say that in other graphQL codebases I've worked in, queries are always defined in `.graphql` files which is much more flexible and maintainable. The schema would be generated from our subgraph definition (maybe it already is, haven't checked), and this schema can be used by a linter to validate query properties.

Frankly I don't think we have the capacity for a full overhaul of our graphql strategy right now, but it's certainly something I'm keen on improving eventually.

@peripheralist happy to approve, just a couple of conflicts to resolve.

aeolianeth (approved):

PR #503, opened 2022-02-10
New feedback form prompting modal after project deploy

johnnyd-eth:

- Displays a new modal that prompts user to give feedback after they deployed a project.

Code changes:
- Adds `?newDeploy=true` when redirecting to the project's page after it is deployed, which displays the modal. This search query is removed from the URL when the modal is closed.
- Changes the component name of the feedback form link (stoned Banny in bottom right).
- Adds function in `utils` to generate feedback form URL

After deploy, user redirected to `/#/p/<handle>?new_deploy=true` which opens modal:

Screen Shot 2022-02-10 at 9 46 57 am

Deploying projects is bugging on Rinkeby right now so I can't attach a good video of the entire process. But I've tested that after deployment user is redirected to the correct URL, and when adding the `?new_deploy=true` to URLs of already existing Rinkeby projects, the modal behaves as expected. So I'm sure it will work. 

## Acceptance checklist


peripheralist:
love that modal @johnnyd-eth! link to the discord is awesome.

do you think we could also use the `newDeploy=true` to avoid that "project not found" that occasionally shows up when the redirect happens before the newly deployed project can be read?

johnnyd-eth:
> love that modal @johnnyd-eth! link to the discord is awesome.
> 
> do you think we could also use the `newDeploy=true` to avoid that "project not found" that occasionally shows up when the redirect happens before the newly deployed project can be read?

Yep the modal does show up in that case. I've also added a new state for when the user closes the modal (so they still don't see the 'not found' message). Note: when you click that reload button it redirects to the project without the `newDeploy` in the URL, so that the modal won't show up again for them. Here's how it would look (and hopefully the project would've actually loaded by the time they click refresh and they'll see it lol):

https://user-images.githubusercontent.com/96150256/153405801-e1b0ce35-3834-4229-8bf1-231be74797f8.mp4

src/components/Dashboard/index.tsx
```diff
@@ -156,9 +166,51 @@ export default function Dashboard() {
...
+          style={{
+            padding: padding.app,
+            height: '100%',
+            ...layouts.centered,
+            textAlign: 'center',
+          }}
+        >
+          <h2>
+            <Trans>
+              {handle} is still pending deployment, please try refreshing
```
aeolianeth, line 194:
```suggestion
              {handle} is still pending deployment. Try refreshing the page
```

src/components/Project/modals/FeedbackPromptModal.tsx
```diff
@@ -0,0 +1,59 @@
...
+        <Trans>
+          All the feedback will be sent to a{' '}
+          <a
+            href="https://discord.gg/NWmBtAHq"
+            rel="noreferrer"
+            target="_blank"
+          >
+            public Discord channel{' '}
+          </a>
+          and looked at by our development team.
```
aeolianeth, line 54:
```suggestion
          for our development team to consider.
```

src/utils/feedbackFormURL.ts
```diff
@@ -0,0 +1,16 @@
+export const feedbackFormURL = (
+  referrer: 'deploy' | 'stoned-banny',
```
aeolianeth, line 2: **praise**: great utility function!

**suggestion (non-blocking)**: with utils like these it can sometimes be nicer to define a single object as the parameter. That way when a developer uses the function, they can see at a glance what a given parameter is doing.

This:
```js
feedbackFormURL('stoned-banny', 'johnnyDAO', '0x00')
```
vs this
```js
feedbackFormURL({
  referrer: 'stoned-banny',
  projectHandle: 'johnnyDAO',
  userAddress, '0x00',
})
```

src/utils/feedbackFormURL.ts
```diff
@@ -0,0 +1,16 @@
...
+  projectHandle?: string,
+  userAddress?: string,
+) => {
+  let url = `https://auditor.typeform.com/to/REMUTIbQ#`
```
aeolianeth, line 6: **suggestion**: it might be helpful to use the `URLSearchParams` api for building the query string:

https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/URLSearchParams#examples

src/components/Project/modals/FeedbackPromptModal.tsx
```diff
@@ -0,0 +1,59 @@
...
+      </p>
+      <p>
+        <Trans>
+          We are keen to hear how we can improve Juicebox for you, can you share
+          with us some feedback while your memory is still fresh?
+        </Trans>
+      </p>
+      <p>
+        <Trans>
+          All the feedback will be sent to a{' '}
```
aeolianeth, line 46:
```suggestion
          Your feedback will be sent to a{' '}
```

src/components/Project/modals/FeedbackPromptModal.tsx
```diff
@@ -0,0 +1,59 @@
...
+      okText={t`Give feedback on project creation`}
+      width={600}
+      centered
+    >
+      <p>
+        <Trans>Congrats on creating your project!</Trans>
+      </p>
+      <p>
+        <Trans>
+          We are keen to hear how we can improve Juicebox for you, can you share
```
aeolianeth, line 40:
```suggestion
How did we do? We'd love to get your feedback so we can make creating Juicebox projects as juicy as possible!
```

aeolianeth (changes requested):
@johnnyd-eth nice work! left a few comments

src/components/Dashboard/index.tsx
johnnyd-eth: Nice

src/components/Project/modals/FeedbackPromptModal.tsx
johnnyd-eth: Like that

src/utils/feedbackFormURL.ts
johnnyd-eth: Ahh yep good tip. I'd make a note of that from now on

src/utils/feedbackFormURL.ts
johnnyd-eth: Nice suggestion. Implemented

src/components/Dashboard/index.tsx
```diff
@@ -156,9 +166,51 @@ export default function Dashboard() {
     terminalAddress,
   ])
 
+  // Close feedback modal and search query from URL
+  const closeFeedbackModal = () => {
+    window.history.replaceState({}, document.title, `/#/p/${handle}`)
```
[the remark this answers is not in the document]
johnnyd-eth: Yep good cool

src/components/Project/modals/FeedbackPromptModal.tsx
johnnyd-eth: > @johnnyd-eth i like that strategy! still feels a little hacky, but much better than before.
> 
> maybe that "refresh" button can refresh the page without removing the `new_deploy=true` param?

The only reason I'm removing it is that in the case we keep it in the URL, the user will have to close the modal that they already just closed (possibly multiple times), which may seem quite sketchy.

johnnyd-eth:
@aeolianeth @peripheralist thanks heaps for your feedback! I've made a few further changes

> @johnnyd-eth i like that strategy! still feels a little hacky, but much better than before.
> 
> maybe that "refresh" button can refresh the page without removing the `new_deploy=true` param?

The only reason I was removing it is that in the case we keep it in the URL, when the user refreshes they'll have to close the modal that they already just closed (possibly doing this multiple times), which may seem quite sketchy. I decided to add another search param `feedbackModalOpen` which is removed on the 'refresh', so that in the case the user clicks refresh and their project is still not available, they will still see the 'not available' rather than the 'not found' message, while also not having the modal popup again. Here's how it looks:

https://user-images.githubusercontent.com/96150256/153685450-4c99ca2d-ce17-47e6-b69e-a287c9c4da8d.mp4

peripheralist:
> The only reason I was removing it is that in the case we keep it in the URL, when the user refreshes they'll have to close the modal that they already just closed

Ah makes sense @johnnyd-eth, good thinking. I like this! Not sure if this is just something with your localhost build, but seems like there's no Loading spinner when Refresh is clicked, which feels a little odd. I'm wondering if the "refresh" action isn't resetting the value of ProjectID?

.github/workflows/lint.yml
```diff
@@ -0,0 +1,19 @@
+name: Lint
```
peripheralist, line 1: @johnnyd-eth was this intended to be included in this PR? not sure if i'm missing something

peripheralist (approved):
awesome work @johnnyd-eth thanks for addressing all the feedback!

curious about that Lint.yml action, but approving for now

.github/workflows/lint.yml
johnnyd-eth: Yep that got in here after I rebased and squashed somehow. I've removed it

aeolianeth: @johnnyd-eth looks like its still there 🤔

peripheralist:
loving it @johnnyd-eth , let's ship. @aeolianeth it says you've got some open comments still. looks like they're all addressed, but i'll leave for you to merge just in case 👌

aeolianeth (approved):
@johnnyd-eth @peripheralist thank you!

Looks like those `lint` workflow changes are just whitespace/formatting, so I'll happily approve and merge!

PR #509, opened 2022-02-10
Make "Claim token as ERC-20" btn a primary btn

johnnyd-eth:

Make "Claim token as ERC-20" btn (in "Manage") modal a primary btn to differentiate from "Burn tokens"

Screen Shot 2022-02-10 at 8 24 08 pm

Screen Shot 2022-02-10 at 8 25 37 pm

## Acceptance checklist

### DIFF
--- a/src/components/Project/Rewards/index.tsx
+++ b/src/components/Project/Rewards/index.tsx
@@ -227,7 +227,11 @@ export default function Rewards() {
               </Button>
             </React.Fragment>
           )}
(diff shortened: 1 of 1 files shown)


peripheralist:
mentioned this in Discord, but my only hesitation here is that it might make "Claim" look like something that users _should_ be doing, vs a very optional thing.

if the goal is just to better differentiate buttons, maybe we could consider adding some icons inline with the text?

PR #517, opened 2022-02-11
Make util functions to replace `tokenSymbol ?? "token"` all over the codebase

johnnyd-eth:

The `tokenSymbol ?? "token"` and its 3 other counterparts weren't being translated properly. I made functions for the 4 different variations in `/utils` and implemented them throughout the codebase. This shouldn't affect anything in the UI, except that a bunch of 'token' and 'tokens' text will now translate properly. 

Apologies for the branch name, initially planned to do the redeem rate slider, then to do these changes and the slider together, but then thought it'd be best to get these changes out in a single PR. 

## Acceptance checklist

### DIFF
--- a/src/utils/tokenSymbolText.ts
+++ b/src/utils/tokenSymbolText.ts
@@ -1,36 +1,24 @@
 import { t } from '@lingui/macro'
 
 // Function to return {tokenSymbol} and/or 'tokens', translated and (possibly) capitalized
-export const tokenSymbolText = (
-  tokenSymbol?: string,
-  capitalize?: boolean,
-  plural?: boolean,
-  includeTokenWord?: boolean,
-) => {
-  // e.g. 'JBX' or 'Token'
-  if (!includeTokenWord) {
-    if (!plural) {
-      return capitalize ? tokenSymbol ?? t`Token` : tokenSymbol ?? t`token`
-    } else {
-      return capitalize ? tokenSymbol ?? t`Tokens` : tokenSymbol ?? t`tokens`
-    }
+export const tokenSymbolText = ({
+  tokenSymbol,
+  capitalize,
+  plural,
+  includeTokenWord,
+}: {
+  tokenSymbol?: string
+  capitalize?: boolean
+  plural?: boolean
+  includeTokenWord?: boolean
+}) => {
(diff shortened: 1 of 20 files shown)


peripheralist:
Beautiful! I hadn't done this before only because it seemed complicated to handle the plural/capital/long/short cases, but I think these util functions work nicely 👏

src/utils/tokenSymbolText.ts
```diff
@@ -0,0 +1,27 @@
+import { t } from '@lingui/macro'
+
+// Functions to return {tokenSymbol} and/or 'tokens', translated and (possibly) capitalized
+
+export const singleTokenShort = (tokenSymbol?: string, capitalize?: boolean) =>
```
aeolianeth, line 5: **suggestion** what do you think about combining all of these functions into one, with this signature:

```js
export const tokenSymbolText = ({
  tokenSymbol,
  capitalize,
  plural,
  includeTokenSymbol,
}: {
  tokenSymbol?: string
  capitalize?: boolean
  plural?: boolean
  includeTokenSymbol?: boolean
}) => {
  // todo
}
```

In here, I would use as many `if` statements as needed, instead of ternary statements. Nested ternary's get a bit unwieldy and can be harder to follow. If the function is getting too big, we can create smaller functions to handle specific logic, but still only expose this single function.

Let me know what you think 🙇

aeolianeth (changes requested):
@johnnyd-eth thanks so much for this PR! A much needed improvement.

I left one suggestion about an alternative approach to this, where we only expose a single function. Happy to discuss more!

src/utils/tokenSymbolText.ts
johnnyd-eth: Yeah it's a great idea. Changed it up, thanks!

johnnyd-eth: Doneski!

aeolianeth (approved):
Thank you for making those changes @johnnyd-eth! I've given this a smoke test on the fleek preview and it looks good to me.

Approving and merging 🚀 

cc @peripheralist

aeolianeth:
Looks like there was some conflicts with translations. Just pushed a rebase with `main` 👍

PR #525, opened 2022-02-12
[BUG] Fix distribution form amount / percent not updating together

johnnyd-eth:

Put state back in ProjectPayoutMods to fix amount-input/percent-slider not updating together. @peripheralist I know it's not ideal using state and form together and totally understand why you thought to remove it, but this was the only way I could get them linked up and responding properly. 

https://user-images.githubusercontent.com/96150256/153690438-430b53da-6bf2-4be1-9757-af1a6ace4a21.mp4

## Acceptance checklist


peripheralist:
nice @johnnyd-eth! great catch, hadn't noticed this bug.

yeah, i don't love the form + state pattern but we're at least relatively consistent with how it's used across the app. makes sense to use it here for this fix

one request—should we also `setEditingPercent(0)` at the end of the `setReceiver()` function?

also, plz take out that console.log() in `setReceiver()` (i think that was my bad 🙈)

johnnyd-eth:
Yep I'm just afk for next 2hrs then will do

peripheralist (approved):
thanks @johnnyd-eth !

PR #527, opened 2022-02-12
Save ~400KB in javascript on `home` and `project` page

aeolianeth:

The PR updates the contract loader to use esmodule `import` syntax instead of `require`. This helps webpack split the bundle better. The overall sum of the bundle size remains the same, but the distribution JS across bundles is improved, meaning that there's more chance of _less_ JS being transferred for a given page.

The end result is that we save ~400KB in transferred JS on the home and project page 🎉 

To achieve this, the `terminal` utils file has been removed and replaced with hooks that rely on the already-loaded `contracts` from `usercontext`.

In the `before` screenshot, the first chunk is quite large (`6.1MB`), but using `import`, our biggest chunk is cut in half (`3.09MB`). The overall bundle size remains the same (`16.31MB`).

| before | after |
| --- | --- |
| Screen Shot 2022-02-12 at 9 00 16 am | Screen Shot 2022-02-12 at 10 54 18 am |

### Home page and project page metrics

These a screenshots from the network tab, filtering for `JS` files. Number of requests increases (33 -> 42), but total transferred JS decreases (5.8MB -> 5.4MB).

| page | Before | After |
| --- | --- | --- |
| home | main-home | branch-home |
| JB project page | Screen Shot 2022-02-12 at 11 31 30 am | branch-jb-project |

| metric | Before | After |
| --- | --- | --- |
| JS file request count | 33 | 42 |
| JS file total transferred size | 5.8MB | 5.4MB |

## Acceptance checklist


src/hooks/v1/TerminalVersion.ts
```diff
@@ -0,0 +1,64 @@
...
+      address.toLowerCase() === contracts.TerminalV1_1.address.toLowerCase()
+    ) {
+      setName(V1ContractName.TerminalV1_1)
+    }
+  }, [address, contracts])
+
+  return name
+}
+
+export function useTerminalVersion(address?: string) {
```
peripheralist, line 46: nice idea making these hooks! maybe for organization let's either split these into three files, or just rename the file something like /v1/terminal.ts?

src/hooks/v1/ContractLoader.ts
```diff
@@ -54,8 +54,22 @@ const loadContract = async (
...
-  let contract = await import(
-    `@jbx-protocol/contracts-v1/deployments/${network}/${contractName}.json`
-  )
+  /**
+   * Conditionally load contracts based on configured network.
+   * This prevents ABIs for unused networks being imported to our JS bundle.
+   */
+  const _loadContract = async () => {
+    if (network === NetworkName.mainnet) {
+      return await import(
```
aeolianeth, line 63: something not working, sec.

peripheralist: alternatively if there's any trouble with this, we could try something like:

```
import * as mainnetV1Contracts from '@jbx-protocol/contracts-v1/mainnet/...'
import * as rinkebyV1Contracts from '@jbx-protocol/contracts-v1/rinkeby/...'
...
if (process.env.REACT_APP_INFURA_NETWORK === 'mainnet') {
  return mainnetV1Contracts
}
```

aeolianeth: Yeah unfortunately we can't quite do that because that directory doens't have an `index.js` file

Screen Shot 2022-02-12 at 12 45 04 pm

The boring fix would be to import each `.json` file explicitly 🤷

peripheralist: if we need to, we could request some architecture tweaks to the contracts package. though if your original method is working then no worries

aeolianeth: @peripheralist Bundle size with this change: 6.1MB => 4.15MB 😁 

Weirdly I don't see a difference in the total transferred size in development...not sure why.

Screen Shot 2022-02-12 at 12 48 13 pm

src/hooks/v1/TerminalVersion.ts
aeolianeth: Yep, 💯 agree

src/hooks/v1/ContractLoader.ts
peripheralist: > @peripheralist Bundle size with this change: 6.1MB => 4.15MB 😁
> 
> Weirdly I don't see a difference in the total transferred size in development...not sure why.
> 
> Screen Shot 2022-02-12 at 12 48 13 pm

sweet! where are you reading the 4.15?

peripheralist: @aeolianeth sweet! where are you reading the 4.15 value?

aeolianeth: @peripheralist wow...sorry, it's 16.1 => 14.15. Was missing the `1` 😅

peripheralist: hahaha it's a thicc app

aeolianeth:
After some discussion in [Discord](https://discord.com/channels/775859454780244028/941888268033736766), it appears this might not result in any meaningful performance improvements on production.

I'll close this for now; my main motivation with this PR was figuring out a performant way to include the V2 contracts. Given we've solved that (with lazy importing V2 when required), we can revisit overall bundle size later on.

Issue #513, opened 2022-02-10
[Parent] Add ability to create V2 project


aeolianeth:
## Design

- [ ] https://github.com/jbx-protocol/juice-interface/issues/510

## Acceptance criteria

As a user, I can create a V2 project that has feature-parity with V1 project creation.

- [x] https://github.com/jbx-protocol/juice-interface/issues/541
- [x] https://github.com/jbx-protocol/juice-interface/issues/542
- [x] https://github.com/jbx-protocol/juice-interface/issues/543
- [x] https://github.com/jbx-protocol/juice-interface/issues/544

aeolianeth:
Given this is shipped to Rinkeby, I think this is good to close!

UI Polish follow-ups happening in https://github.com/jbx-protocol/juice-interface/issues/510

Issue #437, opened 2022-01-29
Add JB contract types with Typechain


aeolianeth:
## Description

TBA

See branch `aeolianeth/typechain`

PR #551, opened 2022-02-15
[FEATURE] v2 create project

aeolianeth:

Base feature branch for the V2 project creation feature.

Related to https://github.com/jbx-protocol/juice-interface/issues/513

## To do

- [ ] https://github.com/jbx-protocol/juice-interface/issues/576

changes._

## Acceptance checklist

- [ ] I have evaluated the [Approval Guidelines](../CONTRIBUTING.md#approval-guidelines) for this PR.
- [ ] V1 functionality remains unaffected
- [ ] Translations are up-to-date


johnnyd-eth:
@aeolianeth notes on the funding form:

- Switching between options in the top select doesn't save properly to redux
- Card text in dark theme

johnnyd-eth:
**Next steps:**
Must be done before merge:
- fix V2 review and deploy modal
- fix bugs in funding tab when saving that tab

Non-essential, maybe another PR:
- remove save button from each tab and auto save to redux when tabs are changed or window is closed

src/Router.tsx
```diff
@@ -20,6 +26,7 @@ export default function Router() {
         <Route path="/create">
```
peripheralist, line 26: i'm not too well-read on Suspense/Lazy. any reason we shouldn't use this for v1 create as well? given i'm sure it adds a bit of loading time, we could probably omit it for projects page (which won't require loading any big ABIs anyway)

src/components/shared/formItems/BondingCurve/ProjectBondingCurveRate.tsx
```diff
@@ -11,19 +11,25 @@ import {
...
-import NumberSlider from '../inputs/NumberSlider'
-import { FormItemExt } from './formItemExt'
+import NumberSlider from '../../inputs/NumberSlider'
+import { FormItemExt } from '../formItemExt'
 
 export default function ProjectBondingCurveRate({
   name,
   hideLabel,
   value,
+  style = {},
```
peripheralist, line 21: does this need a default value?

peripheralist:
might wanna fix some colors for dark mode, i'll drop other cases if i find any 
image

peripheralist:
image

src/hooks/v2/ProjectMetadata.ts
```diff
@@ -0,0 +1,26 @@
+import axios from 'axios'
+
+import { consolidateMetadata } from 'models/project-metadata'
+import { useQuery } from 'react-query'
+import { ipfsCidUrl } from 'utils/ipfs'
+
+export default function useV2ProjectMetadata(metadataCID?: string) {
```
peripheralist, line 7: do we need a separate hook from the one used in v1? if so, should we move the current /hooks/ProjectMetadata.ts to hooks/v1/ ?

src/utils/v2/currency.ts
```diff
@@ -0,0 +1,46 @@
+import { V1CurrencyOption } from 'models/v1/currencyOption'
+import { CSSProperties } from 'react'
+
+import { V2CurrencyOption } from 'models/v2/currencyOption'
+
+import { V1_CURRENCY_ETH, V1_CURRENCY_USD } from 'constants/v1/currency'
+import { V2_CURRENCY_ETH, V2_CURRENCY_USD } from 'constants/v2/currency'
+
+const currencies: Record<
```
peripheralist, line 9: seems like this file and /utils/v1/currency.ts are the same. i'm assuming we have separate files since v2 will support additional currencies? are we sure v2 currencies won't be backwards compatible with the current currency.ts file?

src/constants/v2/metadataDomain.ts
```diff
@@ -0,0 +1 @@
+export const PEEL_METADATA_DOMAIN = 1
```
peripheralist, line 1: WOOOO

src/redux/slices/editingV2Project.ts
```diff
@@ -0,0 +1,176 @@
...
+  version: number
+  projectMetadata?: ProjectMetadataV3
+  fundingCycleData?: SerializedV2FundingCycleData
+  fundingCycleMetadata?: SerializedV2FundingCycleMetadata
+  fundAccessConstraints?: V2FundAccessConstraint[]
+  payoutSplits?: ETHPayoutGroupedSplits
+  reserveTokenSplits?: ReserveTokenGroupedSplits
+}
+
+export interface V2ProjectState {
```
peripheralist, line 37: what's the difference between how this and `EditingV2ProjectState` are used?

src/hooks/v2/ProjectMetadata.ts
aeolianeth: Probably not I think! Great catch 🙏

src/utils/v2/currency.ts
aeolianeth: @peripheralist they are _pretty_ much the same, with the important difference that they use the `V2CurrencyOption`, which is ETH=1, USD=2, as opposed to V1 which is ETH=0, USD=1.

I took the path of least resistance initially, but I think we can probably consolidate some/all of this currency stuff into 1 file and use generics to switch between currency types. Will take a look!

src/constants/v2/metadataDomain.ts
aeolianeth: lfg....

src/utils/v2/currency.ts
aeolianeth: With this one, it turns out to be tricky (impossible?) to create good generic functions here because we're relying on the type name to determine the currency, instead of a property _of_ the type.

For example, one way I think we could improve this is by defining each currency type like this:
```ts
type V1CurrencyETH = {
  id: 0,
  version: '1'
  name: 'ETH'
}
```

This way, we could define a generic function, and switch on `currency.version`. I think this might be out of scope for this PR, but certainly something we can potentially improve. Let me know what you think 🙏

src/redux/slices/editingV2Project.ts
aeolianeth: Ah great catch! Looks like `EditingV2ProjectState` isn't actually used anymore 🙈 . Removed!

src/Router.tsx
aeolianeth: In the case of the V1 ABIS, they're all loaded together in the `V1ContractLoader`. I did test this a while ago and recall that lazy-loading the V1 routes didn't have any effect. In an ideal world, we'd only be loading the required ABI's for each page (which would probably be a pretty huge performance boost, tbh).

I think this is where Typechain can help us eventually (https://github.com/jbx-protocol/juice-interface/issues/437). Basically it does codegen for ABI's, which I'm guessing will be tree-shakable. Definitely worth us looking into this stuff as the ABIs are by far the bulkiest thing in the bundle.

Anyways, the takeaway is that, because V2ContractLoader is only used in the V2 routes, lazy-loading the V2 routes means the V2 ABIs are _only_ loaded on V2 routes. Meaning, V1 routes aren't affected by the addition of the V2 ABI dependency. This is testable by observing the network tab: the chunk that contains the V2 ABIs is loaded when a V2 route is active:


https://user-images.githubusercontent.com/94939382/156080478-397012f8-70b6-4fa5-be8d-c36382a7f853.mp4

johnnyd-eth:
> might wanna fix some colors for dark mode image
> 
> image
> 
> looks like the issue may be that the `<p>` elem is inheriting some antd styles from the `Form` image

Yep I've fixed this is in #612 - JZ's style update branch

src/components/v2/V2Create/DeployProjectButton.tsx
```diff
@@ -0,0 +1,101 @@
...
+        fundAccessConstraints,
+        groupedSplits,
+      },
+      {
+        onDone: () => {
+          console.info('Transaction executed. Awaiting confirmation...')
+        },
+        onConfirmed: e => {
+          console.info('Transaction confirmed:', e)
+          window.location.hash = '/projects?tab=all'
```
aeolianeth, line 69: TODO redirect to the project page of the deployed project

src/components/v2/V2Project/index.tsx
```diff
@@ -0,0 +1,59 @@
+import ProjectHeader from 'components/shared/ProjectHeader'
+import { V2ProjectContext } from 'contexts/v2/projectContext'
+import { useContext } from 'react'
+import { fromPermille } from 'utils/formatNumber'
+
+export default function V2Project() {
```
aeolianeth, line 6: Minimal project page for now. Will be pimped soon!

src/Router.tsx
peripheralist: got it, very cool idea.

src/utils/v2/currency.ts
peripheralist: ohh whoops, missed that the numbers were different haha. in that case the versioning is interesting, but i'm happy to keep them as separate files

aeolianeth:
Oki doki @peripheralist, conflicts resolved, if my eyes aren't deceiving me at this late hour, I believe we are ready for `squash and merge` 🤘 

cc @johnnyd-eth

src/components/v2/V2Create/DeployProjectButton.tsx
```diff
@@ -10,6 +10,26 @@ import {
...
+import { readProvider } from 'constants/readProvider'
+
+const CREATE_EVENT_IDX = 0
+const PROJECT_ID_TOPIC_IDX = 3
+
+/**
+ * Return the project ID created from a `launchProjectFor` transaction.
+ * @param txReceipt receipt of `launchProjectFor` transaction
+ */
+const getProjectIdFromReceipt = (txReceipt: TransactionReceipt): number => {
```
peripheralist, line 26: great implementation here 👏

src/components/v2/V2Create/DeployProjectButton.tsx
```diff
@@ -61,12 +82,22 @@ export default function DeployProjectButton({
         groupedSplits,
       },
       {
-        onDone: () => {
+        onDone() {
```
peripheralist, line 85: never knew this was legitimate notation 😳

peripheralist (approved):
big diff energy 😎

PR #553, opened 2022-02-15
Page archived projects using project ids in subgraph query

peripheralist:

Previously archived projects have been paged using a filter _after_ data is returned from the initial subgraph query. This means that if no archived projects were returned in the original query, they won't be shown after the list is filtered.

This PR updates the projects query hooks to query projects by id using `archivedProjectIds`, removing the need for a second-step filter on data that has already been returned from the query.

Note: Since we aren't using the project "states" anywhere to query both archived and active projects at the same time, they've been removed for simplicity. Projects can now only be queried by archived _OR_ active.

UI changes:
- an `archived` tag has been added to project cards
- the Search bar is no longer hidden when archived projects are being queried (running a search query still has the same behavior and ignores any other filters)

image
image

## Acceptance checklist

### DIFF
--- a/src/components/Projects/index.tsx
+++ b/src/components/Projects/index.tsx
@@ -5,7 +5,7 @@ import Search from 'antd/lib/input/Search'
 import FeedbackFormLink from 'components/shared/FeedbackFormLink'
 import Loading from 'components/shared/Loading'
 
-import { ProjectCategory, ProjectStateFilter } from 'models/project-visibility'
+import { ProjectCategory } from 'models/project-visibility'
 import React, { useContext, useEffect, useMemo, useRef, useState } from 'react'
 
 import Grid from 'components/shared/Grid'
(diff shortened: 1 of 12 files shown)


src/hooks/v1/Projects.ts
```diff
@@ -74,32 +53,37 @@ const queryOpts = (
...
+      key: 'id' as const,
+      value: archivedProjectIds,
+      operator: 'in',
+    })
+  } else if (opts.projectId) {
+    where.push({
+      key: 'id' as const,
+      value: opts.projectId.toString(),
+    })
+  }
```
peripheralist, line 78: filter by archived project IDs if `opts.state === 'archived'`, else filter by `opts.projectId` (if passed)

src/utils/graph.ts
```diff
@@ -83,7 +83,7 @@ export type OrderDirection = 'asc' | 'desc'
 
 export type WhereConfig<E extends EntityKey> = {
   key: EntityKeys<E>
-  value: string | number | boolean | string[]
+  value: string | number | boolean | string[] | number[]
```
peripheralist, line 86: expanded type definition to permit filtering by a string of numbers (archived project IDs). current `WhereConfig` functionality is still compatible without any changes

src/components/Projects/index.tsx
```diff
@@ -166,20 +158,18 @@ export default function Projects() {
...
-            placeholder={t`Search projects by handle`}
-            onSearch={val => {
-              setSearchText(val)
-              history.push(`/projects?tab=all${val ? `&search=${val}` : ''}`)
-            }}
-            defaultValue={searchText}
-            allowClear
-          />
-        ) : null}
+        <Search
```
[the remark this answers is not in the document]
aeolianeth: Nice 👍

src/components/shared/ProjectCard.tsx
```diff
@@ -161,6 +165,23 @@ export default function ProjectCard({
...
+                top: 0,
+                right: 0,
+                padding: '2px 4px',
+                background: colors.background.l1,
+                fontSize: '0.7rem',
+                color: colors.text.tertiary,
+                fontWeight: 500,
+              }}
+            >
+              ARCHIVED
```
aeolianeth, line 182: Looks great!

aeolianeth (approved):
@peripheralist LGTM! Nice code + UX improvement, and thanks for your self-review comments, too 🙏

PR #547, opened 2022-02-14
V2 create tokens tab

johnnyd-eth:
## What does this PR do and why?

Adds tokens tab to v2 create flow

Related to https://github.com/jbx-protocol/juice-interface/issues/543

## Screenshots or screen recordings

https://user-images.githubusercontent.com/96150256/153803842-ca0e9cbe-17aa-43d3-8690-416f35af6014.mp4

## Acceptance checklist

- [ ] I have evaluated the [Approval Guidelines](../CONTRIBUTING.md#approval-guidelines) for this PR.
- [ ] I have tested this PR in [all supported browsers](../CONTRIBUTING.md#browser-support).

Issue #543, opened 2022-02-13
Create V2 Project: Build "Token" tab form


aeolianeth:
https://www.figma.com/proto/rySiZBwwOJaG2KPL32l3aV/JBX-final-website-design?page-id=15%3A6246&node-id=304%3A12639&viewport=303%2C48%2C0.52&scaling=min-zoom&starting-point-node-id=304%3A12520

Data to capture:
- Discount rate
- reserve token allocation + beneficiaries
- Bonding curve

Relevant JB v2 contract docs:

- [JBFundingCycleData](https://docs.juicebox.money/protocol/specifications/data-structures/jbfundingcycledata)
- [JBFundingCycleMetadata](https://docs.juicebox.money/protocol/specifications/data-structures/jbfundingcyclemetadata)

PR #562, opened 2022-02-17
Feature/v2 create token form

aeolianeth:

@johnnyd-eth's work towards https://github.com/jbx-protocol/juice-interface/issues/543.

See original PR for more details: https://github.com/jbx-protocol/juice-interface/pull/547


aeolianeth:
@johnnyd-eth thanks for this! We'll still need to link up redux and make sure the deploy transaction works, but we'll do this later on. UI looks great, merging into `feature/v2-create-project` branch

PR #588, opened 2022-02-23
Add ability to config custom pinata gateway

aeolianeth:

If a dev isn't using the private `jbx` pinata credentials, any IPFS data they upload won't be available at the currently hard-coded `jbx.mypinata.cloud` gateway.

This PR lets devs set a `REACT_APP_PINATA_GATEWAY_HOSTNAME` in their `.env`.

## Infra updates

Fleek has been updated on `rinkeby` and `mainnet` to specify `REACT_APP_PINATA_GATEWAY_HOSTNAME=jbx.mypinata.cloud`

Fleek:
Screen Shot 2022-02-25 at 7 41 54 am

## Acceptance checklist

- [-] I have tested this PR in [all supported browsers](../CONTRIBUTING.md#browser-support).

### DIFF
--- a/.example.env
+++ b/.example.env
@@ -5,5 +5,6 @@ REACT_APP_SUBGRAPH_URL= # URL of Juicebox subgraph from thegraph.com
 
 REACT_APP_PINATA_PINNER_KEY= # Pinata API key
 REACT_APP_PINATA_PINNER_SECRET= # Pinata API secret
+REACT_APP_PINATA_GATEWAY_HOSTNAME= # Pinata gateway e.g. 'gateway.pinata.cloud'
 
 REACT_APP_BLOCKNATIVE_API_KEY=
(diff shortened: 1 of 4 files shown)


peripheralist:
great call on this @aeolianeth. why don't we just fall back to the generic gateway.pinata.cloud instead of the jbx one? bit less to think about for anyone trying to use the app locally

aeolianeth:
> great call on this @aeolianeth. why don't we just fall back to the generic gateway.pinata.cloud instead of the jbx one? bit less to think about for anyone trying to use the app locally

yeah totally down with that suggestion! just need to set the env var in fleek before merging this PR. I guess we can also verify it works using the fleek preview 👍 . Will handle this

peripheralist (approved):
awesome. nice thinking pulling that into its own constants/ file!

PR #589, opened 2022-02-23
Fix custom strategy bug + general clean up

aeolianeth:

- fix bug where selecting `custom strategy` breaks the page
- clean up ballot strategy utils/constants
- use aliased components/ directory imports for shared components (where it makes sense)

## Acceptance checklist


src/components/shared/forms/RulesForm.tsx
```diff
@@ -89,7 +92,7 @@ export default function RulesForm({
           </Trans>
         </p>
 
-        {ballotStrategies()[selectedIndex ?? 0].address ===
+        {BALLOT_STRATEGIES[selectedIndex ?? 0]?.address ===
```
aeolianeth, line 95: This is the bug fix

aeolianeth:
@peripheralist @johnnyd-eth could ya'll pls review?

peripheralist:
hm, i'm confused why the bug was happening before.

iirc @johnnyd-eth organized those files/strategies the way they were to avoid some bugs with lingui not working with them, although i don't think we ever figured out what the real issue was.

aeolianeth:
@peripheralist good call-out, in any case, I think @johnnyd-eth is fixing this in https://github.com/jbx-protocol/juice-interface/pull/608, so I'll close this for now

Issue #600, opened 2022-02-23
[UNARCHIVE] Project: "Test"


peripheralist:
<b>Chain:</b> rinkeby 
 <b>Handle:</b> test 
 <b>Id:</b> 24

peripheralist:
JUST TESTIN not a legit issue

Issue #587, opened 2022-02-23
[IDEA] Automate project archival requests


peripheralist:
## Problem to solve

There's currently no way for project owners to archive projects without requesting manually in Discord. This creates a secondary problem of needing to verify that the requester is in fact the owner of the project being archived.

## Proposal

1. **Add an `archived` property to project metadata, and a form for updating the property.** Allowing project owners to manually set their project to `archived: true` via the app will allow:
- their ownership of the project to be verified
- the project can immediately appear as "archived" or "pending archive" (even though it will need to be manually filtered into the archived list)
2. **Connect the "archive project" form to a Discord channel, or have it create a Github issue automatically.** When a project is archived, the Peel team will be notified that it should be manually added to the archived list.

peripheralist:
duplicate of #567 , will keep this since it includes a more up to date proposal

PR #601, opened 2022-02-23
Allow archiving projects via Project Metadata

peripheralist:

Addresses #587 

Allows project owners to archive/unarchive their projects via the project metadata, which also automatically notifies us via Github to update `archivedProjectIds`.

- New ProjectMetadataV4 with `archived` property. Projects will appear archived if their `metadata.archived === true` OR if `archivedProjectIds.includes(<id>)`
- New archive/unarchive functionality in project tools drawer. This will update the project's metadata, and automatically create a Github issue to notify us to update the `archivedProjectIds` as needed (see #599 & #600)
- Adds Octokit integration to allow automating creation of github issues. _Note:_ This requires a new .env var `REACT_APP_GITHUB_ACCESS_TOKEN`

image
image

## Acceptance checklist

### DIFF
--- a/src/components/v1/V1Project/modals/ProjectToolDrawerModal.tsx
+++ b/src/components/v1/V1Project/modals/ProjectToolDrawerModal.tsx
@@ -298,12 +298,21 @@ export default function ProjectToolDrawerModal({
               </Trans>
             </p>
             <p>
-              <Trans>
-                <strong>Note:</strong>{' '}
-                {terminal?.version === '1.1'
-                  ? 'Unless payments are paused in your funding cycle settings, your project will still be able to receive payments directly through the jbx protocol contracts.'
-                  : 'Your project will still be able to receive payments directly through the jbx protocol contracts.'}
-              </Trans>
+              <strong>
+                <Trans>Note:</Trans>
+              </strong>{' '}
+              {terminal?.version === '1.1' ? (
+                <Trans>
+                  Unless payments are paused in your funding cycle settings,
+                  your project will still be able to receive payments directly
+                  through the Juicebox protocol contracts.
+                </Trans>
+              ) : (
+                <Trans>
+                  Your project will still be able to receive payments directly
+                  through the Juicebox protocol contracts.
+                </Trans>
+              )}
             </p>
             <Button
               onClick={() => setArchived(true)}
--- a/src/components/v1/V1Project/modals/ProjectToolDrawerModal.tsx
+++ b/src/components/v1/V1Project/modals/ProjectToolDrawerModal.tsx
@@ -4,13 +4,17 @@ import { useForm } from 'antd/lib/form/Form'
 import { FormItems } from 'components/shared/formItems'
 import InputAccessoryButton from 'components/shared/InputAccessoryButton'
 import FormattedNumberInput from 'components/shared/inputs/FormattedNumberInput'
+import { readNetwork } from 'constants/networks'
+import { octokit } from 'constants/octokit'
 import { V1ProjectContext } from 'contexts/v1/projectContext'
 import useUnclaimedBalanceOfUser from 'hooks/v1/contractReader/UnclaimedBalanceOfUser'
 import { useAddToBalanceTx } from 'hooks/v1/transactor/AddToBalanceTx'
 import { useSafeTransferFromTx } from 'hooks/v1/transactor/SafeTransferFromTx'
+import { useSetProjectUriTx } from 'hooks/v1/transactor/SetProjectUriTx'
 import { useTransferTokensTx } from 'hooks/v1/transactor/TransferTokensTx'
 import { useContext, useState } from 'react'
 import { formatWad, fromWad, parseWad } from 'utils/formatNumber'
+import { uploadProjectMetadata } from 'utils/ipfs'
 import { tokenSymbolText } from 'utils/tokenSymbolText'
 
 export default function ProjectToolDrawerModal({
@@ -20,15 +24,18 @@ export default function ProjectToolDrawerModal({
   visible?: boolean
   onClose?: VoidFunction
 }) {
-  const { tokenSymbol, owner } = useContext(V1ProjectContext)
+  const { tokenSymbol, owner, terminal, metadata, projectId, handle } =
+    useContext(V1ProjectContext)
   const safeTransferFromTx = useSafeTransferFromTx()
   const transferTokensTx = useTransferTokensTx()
   const addToBalanceTx = useAddToBalanceTx()
+  const setUriTx = useSetProjectUriTx()
 
   const [loadingAddToBalance, setLoadingAddToBalance] = useState<boolean>()
   const [loadingTransferTokens, setLoadingTransferTokens] = useState<boolean>()
   const [loadingTransferOwnership, setLoadingTransferOwnership] =
     useState<boolean>()
+  const [loadingArchive, setLoadingArchive] = useState<boolean>()
   const [transferTokensForm] = useForm<{ amount: string; to: string }>()
   const [addToBalanceForm] = useForm<{ amount: string }>()
   const [transferOwnershipForm] = useForm<{ to: string }>()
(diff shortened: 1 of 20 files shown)


src/constants/octokit.ts
```diff
@@ -0,0 +1,5 @@
+import { Octokit } from '@octokit/core'
+
+export const octokit = new Octokit({
+  auth: process.env.REACT_APP_GITHUB_ACCESS_TOKEN,
```
peripheralist, line 4: We may want to throw an error on load if the .env var isn't set

johnnyd-eth (approved):
This is so sick @peripheralist, absolutely key update. This flow is clearly a massive improvement from what we've got, and I can't think of how we could make it any easier for ourselves and project owners than this. I left one minor comment but I'm approving in advance ✅

src/components/v1/V1Project/modals/ProjectToolDrawerModal.tsx
```diff
@@ -218,6 +259,62 @@ export default function ProjectToolDrawerModal({
...
+                unarchive a project at any time. Please allow a few days for
+                your project to appear under the "archived" filter on the
+                Projects page.
+              </Trans>
+            </p>
+            <p>
+              <Trans>
+                <strong>Note:</strong>{' '}
+                {terminal?.version === '1.1'
+                  ? 'Unless payments are paused in your funding cycle settings, your project will still be able to receive payments directly through the jbx protocol contracts.'
```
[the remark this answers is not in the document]
peripheralist: ohh good catch, wrote this off the cuff. should probably be "Juicebox protocol" actually

src/components/v1/V1Project/modals/ProjectToolDrawerModal.tsx
```diff
@@ -76,6 +83,40 @@ export default function ProjectToolDrawerModal({
...
+        onConfirmed: () => {
+          // Create github issue when archive is requested
+          // https://docs.github.com/en/rest/reference/issues#create-an-issue
+          octokit.request('POST /repos/jbx-protocol/juice-interface/issues', {
+            title: `[${archived ? 'ARCHIVE' : 'UNARCHIVE'}] Project: "${
+              metadata?.name
+            }"`,
+            body: `<b>Chain:</b> ${
+              readNetwork.name
+            } \n <b>Handle:</b> ${handle} \n <b>Id:</b> ${projectId?.toString()}`,
```
aeolianeth, line 113: **suggestion**: I imagine this might take markdown, wondering if this works:

```suggestion
            body: `**Chain:** ${readNetwork.name}
            **Project handle:** ${handle}
            **Project ID:** ${projectId?.toString()}`,
```

src/components/v1/V1Project/modals/ProjectToolDrawerModal.tsx
```diff
@@ -76,6 +83,40 @@ export default function ProjectToolDrawerModal({
...
+    }
+
+    setUriTx(
+      { cid: uploadedMetadata.cid },
+      {
+        onDone: () => setLoadingArchive(false),
+        onConfirmed: () => {
+          // Create github issue when archive is requested
+          // https://docs.github.com/en/rest/reference/issues#create-an-issue
+          octokit.request('POST /repos/jbx-protocol/juice-interface/issues', {
```
aeolianeth, line 107: **suggestion (non-blocking)** would be cool to add a github label although couldn't find docs for how to do that

src/components/v1/V1Project/modals/ProjectToolDrawerModal.tsx
aeolianeth, line 304: **suggestion** can we wrap these strings in `Trans` tags?

aeolianeth (changes requested):
@peripheralist love it! The github issue workflow feels very good, I think this will scale well into the medium term.

I left a few minor comments, but I'm most interested in your thoughts on potentially tackling this a different way to avoid adding a dependency. LMK! 🙏 

cc @johnnyd-eth

src/constants/octokit.ts
```diff
@@ -0,0 +1,5 @@
+import { Octokit } from '@octokit/core'
+
+export const octokit = new Octokit({
```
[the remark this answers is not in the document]
peripheralist: agreed. feels meh in constants, but couldn't find a better place to put it. to me this feels logically similar to ReadProvider which is in the same dir, so 🤷

src/constants/octokit.ts
```diff
@@ -0,0 +1,5 @@
+import { Octokit } from '@octokit/core'
```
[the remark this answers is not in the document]
peripheralist: good looking out here. looks like the package is only 52kB so pretty tiny https://www.npmjs.com/package/@octokit/core. kinda odd, we could only depend on this package https://www.npmjs.com/package/@octokit/request which seems like it's supposed to be slimmer, but its size is actualy larger at 57kB... 🤔 

i'm not sure the best way to make that tree shaking work. currently the ProjectToolDrawer itself is loaded even on the app landing page lol... 
image

src/components/v1/V1Project/modals/ProjectToolDrawerModal.tsx
peripheralist: thank you!!

src/components/v1/V1Project/modals/ProjectToolDrawerModal.tsx
peripheralist: hm... really thought that would work

image

image

src/constants/octokit.ts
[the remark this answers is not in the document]
aeolianeth: Sweet thanks for looking into it. I think I would still much prefer this lives in a script outside of the app (github action triggered by a POST request feels good to me), but also want to keep a bias towards getting this available to users. Happy to leave this up to you 🙏

src/components/v1/V1Project/modals/ProjectToolDrawerModal.tsx
aeolianeth: LAME. All good haha, thanks for indulging me 😊

src/components/v1/V1Project/modals/ProjectToolDrawerModal.tsx
aeolianeth: yeah love either. We can slap whatever we use on the Crowdin PRs too

src/components/v1/V1Project/modals/ProjectToolDrawerModal.tsx
aeolianeth: I think there's a way we can lint for missing translations 👀 . Will look into it e v e n t u a l l y

src/constants/octokit.ts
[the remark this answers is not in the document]
peripheralist: > Sweet thanks for looking into it. I think I would still much prefer this lives in a script outside of the app (github action triggered by a POST request feels good to me), but also want to keep a bias towards getting this available to users. Happy to leave this up to you 🙏

great suggestion! agreed, let me try to get that working.

peripheralist:
update: was able to remove the Octokit dependency and just use a simple `axios.post()` to create github issue. (i tried this first, but must have messed something up the first time)

other important change—i've moved the issue creation step _before_ the `setUri` tx is created. this ensures the issue gets created even if a user leaves the page while the tx is pending. the tradeoff is that it's now possible for an issue to be created for a project whose archive tx didn't complete, so we'll just wanna double check all projects that have archive/unarchive issues created to make sure that their metadata was actually updated.

src/utils/ipfs.ts
```diff
@@ -271,20 +271,14 @@ export const unpinIpfsFileByCid = (cid: string | undefined) =>
...
-    {
-      ...metadata,
-      version: 3,
-    },
-    {
-      name: 'juicebox-project-metadata.json',
-      metadata: {
-        tag: IPFS_TAGS.METADATA,
-      },
+  pinJSONToIpfs(consolidateMetadata(metadata), {
```
peripheralist, line 276: passing improvement, just ensures that anytime metadata is uploaded it's converted to the latest version

PR #602, opened 2022-02-24
[BUG] Fix bug in v1 incentives form 

### LINKED ISSUE #595: [BUG] Reconfiguring "incentives" form: numbers don't match slider values
## Summary

When reconfiguring a funding cycle, clicking on "incentives" and the following occurs:
- it shoes bonding curve at 100%, but the slider seems correct at its current amount of 70%
- and discount rate shows at 0%, but the slider seems correct at 10%

## Steps to reproduce

1. Go to project page
2. Click "reconfigure upcoming"
3. click "incentives"
4. Observe. the numbers will not being in sync with the slider

## What is the _current_ bug behavior?

the numbers are initially out of sync with the slider

## What is the expected _correct_ behavior?

the numbers should always be in sync with the slider, even intiailly.

johnnyd-eth:

Fixes #595. Form was not rerendering when the initial value was updated from undefined to actual value. Problem was occurring in both project create and 'reconfigure funding' modal. 

I made the form rely purely on `useForm` and removed all states from it. 

https://user-images.githubusercontent.com/96150256/155434856-f912a7c9-7510-4446-b03c-618896f44e83.mp4

## Acceptance checklist


peripheralist:
lawd, these antd Form + state patterns are just too much.

i actually think we've seen this issue elsewhere—after poking around a bit, it looks like we can solve this by removing the `name=` properties from the Form item components, instead of adding a FormInstance that we shouldn't need. i think the issue is that having the `name` property set causes the wrapped Input to look for a value that doesn't exist, which overrides the value we're trying to pass to it via the state object.

does that check out on your end @johnnyd-eth ?

src/components/v1/V1Project/modals/ReconfigureFCModal.tsx
```diff
@@ -594,15 +602,14 @@ export default function ReconfigureFCModal({
...
-          initialBondingCurveRate={fromPerbicent(editingFC.bondingCurveRate)}
+          form={incentivesForm}
           disableBondingCurve={
             !hasFundingTarget(editingFC)
               ? t`Bonding curve disabled while no funding target is set.`
               : undefined
           }
           onSave={async (discountRate: string, bondingCurveRate: string) => {
-            await ticketingForm.validateFields()
+            await incentivesForm.validateFields()
```
aeolianeth, line 612: good catch lol

src/components/v1/V1Create/index.tsx
```diff
@@ -152,6 +171,7 @@ export default function V1Create() {
     restrictedActionsForm,
   ])
 
+  // The following useEffect's set the initial values of all the forms' fields from redux
```
aeolianeth, line 174: thank you ser

aeolianeth (commented):
@johnnyd-eth this is TIGHT! Thanks for taking the time to find a different solution, this feels far more maintainable as we now have a single source of truth for our state (the form) - nice work!

I haven't had a chance to test yet, so please remind me if I don't get to it by tomorrow. Otherwise @peripheralist, feel free to merge if you're happy with it!

johnnyd-eth:
> lawd, these antd Form + state patterns are just too much.
> 
> i actually think we've seen this issue elsewhere—after poking around a bit, it looks like we can solve this by removing the `name=` properties from the Form item components, instead of adding a FormInstance that we shouldn't need. i think the issue is that having the `name` property set causes the wrapped Input to look for a value that doesn't exist, which overrides the value we're trying to pass to it via the state object.
> 
> does that check out on your end @johnnyd-eth ?

@peripheralist I did try removing the `name`, but remembered that the form validations I put in a few months ago were relying on them, and were failing without `name`. So I figured we just go all in on the `form` and remove the state, which seems to fix the bug, give a single source for values and keep validation working.

aeolianeth:
AH @peripheralist, sorry i did not see your initial comment! Yes, these forms... 😅 . I will leave this in your capable hands 🙏 🙏 

The one thing I'll note is that I'm pretty sure `<Form />` elements without a binded `form` instance causes a console error (although not sure in this specific case, I've seen it elsewhere)

peripheralist:
if we can get it down to only one source of truth it should definitely be the FormInstance! the only concern with this pattern is that any changes to the sliders won't reset when the form is _cancelled_ since the FormInstance value updates with the `input.onChange` (the actual `EditingProject` state values won't update unless you click save).

maybe we could fix this by resetting the FormInstance value to the current `EditingProject` every time the form component is closed?

johnnyd-eth:
> if we can get it down to only one source of truth it should definitely be the FormInstance! the only concern with this pattern is that any changes to the sliders won't reset when the form is _cancelled_ since the FormInstance value updates with the `input.onChange` (the actual `EditingProject` state values won't update unless you click save).
> 
> maybe we could fix this by resetting the FormInstance value to the current `EditingProject` every time the form component is closed?
 
@peripheralist sorry this one slipped my mind for a few days with the v2 push! Great points you've made there. I may the form fields reset when the drawer is closed. Then, I was finding the slider value to not be updating reliably. To add some robustness I decided to add pass through a `formState` prop which is number that increments each time the drawer is opened. Then by adding this `formState` to the `useEffect` in the `NumberSlider`, we make sure the slider value is updated to the form field value each time the drawer is opened.

Without `formState` (slider value remains at the unsaved value):

https://user-images.githubusercontent.com/96150256/156121014-fdb52f07-f632-4b7a-b072-41746dcef1e8.mp4

With `formState`:

https://user-images.githubusercontent.com/96150256/156121137-7d4fc890-ad8c-4b38-81ab-c5ee11cfedcb.mp4

src/components/v1/V1Create/index.tsx
```diff
@@ -690,16 +708,14 @@ export default function V1Create() {
           {...drawerStyle}
           onClose={() => {
             viewedCurrentStep()
+            resetIncentiveForm()
             setIncentivesFormModalVisible(false)
+            setIncentivesFormState(incentivesFormState + 1)
           }}
         >
           <IncentivesForm
```
peripheralist, line 716: Nice @johnnyd-eth! Seems to work great. What do you think about using a pattern like this instead of `formState`? This causes the entire form component to re-render whenever the modal is closed. Still kinda awk, but feels a bit more lightweight and would save us having to pass `formState` down through so many components

```suggestion
{incentivesFormModalVisible &&
          <IncentivesForm
```

johnnyd-eth: Ahh yeah that's way better lol I over-thunk it. Cheers Peri!

PR #610, opened 2022-02-26
Remove v1-specific copy from ProjectDetailsForm

aeolianeth:

This PR picks some of the changes from https://github.com/jbx-protocol/juice-interface/pull/551.

It makes `src/components/shared/forms/ProjectDetailsForm.tsx` form-only and includes no copy. This is so that we can re-use this form in V2.

UI is unchanged

## Acceptance checklist


src/components/shared/forms/ProjectDetailsForm.tsx
```diff
@@ -0,0 +1,77 @@
...
+  infoUri: string
+  handle: string
+  logoUri: string
+  twitter: string
+  discord: string
+  payButton: string
+  payDisclosure: string
+}
+
+export default function ProjectDetailsForm({
```
aeolianeth, line 20: This component is mostly the same is previously. I'll point out the specific changes below.

src/components/shared/forms/ProjectDetailsForm.tsx
```diff
@@ -0,0 +1,77 @@
...
+  form,
+  onFinish,
+  hideProjectHandle = false,
+}: {
+  form: FormInstance<ProjectDetailsFormFields>
+  onFinish: (values: ProjectDetailsFormFields) => void
```
aeolianeth, line 26: `onSave` changed to `onFinished` to match Antd `<Form />` prop name

src/components/shared/forms/ProjectDetailsForm.tsx
```diff
@@ -0,0 +1,77 @@
...
+        formItemProps={{
+          rules: [{ required: true }],
+        }}
+        onChange={name => {
+          const val = name ? normalizeHandle(name) : ''
+          // Use `handle` state to enable ProjectHandle to validate while typing
+          form.setFieldsValue({ handle: val })
+        }}
+      />
+      {!hideProjectHandle && (
```
aeolianeth, line 42: V2 doesn't use `handle`'s. This gives us the ability to hide the handle input for V2.

src/components/shared/forms/ProjectDetailsForm.tsx
```diff
@@ -0,0 +1,77 @@
...
+        initialUrl={form.getFieldValue('logoUri')}
+        onSuccess={logoUri => {
+          const prevUrl = form.getFieldValue('logoUri')
+          // Unpin previous file
+          form.setFieldsValue({ logoUri })
+          if (prevUrl) unpinIpfsFileByCid(cidFromUrl(prevUrl))
+        }}
+      />
+      <Form.Item>
+        <Button htmlType="submit" type="primary">
```
aeolianeth, line 71: Remove `onClick` handler, rely on `Form`'s `onFinish` handler instead.

src/components/v1/V1Create/index.tsx
```diff
@@ -333,10 +336,11 @@ export default function V1Create() {
                 color: viewed
                   ? colors.text.primary
                   : colors.text.action.primary,
-                border: viewed
-                  ? '1px solid ' + colors.stroke.secondary
-                  : '1px solid' + colors.stroke.action.primary,
-                borderLeftWidth: active ? 10 : 1,
+                borderColor: viewed
```
aeolianeth, line 339: Passing fix. This fixes a console error where apparently React doesn't like it when styles are overridden. In this case `borderLeftWidth` was overriding the above `border` style.

src/components/v1/V1Create/index.tsx
```diff
@@ -578,15 +582,25 @@ export default function V1Create() {
...
+            </h1>
+            <p>
+              <Trans>
+                Changes to these attributes can be made at any time and will be
+                applied to your project immediately.
+              </Trans>
+            </p>
+            <ProjectDetailsForm
+              form={projectForm}
+              onFinish={() => {
```
aeolianeth, line 597: Note that `await projectForm.validateFields()` has been removed. Antd already calls this function before its `onFinish` callback is called, so calling `validateFields` again here would be redundant.

peripheralist: beautiful! being able to remove click handlers and externally handled validation feels like a big win. great call

peripheralist:
Great PR, nicely scoped. great optimizations using the `onFinish` callback, and love reducing this component to not include any copy 👏

aeolianeth:
Thanks for the review @peripheralist ! I'll assume your approval and merge ✅

PR #630, opened 2022-03-10
Update fee to 2.5% in FAQs

johnnyd-eth:

Updates fee from 5% to 2.5% in FAQs

Screen Shot 2022-03-10 at 1 29 53 pm

## Acceptance checklist

### DIFF
--- a/src/components/Landing/QAs.tsx
+++ b/src/components/Landing/QAs.tsx
@@ -7,6 +7,8 @@ export const OverflowVideoLink = ({ children }: PropsWithChildren<{}>) => (
   <ExternalLink href="https://youtu.be/9Mq5oDh0aBY">{children}</ExternalLink>
 )
 
+const JB_FEE = 2.5
+
 const JBDiscordLink = ({ children }: PropsWithChildren<{}>) => (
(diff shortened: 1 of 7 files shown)


src/components/Landing/QAs.tsx
```diff
@@ -46,7 +46,7 @@ export default function QAs() {
           .
         </Trans>,
         <Trans>
-          Projects building on Juicebox pay a 5% JBX membership fee from
+          Projects building on Juicebox pay a 2.5% JBX membership fee from
```
aeolianeth, line 49: **suggestion** can we make the `2.5` a variable? Doing so will mean that the translations won't be erased if we need to change the value in future.

**suggestion (non-blocking)**: Ideally we'll source it from the contract using the `TerminalFee` hook for v1.1, but this could be done in a follow-up (don't want to unnecessarily block the PR).

aeolianeth (changes requested):
@johnnyd-eth one comment ser

src/components/Landing/QAs.tsx
johnnyd-eth: Good call @aeolianeth. Addressed the first point

aeolianeth (approved):
@johnnyd-eth ty, lgtm!! merging

PR #637, opened 2022-03-13
New manage tokens modal

aeolianeth:

Attempts to address user confusion with `Manage Tokens` modal by adding short descriptions to buttons.

| state | img |
| --- | --- |
| Has print ticket permission, `canPrintPreminedTickets` is false | Screen Shot 2022-03-13 at 8 20 26 pm |
| Has print ticket permission, `canPrintPreminedTickets` is false | Screen Shot 2022-03-13 at 8 22 35 pm | 
| Doesn't have print ticket permission | Screen Shot 2022-03-13 at 8 23 36 pm |
| Project has no overflow (can't "redeem" any ETH) | Screen Shot 2022-03-13 at 8 30 40 pm |

## Acceptance checklist

### DIFF
--- a/src/components/v1/V1Project/Rewards/ManageTokensModal.tsx
+++ b/src/components/v1/V1Project/Rewards/ManageTokensModal.tsx
@@ -1,5 +1,6 @@
 import { t, Trans } from '@lingui/macro'
 import { Modal, Space, Tooltip } from 'antd'
+import ExternalLink from 'components/shared/ExternalLink'
 import RichButton from 'components/shared/RichButton'
 import { V1ProjectContext } from 'contexts/v1/projectContext'
 import useCanPrintPreminedTokens from 'hooks/v1/contractReader/CanPrintPreminedTokens'
@@ -15,20 +16,34 @@ import ConfirmUnstakeTokensModal from '../modals/ConfirmUnstakeTokensModal'
 import PrintPreminedModal from '../modals/PrintPreminedModal'
 import RedeemModal from '../modals/RedeemModal'
 
+const BURN_DEFINITION_LINK =
+  'https://www.investopedia.com/tech/cryptocurrency-burning-can-it-manage-inflation/'
+
+const BurnTokensHelp = () => {
+  return (
+    <Trans>
+      <ExternalLink href={BURN_DEFINITION_LINK}>Learn more</ExternalLink> about
+      burning tokens.
+    </Trans>
+  )
+}
+
 const RedeemButtonTooltip = ({
   buttonDisabled,
   children,
 }: PropsWithChildren<{
   buttonDisabled: boolean
 }>) => {
-  if (!buttonDisabled) return <>{children}</>
-
   return (
     <Tooltip
       title={
-        <Trans>
-          Cannot redeem tokens for ETH because this project has no overflow.
-        </Trans>
+        buttonDisabled ? (
+          <Trans>
+            Cannot redeem tokens for ETH because this project has no overflow.
+          </Trans>
+        ) : (
+          <BurnTokensHelp />
+        )
       }
       placement="right"
     >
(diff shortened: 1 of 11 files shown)


johnnyd-eth:
This is great @aeolianeth. Definitely necessary, I remember being quite confused with these when I first came across it myself. 

Screen Shot 2022-03-14 at 6 44 36 pm

One thought, should we use solid chevrons like we do in the rest of the app (e.g. reconfig modal)?

src/components/v1/V1Project/Rewards/ManageTokensModal.tsx
```diff
@@ -0,0 +1,147 @@
+import { t, Trans } from '@lingui/macro'
```
aeolianeth, line 1: This is mostly the same as previous, but with the new buttons

src/components/v1/V1Project/Rewards/index.tsx
```diff
@@ -1,15 +1,12 @@
 import { BigNumber } from '@ethersproject/bignumber'
```
aeolianeth, line 1: Changes in this file are for moving Manage Tokens modal to its own component

src/components/v1/V1Project/Rewards/index.tsx
```diff
@@ -75,21 +65,16 @@ export default function Rewards() {
...
-    (metadata.version === 0
-      ? canPrintPreminedV1Tickets
-      : metadata.ticketPrintingIsAllowed)
 
   const labelStyle: CSSProperties = {
     width: 128,
   }
 
-  const tokensLabel = tokenSymbol ? tokenSymbol + ' ' + t`ERC20` : t`Tokens`
+  const tokensLabel = tokenSymbolText({
```
aeolianeth, line 73: This change means we don't show the `ERC20` text. Not sure how important it is to include it...

aeolianeth:
Thanks @johnnyd-eth, great suggestion! Applied it, would you mind giving this a code review now? 🙏 

Screen Shot 2022-03-14 at 10 28 08 pm

src/components/v1/V1Project/Rewards/index.tsx
johnnyd-eth: Yeah I don't think we need it

johnnyd-eth:
This is a great improvement @aeolianeth.

Few thoughts:
- Should we have tooltips for any case where a card is disabled? e.g. redeem button has no tooltip when user has no tokens in the project. 
- Might be out of the scope of this PR, but while we're on this modal, I've noticed there's no limit on the 'Redeem tokens for ETH' input. Maybe we should make input go red when total token supply is exceeded, and replace the 'You will receive minimum 0 ETH' for an error message?


Screen Shot 2022-03-15 at 4 07 25 pm

aeolianeth:
> This is a great improvement @aeolianeth.
> 
> Few thoughts:
> 
> * Should we have tooltips for any case where a card is disabled? e.g. redeem button has no tooltip when user has no tokens in the project.
> * Might be out of the scope of this PR, but while we're on this modal, I've noticed there's no limit on the 'Redeem tokens for ETH' input. Maybe we should make input go red when total token supply is exceeded, and replace the 'You will receive minimum 0 ETH' for an error message?
> 
> Screen Shot 2022-03-15 at 4 07 25 pm

Thanks for a great review @johnnyd-eth, appreciate your thoughtful suggestions!

I've added the conditional tooltip, and created a follow-up issue for the input field: https://github.com/jbx-protocol/juice-interface/issues/652

Screen Shot 2022-03-16 at 8 23 38 am

aeolianeth:
@johnnyd-eth made a few more improvements based on our chat earlier:

Screen Shot 2022-03-16 at 10 51 10 am
Screen Shot 2022-03-16 at 10 50 28 am

johnnyd-eth:
@aeolianeth Yep love those changes, linking to the investopedia is a great idea. Nothing more to add from me. Great improvement!

johnnyd-eth (approved):

PR #658, opened 2022-03-16
New pattern for downloading project activity (only supports payments so far)

peripheralist:

Adds download icon next to project activity, which opens `DownloadActivityModal` modal

`DownloadActivityModal` includes "Payments" button, which opens `DownloadPaymentsModal`

`DownloadPaymentsModal` allows the user to download all payments from a specific block number (defaults to latest) using the same pattern for downloading Holders info.

Aside: Also fixes bug with Date not being shown properly when downloading Holders csv

image
image
image
image

## Acceptance checklist


src/components/v1/V1Project/modals/DownloadPaymentsModal.tsx
```diff
@@ -0,0 +1,117 @@
...
+        block: {
+          number: blockNumber,
+        },
+        where: {
+          key: 'project',
+          value: projectId.toString(),
+        },
+      })
+
+      if (!payments) throw new Error('No data.')
```
aeolianeth, line 54: **suggestion** is it worth displaying some error message in the UI?

aeolianeth (approved):
Cool feature @peripheralist ! Left one minor comment, pre-approving 🤘

src/components/v1/V1Project/modals/DownloadPaymentsModal.tsx
peripheralist: Yeah fair point. Especially since we've been seeing more Graph errors lately than usual.

In general we could probably be relying more on the Antd `Notification` element, I'll whip something up

aeolianeth: Ohhh yeah, great idea ❤️ .

Issue #643, opened 2022-03-14
Improve currency context 


johnnyd-eth:
## Proposal

Currency context will simply map a currency string ('ETH' or 'USD') to the integer they use in the contracts (0, 1 or 2). Then throughout the app we can just use the strings everywhere, only using the currency context at the point we interact with the contracts, where we need the currency integer

peripheralist:
Will this allow for 'ETH' and 'USD' to be strongly typed? Right now currency context allows accidentally misspelling `currencyEth` or `currencyUsd` since the object keys are just `string` type. I don't mind keeping the dictionary pattern, something like

```
const { currencies: { ETH, USD } } = useContext(CurrencyContext)
```

as long as `currencies` is typed, so that `{ currencies: { ETHH } }` would show an error

PR #702, opened 2022-03-19
Improve currency context typing

aeolianeth:

- Updates `wadToCurrency` function to accept a currency name (`'ETH'` or `'USD'`) instead of the ID (0 or 1). This is to enable us to use it in V2.
- Updates `CurrencySymbol` component to accept a currency name (`'ETH'` or `'USD'`) instead of the ID (0 or 1). Similarly, for easier V2 use (and IMO just nicer).
- Improves typing of `CurrencyContext`, partially resolving https://github.com/jbx-protocol/juice-interface/issues/643. Specifically we use `CurrencyName`s in the `currencies` object in the context.

No user facing changes.

## Acceptance checklist

### DIFF
--- a/src/components/v1/V1Dashboard/index.tsx
+++ b/src/components/v1/V1Dashboard/index.tsx
@@ -5,7 +5,6 @@ import {
   V1ProjectContext,
   V1ProjectContextType,
 } from 'contexts/v1/projectContext'
-import { CurrencyContext } from 'contexts/currencyContext'
 
 import useBalanceOfProject from 'hooks/v1/contractReader/BalanceOfProject'
 import useCurrentFundingCycleOfProject from 'hooks/v1/contractReader/CurrentFundingCycleOfProject'
@@ -24,7 +23,7 @@ import { useCurrencyConverter } from 'hooks/v1/CurrencyConverter'
 import { useProjectMetadata } from 'hooks/ProjectMetadata'
 import { useProjectsQuery } from 'hooks/v1/Projects'
 import { V1CurrencyOption } from 'models/v1/currencyOption'
-import { useContext, useEffect, useMemo, useState } from 'react'
+import { useEffect, useMemo, useState } from 'react'
 import { useHistory, useLocation, useParams } from 'react-router-dom'
 import { getTerminalName, getTerminalVersion } from 'utils/v1/terminals'
 import useTerminalOfProject from 'hooks/v1/contractReader/TerminalOfProject'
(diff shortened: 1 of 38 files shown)


src/components/Projects/TrendingProjectCard.tsx
```diff
@@ -95,7 +98,9 @@ export default function TrendingProjectCard({
 
   const paymentCount = project.trendingPaymentsCount
 
-  return project ? (
+  if (!project) return <Loading />
```
aeolianeth, line 101: just a small reorg of code, not functional change.

src/components/v2/V2Create/index.tsx
```diff
@@ -56,51 +53,49 @@ export default function V2Create() {
...
     <V2UserProvider>
-      <CurrencyContext.Provider value={V2_CURRENCY_CONTEXT}>
-        {isRinkeby ? <V2WarningBanner /> : null}
-        <div style={{ margin: '4rem', marginBottom: 0 }}>
-          {!isRinkeby && (
-            <div style={{ padding: '1rem', textAlign: 'center' }}>
-              <V2MainnetWarning />
-            </div>
-          )}
+      {isRinkeby ? <V2WarningBanner /> : null}
```
aeolianeth, line 56: Same as previous, without currency context (it's moved to the router)

src/components/v2/V2Project/V2PayButton.tsx
```diff
@@ -11,14 +11,18 @@ import { V2_CURRENCY_ETH, V2_CURRENCY_USD } from 'utils/v2/currency'
...
+
 import V2ConfirmPayOwnerModal from './V2ConfirmPayOwnerModal'
 
 export default function V2PayButton({
   payAmount,
   payInCurrency,
 }: {
   payAmount: string
-  payInCurrency: number
+  payInCurrency: CurrencyOption // TODO make the V2CurrencyOption
```
aeolianeth, line 25: I'm sure there exists a better solution to this, but this change is still a net-improvement.

aeolianeth:
Hey @johnnyd-eth could you review this when you have a chance please?

src/components/v1/V1Dashboard/index.tsx
```diff
@@ -235,7 +237,7 @@ export default function V1Dashboard() {
 
   return (
     <V1ProjectContext.Provider value={project}>
-      <CurrencyContext.Provider value={V1_CURRENCY_CONTEXT}>
+      <V1CurrencyProvider>
```
johnnyd-eth, line 240: **Praise**

johnnyd-eth (approved):
@aeolianeth these are all really solid improvements. The process of just using 'ETH' and 'USD' throughout all components and then using the `CurrencyContext` when we need to get their integer values for contract calls is excellent. Don't want to hold this up at all

aeolianeth:
> @aeolianeth these are all really solid improvements. The process of just using 'ETH' and 'USD' throughout all components and then using the `CurrencyContext` when we need to get their integer values for contract calls is excellent. Don't want to hold this up at all

Yeah and I think we can iteratively replace ints with `ETH | USD` where and when it makes sense, don't need to do it all in one go. Glad you agree on this approach! I wasn't entirely sure but I'm pretty convinced now.

PR #711, opened 2022-03-21
Update to react-scripts@4.0.3

aeolianeth:
This is so we can use ESLINT_NO_DEV_ERRORS.

ESLint errors still happen, but we can keep deving,
improving productivity.


aeolianeth:
cc @johnnyd-eth @peripheralist

PR #736, opened 2022-03-27
V2 splits + fixes to V2 units

aeolianeth:

Fixes the way we create a display splits for V2 projects. I've left comments throughout to explain the big rocks.

Also fixes (I think) the units for every other value in V2 (reserved rate, discount rate, etc.). There's still some testing we'll have to do (like making sure payout recipients actually get the right amount of funds), but think we need to get this in ASAP to save us some sanity.

## How to test

You can test project creation to make sure you end up with the same splits as you defined.

Also try viewing `/#/v2/p/47`.

## Acceptance checklist


package.json
```diff
@@ -32,6 +32,7 @@
     "eslint-config-prettier": "^8.3.0",
     "ethereum-block-by-date": "^1.4.2",
     "ethers": "^5.1.0",
+    "lodash": "^4.17.21",
```
aeolianeth, line 35: I think we can use this elsewhere and it tree-shakes well. I don't have time tonight to do a bundle analysis, should be okay i think.

src/components/Landing/Payments.tsx
```diff
@@ -1,7 +1,7 @@
 import RichNote from 'components/shared/RichNote'
 import FormattedAddress from 'components/shared/FormattedAddress'
 import Loading from 'components/shared/Loading'
-import ProjectHandle from 'components/shared/ProjectHandle'
+import V1ProjectHandle from 'components/v1/shared/V1ProjectHandle'
```
aeolianeth, line 4: For clarity reasons. We'll need to come up with a strat for V2 (probably ENS name). Future problem.

src/components/v2/V2Project/index.tsx
```diff
@@ -46,8 +39,8 @@ export default function V2Project() {
         <Col md={12} xs={24}>
           <PayInputGroup
             PayButton={V2PayButton}
-            reservedRate={reservedRatePercent}
-            weight={weight}
+            reservedRate={fundingCycleMetadata?.reservedRate.toNumber()}
```
aeolianeth, line 42: smol fix, this takes a permyriad not a percent

src/contexts/v2/projectContext.ts
```diff
@@ -10,8 +10,8 @@ export type V2ProjectContextType = {
   fundingCycleMetadata: V2FundingCycleMetadata | undefined
   fundingCycle: V2FundingCycle | undefined
   queuedFundingCycle: V2FundingCycle | undefined
-  distributionLimit: string | undefined
-  queuedDistributionLimit: string | undefined
+  distributionLimit: BigNumber | undefined
```
aeolianeth, line 13: Fixes. These contract reads return BigNumber's and not hooks.

src/hooks/v2/contractReader/ProjectSplits.ts
```diff
@@ -20,5 +45,11 @@ export default function useProjectSplits({
       projectId && domain
         ? [projectId.toHexString(), domain, splitGroup]
         : null,
+    formatter: useCallback((value): Split[] => {
```
aeolianeth, line 48: The contract read returns BigNumbers, but we can't store that in Redux, and in general its less convenient. The maximum values for all of these properties _doesn't_ exceed the JS max value, so we're safe.

https://github.com/jbx-protocol/juice-contracts-v2/blob/3671aa3f10de661dad9e99db1ecc574a6f4a3b89/contracts/libraries/JBConstants.sol

src/utils/v2/math.ts
```diff
@@ -0,0 +1,91 @@
+import { BigNumber } from '@ethersproject/bignumber'
```
aeolianeth, line 1: I'm gonna write unit tests for all these functions in a follow-up.

src/utils/v2/math.ts
```diff
@@ -0,0 +1,91 @@
...
+export const MAX_REDEMPTION_RATE = TEN_THOUSAND
+export const MAX_DISCOUNT_RATE = ONE_BILLION
+export const SPLITS_TOTAL_PERCENT = ONE_BILLION
+
+/**
+ * Express a given discount rate (parts-per-billion) as a percentage.
+ * @param discountRate - discount rate as parts-per-billion.
+ * @returns {string} discount rate expressed as a percentage.
+ */
+export const formatDiscountRate = (discountRate: BigNumber) => {
```
aeolianeth, line 16: My idea with these functions is that they end up being more clear, despite a bit of duplication.

Instead of trying to remember what units the discount rate etc. is in, we just use the `formatDiscountRate` function (and friends).

src/components/v2/V2Project/index.tsx
johnnyd-eth: Nice

src/contexts/v2/projectContext.ts
johnnyd-eth: Good catch, my bad!

src/utils/v2/math.ts
johnnyd-eth: Yep I like it SGTM

johnnyd-eth (approved):
Yeah all looks great to me @aeolianeth, no comments from me. Agree we should get this out ASAP if you're happy with it

aeolianeth:
> Yeah all looks great to me @aeolianeth, no comments from me. Agree we should get this out ASAP if you're happy with it

send ittttt

PR #739, opened 2022-03-27
Bump @jbx-protocol/contracts-v2 to v0.0.13

aeolianeth:


aeolianeth:
@johnnyd-eth could you please review this and test if the reconfig stuff works? If it doesn't, I'd say merge this and lets look into fixing it in a follow-up 👍

johnnyd-eth:
@aeolianeth right now on this branch I can't load v2 projects locally. But the code here all looks great, so perhaps we just merge this and test the reconfig tx on Rinkeby (where v2 projects still work)?

aeolianeth:
> @aeolianeth right now on this branch I can't load v2 projects locally. I've `yarn install`ed, is there something else I need to do? Otherwise the code here looks great, so assuming it works for you, I've approved this thinking we perhaps just merge this and test the reconfig tx on Rinkeby?

@johnnyd-eth serves me right, i didn't actually test it locally 😂 . I've made a few changes to support the new contracts, keep in mind that all previous test projects on Rink have been blown away. Give `/#/v2/p/2` a crack if you like 👍 

I'll go ahead and merge this when pipeline is ✅

PR #780, opened 2022-04-09
Fix amount<->percent conversions on v2 payout modal 

johnnyd-eth:

- Fix amount<->percent conversions by using `feePercentage` in converter functions 
- Adds `currencyContext` to `projectPayoutMods` to make it actually version-agnostic
- Moves `PayoutModsList` back to v1 since it's got a lot of v1 stuff and not currently used in v2 

## Acceptance checklist

### DIFF
--- a/src/components/shared/forms/PayModsForm.tsx
+++ b/src/components/shared/forms/PayModsForm.tsx
@@ -8,9 +8,10 @@ import { useForm } from 'antd/lib/form/Form'
 import { V1CurrencyOption } from 'models/v1/currencyOption'
 import { PayoutMod } from 'models/mods'
 import { useContext, useLayoutEffect, useState } from 'react'
-import { fromWad } from 'utils/formatNumber'
+import { fromWad, perbicentToPercent } from 'utils/formatNumber'
 
 import { getTotalPercentage } from 'components/shared/formItems/formHelpers'
+import { CurrencyContext } from 'contexts/currencyContext'
 
 export default function PayModsForm({
   initialMods,
@@ -35,6 +36,10 @@ export default function PayModsForm({
   const {
     theme: { colors },
   } = useContext(ThemeContext)
+
+  const {
+    currencies: { ETH },
+  } = useContext(CurrencyContext)
 
   // Calculates sum of percentages of given payouts
   function calculateTotalPercentage(newMods: PayoutMod[] | undefined) {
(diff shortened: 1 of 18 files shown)


src/components/v2/V2Create/tabs/FundingTab/FundingTabContent.tsx
```diff
@@ -245,8 +250,8 @@ export default function FundingTabContent({
           <ProjectPayoutMods
             mods={splits.map(toMod)}
             target={target ?? '0'}
-            currency={toV1Currency(targetCurrency)}
-            feePerbicent={ETHPaymentTerminalFee}
+            currencyName={targetCurrency === ETH ? 'ETH' : 'USD'}
+            feePercentage={formatFee(ETHPaymentTerminalFee)}
```
aeolianeth, line 254:
```suggestion
            feePercentage={ETHPaymentTerminalFee ? formatFee(ETHPaymentTerminalFee) : undefined}
```

src/utils/v2/math.ts
```diff
@@ -99,7 +99,8 @@ export const redemptionRateFrom = (percentage: string) => {
  * @param feePerBillion - fee as parts-per-billion.
  * @returns {string} fee expressed as a percentage.
  */
-export const formatFee = (feePerBillion: BigNumber) => {
+export const formatFee = (feePerBillion: BigNumber | undefined) => {
+  if (!feePerBillion) return
```
aeolianeth, line 103: I kinda think this pattern throughout the codebase is a bit of a code small. I think e.g. we should conditionally render `ProjectPayoutMods` if `ETHPaymentTerminalFee` is defined, or at least do what I suggested in the above comment.

Let me know what you think

aeolianeth (changes requested):
@johnnyd-eth thanks heaps for fixing this, great catch and this solution is 👌 

couple comments

src/utils/v2/math.ts
johnnyd-eth: Bit on the nose I agree. Will change

src/components/shared/forms/PayModsForm.tsx
```diff
@@ -97,14 +102,14 @@ export default function PayModsForm({
         <FormItems.ProjectPayoutMods
           mods={mods}
           target={fromWad(target)}
-          currency={currency}
+          currencyName={currency === ETH ? 'ETH' : 'USD'}
```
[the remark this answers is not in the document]
johnnyd-eth: @aeolianeth This is a shared component, so I don't think so. But for the v1 specific components, yep

src/components/v1/PayoutModsList.tsx
```diff
@@ -228,8 +237,10 @@ export default function PayoutModsList({
               lockedMods={lockedMods}
               onModsChanged={setEditingMods}
               target={fromWad(fundingCycle.target)}
-              currency={fundingCycle.currency.toNumber() as V1CurrencyOption}
-              feePerbicent={feePerbicent}
+              currencyName={
+                fundingCycle.currency.toNumber() === ETH ? 'ETH' : 'USD'
```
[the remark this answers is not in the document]
johnnyd-eth: Yep

src/utils/v2/math.ts
aeolianeth: hahahahaha

aeolianeth: bit pongy

aeolianeth (approved):

Issue #634, opened 2022-03-11
[BUGish] Show error message when payment fails


aeolianeth:
## Description

When a user tries to pay a juicebox project and it fails, there is no feedback to the user. This leaves them confused.

## Proposal

Show a basic error message in the Payment modal

johnnyd-eth:
I thought I did this at some point, but we should also prompt wallet connect if no wallet connected and user attempts a payment

benschac:
Can I grab this issue?

johnnyd-eth:
> Can I grab this issue?

Yeah would be huge @benschac go for it!

benschac:
Quick questions:
> Show a basic error message in the Payment modal
I have been looking around for Modal Component Design (https://www.figma.com/file/8NjqzSpT8nO6MMlHbo8M5K/JBX-Component-Library?node-id=0%3A1) and Home page (https://www.figma.com/file/DczKN99Aaew2dL8Khtb4fE/JUICEBOX_HOMEPAGE?node-id=150%3A146) associated design for error messages.

Maybe I'm just looking in the wrong place. I haven't seen any design for an error message. 

Where could I find the error message design?

aeolianeth:
> Where could I find the error message design?

@benschac there isn't one yet!

I would suggest we can use the antd `Notification` (with an error state) for this.

https://ant.design/components/notification/

We can use the `error` variant of the notification in this case.

Let me know if you have any trouble setting that up, and I'll try and point you in the right direction!

zouantchaw:
is this still up for grabs?

benschac:
I was going to work on it this weekend.

benschac:
#738

benschac:
I have a couple of questions:

1. https://github.com/jbx-protocol/juice-interface/blob/main/src/hooks/Transactor.ts#L194. There's already a notif for the failure of transactions. It seems like this code meets the requirement of the issue when triggered.

<img height="400" src="https://user-images.githubusercontent.com/2502947/160292706-879afb68-9a6d-443d-a515-fc3ced9f1b28.jpg" />



2. Is there a way most folks on the team force a transaction error/failure to test/confirm functionality? I manually tested this failure state by throwing an error in the try-catch block in the screenshot above. I'm curious if the culprit might be the transaction not throwing an error on failure?

aeolianeth:
@benschac this is my bad, I should have been more specific in this issue description.

[...]

PR #789, opened 2022-04-10
Show error message when payment fails

benschac:
- Add an onError callback to the transaction options object. I took this suggestion from @aeolianeth in #634 after doing a little bit of investigation.

- Add a notif for the end user when the transaction is missing required parameters.

#### How to test:
- Add a true boolean to `PayV1ProjectTx.ts` hook where the new `onError` function is introduced.
- Click the payment button in the v1 payment modal (shown in gif) in `http://localhost:3000/#/p/<project>`.
- error message should be shown to user.

Gif of functionality:

## Acceptance checklist


benschac:
* main
  * **PR #789** <a href="https://app.graphite.dev/github/pr/jbx-protocol/juice-interface/789" target="_blank">Graphite</a>  (this PR)

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jbx-protocol/juice-interface/789?utm_source=stack-comment).

src/hooks/Transactor.ts
```diff
@@ -21,6 +21,7 @@ export type TransactorOptions = {
   onDone?: VoidFunction
   onConfirmed?: TransactorCallback
   onCancelled?: TransactorCallback
+  onError?: ErrorCallback
```
benschac, line 24: I poked around `lib/dom.d.ts` a bit and `ErrorCallback` looked like the most solid choice. Lmk if there's a better / specific `dom.ts` type that works here.

src/hooks/v1/transactor/PayV1ProjectTx.ts
```diff
@@ -23,7 +23,21 @@ export function usePayV1ProjectTx(): TransactorInstance<{
...
-      txOpts?.onDone?.()
+      let missingParam = !transactor
+        ? 'transactor'
+        : !projectId
+        ? 'projectId'
+        : !contracts?.TicketBooth
+        ? 'contracts.TicketBooth'
+        : !terminal?.version
+        ? 'terminal.version'
+        : null
```
benschac, line 34: I've seen a bunch of ternary in the code base and assumed this would be okay. I know there are a lot of opinions around nested ternary.

src/hooks/v1/transactor/PayV1ProjectTx.ts
```diff
@@ -23,7 +23,21 @@ export function usePayV1ProjectTx(): TransactorInstance<{
...
+
+      txOpts?.onError?.(
+        new DOMException(
```
benschac, line 37: `ErrorCallback` takes `DOMException`, I also didn't see it used anywhere else in the codebase. After looking around a bit: https://stackoverflow.com/questions/16142583/whats-the-difference-between-error-and-exception-in-javascript, Exception seemed like the right choice.

src/hooks/Transactor.ts
benschac: or in notify.d.ts

src/hooks/v1/transactor/PayV1ProjectTx.ts
aeolianeth: Appreciate the comment! I like this approach.

src/hooks/v1/transactor/PayV1ProjectTx.ts
aeolianeth: Well there ya go! Don't think I've specifically encounted DOMException before. LGTM

src/hooks/Transactor.ts
aeolianeth: Seems legit!

aeolianeth (approved):

aeolianeth:
@benschac thanks so much for this! Great improvement, I also really like that the error message is specific. We can do a better job of this site-wide.

I've added a merge commit to bring this up to date with `main` and will merge when pipeline succeeds 💪

PR #802, opened 2022-04-14
A bunch of random V2 create fixes

aeolianeth:

This PR is big and sweeping, sorry. But the main thing it fixes is the Tokens tab, as well as adds complete unit test coverage of our V2 math utils, and also uses these utils in a better way throughout.

I don't have time to add comments to this PR, hoping for your trust in the code, and probably best to just test the feature and let me know if anything is sus.

**Note** on UX, I've removed some of the opinionated defaults for form values (like reserved rate and discount rate). I think this is for the best, or at the least, it reduces complexity and bug risk in the implementation. If you feel strongly about keeping it lets chat in discord.

## Acceptance checklist

### DIFF
--- a/src/components/Landing/QAs.tsx
+++ b/src/components/Landing/QAs.tsx
@@ -123,7 +123,7 @@ export default function QAs() {
           For example, with a bonding curve of 70%, redeeming 10% of the token
           supply at any given time will claim around 7% of the total overflow.
         </Trans>,
-        <Trans>The rest is left to share between token hodlers.</Trans>,
+        <Trans>The rest is left to share between token holders.</Trans>,
         <Trans>
           For more info, check out this{' '}
           <ExternalLink href="https://youtu.be/dxqc3yMqi5M">
--- a/src/components/shared/Project/FundingCycleSection.tsx
+++ b/src/components/shared/Project/FundingCycleSection.tsx
@@ -6,7 +6,7 @@ import { useContext, useState } from 'react'
 
 import SectionHeader from 'components/shared/SectionHeader'
 
-type TabType = {
+export type TabType = {
   key: string
   label: string | JSX.Element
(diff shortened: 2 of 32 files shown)


johnnyd-eth:
Nice @aeolianeth. Got a few comments, but if you want to make an issue of them and we get to them in a follow up.

Screen Shot 2022-04-14 at 12 21 13 pm

- Discount rate should be disabled when funding duration is 0
- When reserved rate is enabled, value should go to 50% I think. Not 0

Screen Shot 2022-04-14 at 12 24 44 pm

- 'Disabled when no funding target is set.' remains after funding target has been set

- **Praise**: nice fix of the duration and time until next funding cycle
- **Question:** why are the switches removed from discount rate and redemption rate fields?
- **Idea:** maybe we should add a sentence in the token tab, something like, 'Once your project is deployed you can launch your own ERC20 token with a custom name and code (e.g. DOGE)'

aeolianeth:
@johnnyd-eth 

Let's do these ones: 
- Discount rate should be disabled when funding duration is 0
- 'Disabled when no funding target is set.' remains after funding target has been set

Let's leave everything else as-is for now. I'm concerned these forms are too complex from both UX and implementation PoV. We can push back on the design mockups where it makes sense to, and improve later on. Simple means we can iterate faster.

johnnyd-eth:
> 

@aeolianeth ok I've fixed and pushed fixes to those 2 points. Happy to merge this now?

johnnyd-eth (approved):

johnnyd-eth:
Screen Shot 2022-04-14 at 2 03 13 pm

- When discount rate disabled, should be either hide the slider or set it 0%? Personally would prefer hiding the sliders (in disabled redemp rate as well) because the 0% and 100% are meaningless in these disabled states

Issue #787, opened 2022-04-10
V2 Create fixups


johnnyd-eth:
**BUGS:** 

<img width="500" alt="Screen Shot 2022-04-10 at 2 05 10 pm" src="https://user-images.githubusercontent.com/96150256/162619656-798def4c-2288-4bd4-a9da-13ba95152794.png">

- [x] Time till next funding cycle super messed up
- [x] DiscountRate not updating in the preview at all
- [x] Hide upcoming funding tab in preview 

- [x] Rules tab not loading initial redux values properly:
1. Save a redux state with any change to rules tab
2. Refresh page and go back to rules tab, will see default states loaded 

**NON-URGENT:** 

<img width="619" alt="Screen Shot 2022-04-12 at 2 43 19 pm" src="https://user-images.githubusercontent.com/96150256/162976204-c7b49357-0598-4eec-8a3f-f4aa0d0f8689.png">

- Should we remove distribute buttons on payouts and reserved cards in the preview?

- [ ] Duration only rounds to hrs and shows '- -' when less than 1hr

<img width="576" alt="Screen Shot 2022-04-16 at 5 49 48 pm" src="https://user-images.githubusercontent.com/96150256/163684059-ae59b713-992c-4a4c-9a8b-ce78140890ad.png">

**SEE COMMENTS FOR MORE**

johnnyd-eth:
@aeolianeth More on this:

<img width="551" alt="Screen Shot 2022-04-12 at 10 57 36 am" src="https://user-images.githubusercontent.com/96150256/162934393-adbddc81-bd1f-45d7-affe-9ac34570e3b4.png">

- '197532 days' until next FC even though duration is 8hrs?
- I think duration shouldn't round to hrs while we're supporting seconds (not pressing though)
- Discount rate, redemp rate, reserved rate, pausePay and allowMinting not updating in the FC card. Everything else is

johnnyd-eth:
<img width="615" alt="Screen Shot 2022-04-15 at 12 20 09 am" src="https://user-images.githubusercontent.com/96150256/163492465-fd940d33-6cdf-4268-bc3f-8affcb4e9f3f.png">

- Spacing here. 

- Make preview a max-height of the window height and scrollable with that window

johnnyd-eth:
[...]

PR #814, opened 2022-04-15
Create switch heading and use it some places

aeolianeth:

- fixes the rules form
- adds a SwitchHeading component, uses it for the Token tab, hopefully fixes Token tab fo good.
- resets Token tab values from the Funding tab where appropriate

I'm fairly sure this checks off https://github.com/jbx-protocol/juice-interface/issues/787, but pls note anything i've missed or that is sus.

changes._

## Acceptance checklist

### DIFF
--- a/src/components/v2/V2Create/forms/FundingForm/index.tsx
+++ b/src/components/v2/V2Create/forms/FundingForm/index.tsx
@@ -1,6 +1,6 @@
 import { Trans } from '@lingui/macro'
 
-import { Button, Form, Space, Switch } from 'antd'
+import { Button, Form, Space } from 'antd'
 
 import { useCallback, useContext, useLayoutEffect, useState } from 'react'
 
@@ -43,6 +43,8 @@ import BudgetTargetInput from 'components/shared/inputs/BudgetTargetInput'
 import { Link } from 'react-router-dom'
 
 import FormItemWarningText from 'components/shared/FormItemWarningText'
+
+import SwitchHeading from 'components/shared/SwitchHeading'
 
 import { ETH_TOKEN_ADDRESS } from 'constants/v2/juiceboxTokens'
 
@@ -188,22 +190,19 @@ export default function FundingForm({ onFinish }: { onFinish: VoidFunction }) {
           color: theme.colors.text.primary,
         }}
       >
-        <div style={{ display: 'flex' }}>
-          <Switch
-            checked={durationEnabled}
-            onChange={checked => {
-              setDurationEnabled(checked)
-              if (!checked) {
-                fundingForm.setFieldsValue({ duration: '0' })
-              }
-              fundingForm.setFieldsValue({ duration: '30' })
-            }}
-            style={{ marginRight: 10 }}
-          />
-          <h3>
-            <Trans>Funding cycles</Trans>
-          </h3>
-        </div>
+        <SwitchHeading
+          checked={durationEnabled}
+          onChange={checked => {
+            setDurationEnabled(checked)
+
+            if (!checked) {
+              fundingForm.setFieldsValue({ duration: '0' })
+            }
+            fundingForm.setFieldsValue({ duration: '30' })
+          }}
+        >
+          <Trans>Funding cycles</Trans>
+        </SwitchHeading>
 
         <Space size="middle" direction="vertical">
           <div>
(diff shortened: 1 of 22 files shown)


src/components/v2/V2Create/forms/FundingForm/index.tsx
```diff
@@ -153,6 +159,24 @@ export default function FundingForm({ onFinish }: { onFinish: VoidFunction }) {
       )
       dispatch(editingV2ProjectActions.setDuration(durationInSeconds ?? '0'))
 
+      // reset discount rate if duration is 0
+      if (!durationInSeconds || durationInSeconds === '0') {
```
aeolianeth, line 163: if someone goes to token tab, then back to funding tab and makes changes, then we might need to reset some values. we do that below

src/components/v2/V2Create/forms/RulesForm/index.tsx
```diff
@@ -20,52 +20,35 @@ import {
...
-
-  const onFormSaved = useCallback(
-    (fields: RulesFormFields) => {
-      dispatch(editingV2ProjectActions.setPausePay(fields.pausePay))
-      dispatch(editingV2ProjectActions.setAllowMinting(fields.allowMinting))
-      dispatch(editingV2ProjectActions.setBallot(ballotStrategy.address))
-      onFinish?.()
-    },
-    [dispatch, onFinish, ballotStrategy],
+  const [pausePay, setPausePay] = useState<boolean>(
```
aeolianeth, line 39: Totally bail on using the `form`. It's a massive pain to get working, there's no need for it here anyway.

src/components/v2/V2Create/forms/TokenForm/index.tsx
```diff
@@ -84,49 +127,25 @@ export default function TokenForm({ onFinish }: { onFinish: VoidFunction }) {
...
+  ])
 
   return (
-    <Form
-      form={tokenForm}
-      initialValues={initialValues}
-      layout="vertical"
-      onFinish={onTokenFormSaved}
-    >
+    <Form layout="vertical" onFinish={onTokenFormSaved}>
```
aeolianeth, line 148: Bail on `form` here too lol

src/components/v2/V2Create/tabs/FundingCycleTab/FundingCycleTabContent.tsx
```diff
@@ -107,6 +107,7 @@ export default function FundingCycleTabContent({
           setFundingDrawerVisible(false)
         }}
         getContainer={false}
+        destroyOnClose
```
aeolianeth, line 110: this is needed to we force to load from Redux every time the drawer is opened.

src/hooks/v2/contractReader/ProjectReservedTokens.ts
```diff
@@ -14,8 +14,8 @@ export default function useProjectReservedTokens({
     contract: V2ContractName.JBController,
     functionName: 'reservedTokenBalanceOf',
     args:
-      projectId && reservedRate
-        ? [BigNumber.from(projectId).toHexString(), reservedRate]
+      projectId && reservedRate && !projectId.eq(0)
```
aeolianeth, line 17: so this hook doesn't try to load every time (even ont eh create form)

src/components/v2/V2Create/forms/FundingForm/index.tsx
johnnyd-eth: Yeah got it

src/components/v2/V2Create/forms/RulesForm/index.tsx
johnnyd-eth: Yeah lol don't blame you I've been tempted. Good call here for sure

src/components/v2/V2Create/tabs/FundingCycleTab/FundingCycleTabContent.tsx
johnnyd-eth: Yep

johnnyd-eth:
Nice @aeolianeth this looks rock-solid in the UI, I couldn't find any  Only question I have:

- Will put this to everyone else (I'm still undecided as to what I prefer), but wondering if enabling the switch should set the field to a value (e.g. discount rate to 10% rather than staying at 0%). It feels weird to 'enable' a field and have it do nothing

johnnyd-eth:
Also, I removed the close #787 because I left a few more comments on it outside the description which I'll work on today

johnnyd-eth:
But overall this is great! Approving and happy if we do the setting value thing in a follow-up if we end up deciding to do that

johnnyd-eth (approved):

PR #832, opened 2022-04-20
Add V2 feature flag switch

aeolianeth:

This PR introduces a bare-bones feature flag system to the codebase. We store feature flag states in localstorage.

To better control the rollout of V2, all V2 features (project creation and viewing) is now behind the feature flag.
When the `ENABLE_V2` flag is enabled, we will:
- make the **Design your project** button on the homepage redirect to the V2 create page
- hide the V1 create form on the homepage
- enable the V2 project page.

When it's disabled, we show the site as it is now (with no v2 features, only v1).

The feature flag is **disabled by default** for mainnet but **enabled by default** for Rinkeby.

There is a toggle on the very bottom of the homepage to enable it.

https://user-images.githubusercontent.com/94939382/164263055-88f0070a-1985-4c26-8329-37cce1546690.mp4

## Acceptance checklist


aeolianeth:
@johnnyd-eth for your review!

cc @peripheralist for awareness

peripheralist:
Really nice work with the feature flag pattern! Love it

Wondering if we should add a "caution" type banner to the v2 create form just in case someone who doesn't understand what they're doing decides to be ballsy with the toggle

aeolianeth:
> Really nice work with the feature flag pattern! Love it
> 
> Wondering if we should add a "caution" type banner to the v2 create form just in case someone who doesn't understand what they're doing decides to be ballsy with the toggle

<3

I'm happy to leave it for now. Would like to remove eventually of course. Maybe that can be part of the feature flag cleanup: https://github.com/jbx-protocol/juice-interface/issues/833

johnnyd-eth:
Very slick @aeolianeth. Gave it a good test and functions perfectly. I'm just thinking we should be extra cautious for now and add another sentence to the toggle tooltip, something like: 'The V2 frontend is still in development - some features are missing and there are known bugs.'

aeolianeth:
@johnnyd-eth LOVE those suggestions. Adressed, came up with a better approach to the toggle too. Now done at the route level.

johnnyd-eth:
@aeolianeth nice functions perfectly now. I was thinking of adding this sentence to this tooltip here though, since it really applies to V2 as a whole not just the v2 project page. And if anyone comes across it, it will likely be from here. 

Screen Shot 2022-04-22 at 1 14 00 am

Looks great otherwise so I'll approve but I think we should show a caution message for the switch on the homepage

johnnyd-eth (approved):

PR #845, opened 2022-04-21
Redo distribution section and modal for v2

johnnyd-eth:

The previous version of this is `ProjectPayoutMods`. It's a massive file containing 3 main components: the payout section itself, the payout cards and the modal that appears when you click 'Add a payout' or click to edit a card. Also, in this case there a single modal element being rendered, who's state (all its fields, add/edit mode) was being altered from the parent component. 

I've separated these components out in the `v2/shared/DistributionSplitsSection` folder. I've moved as much logic (field setting, saving fields, etc.) into the modal component. To avoid setting the modal's state from the parent `index` file, I've just put an 'Add' version of the modal in the `index` and an 'Edit' version of it in each of the `DistributionSplitCard`s

changes._

## Acceptance checklist


src/components/shared/formItems/formHelpers.tsx
```diff
@@ -58,63 +50,6 @@ export function validateEthAddress(
...
-) {
-  const targetSubFeeBN = amountSubFee(
-    parseWad(stripCommas(target)),
-    percentToPerbicent(feePercentage),
-  )
-  const targetSubFee = stripCommas(formatWad(targetSubFeeBN) ?? '0')
-  const percent = (((amount ?? 0) * 1.0) / parseFloat(targetSubFee)) * 100
-  return parseFloat(percent.toFixed(8))
-}
-
```
johnnyd-eth, line 117: Moved all these v1 specific functions to a new `utils/v1/payouts` file

src/components/v1/ProjectPayoutMods.tsx
```diff
@@ -437,15 +436,15 @@ export default function ProjectPayoutMods({
           }}
           block
         >
-          <Trans>Add a split</Trans>
+          <Trans>Add a payout</Trans>
```
johnnyd-eth, line 439: Moved the `ProjectPayoutMods` from shared to v1. Also reverted back wording changes that we'd made for v2

src/components/v2/shared/DistributionSplitsSection/DistributionSplitCard.tsx
```diff
@@ -0,0 +1,200 @@
...
+  CrownFilled,
+  LockOutlined,
+  CloseCircleOutlined,
+} from '@ant-design/icons'
+import { V2ProjectContext } from 'contexts/v2/projectContext'
+
+import DistributionSplitModal from './DistributionSplitModal'
+import { CurrencyName } from 'constants/currency'
+
+export default function DistributionSplitCard({
```
johnnyd-eth, line 18: The cards that appear in this list
Screen Shot 2022-04-24 at 10 29 18 am

src/components/v2/shared/DistributionSplitsSection/DistributionSplitModal.tsx
```diff
@@ -0,0 +1,301 @@
...
+  projectId: string
+  beneficiary: string
+  percent: number
+  amount: number
+  lockedUntil: moment.Moment
+}
+
+type SplitType = 'project' | 'address'
+
+export default function DistributionSplitModal({
```
johnnyd-eth, line 33: The modal that appears when you click a card ('Edit' mode) or click 'Add a split' ('Add' mode)

src/components/v2/shared/DistributionSplitsSection/DistributionSplitCard.tsx
```diff
@@ -0,0 +1,200 @@
...
+            onSplitsChanged([
+              ...splits.slice(0, splitIndex),
+              ...splits.slice(splitIndex + 1),
+            ])
+            e.stopPropagation()
+          }}
+          icon={<CloseCircleOutlined />}
+        />
+      )}
+      <DistributionSplitModal
```
johnnyd-eth, line 189: Instance of `DistributionSplitModal` in each card, rather than a single one from the parent `index`

src/components/v1/ProjectPayoutMods.tsx
```diff
@@ -437,15 +436,15 @@ export default function ProjectPayoutMods({
...
         </Button>
       </Space>
 
       <Modal
```
johnnyd-eth, line 443: The original `DistributionSplitModal`, which is now it's own file that handles everything itself (doesn't rely on `index` for anything other than finally setting the split)

src/components/v2/shared/DistributionSplitsSection/DistributionSplitModal.tsx
```diff
@@ -0,0 +1,301 @@
...
+      beneficiary,
+      percent,
+      handle,
+      lockedUntil,
+      preferClaimed: true,
+      projectId: undefined,
+      allocator: undefined,
+    }
+
+    onSplitsChanged(
```
johnnyd-eth, line 90: Only communicates upstream to set the new split by calling `onSplitsChanged`

src/components/v2/shared/SplitItem.tsx
```diff
@@ -31,6 +32,7 @@ export default function SplitItem({
   showSplitValue: boolean
   valueSuffix?: string | JSX.Element
   valueFormatProps?: { precision?: number }
+  onClick?: VoidFunction
```
johnnyd-eth, line 35: This is unnecessary, will remove

src/components/v2/shared/DistributionSplitsSection/DistributionSplitModal.tsx
```diff
@@ -0,0 +1,381 @@
...
+import * as Moment from 'moment'
+import moment from 'moment'
+import TooltipLabel from 'components/shared/TooltipLabel'
+
+import { CurrencyName } from 'constants/currency'
+
+export type AddOrEditSplitFormFields = {
+  projectId: string
+  beneficiary: string
+  percent: number
```
johnnyd-eth, line 44: Only need these fields for validation

src/components/v2/shared/DistributionSplitsSection/DistributionSplitModal.tsx
```diff
@@ -0,0 +1,381 @@
...
+          </div>
+        </Form.Item>
+        <Form.Item
+          name="lockedUntil"
+          label={t`Lock until`}
+          extra={t`If locked, this can't be edited or removed until the lock expires or the funding cycle is reconfigured.`}
+        >
+          <DatePicker
+            value={lockedUntil}
+            disabledDate={disabledDate}
```
johnnyd-eth, line 374: Now can't set lockedUntil before current day

src/components/v1/ProjectPayoutMods.tsx
```diff
@@ -437,15 +436,15 @@ export default function ProjectPayoutMods({
...
-        title={modalMode === 'Edit' ? t`Edit existing split` : t`Add a split`}
+        title={modalMode === 'Edit' ? t`Edit existing split` : t`Add a payout`}
         visible={editingModIndex !== undefined}
         onOk={setReceiver}
-        okText={modalMode === 'Edit' ? t`Save split` : t`Add split`}
+        okText={modalMode === 'Edit' ? t`Save split` : t`Add payout`}
```
aeolianeth, line 447: **suggestion** should this be `Save payout`?

```suggestion
        okText={modalMode === 'Edit' ? t`Save payout` : t`Add payout`}
```

aeolianeth: Should every reference to `split` in this file be changed to `payout`?

src/components/v2/V2Create/forms/FundingForm/index.tsx
```diff
@@ -98,6 +108,22 @@ export default function FundingForm({ onFinish }: { onFinish: VoidFunction }) {
       fundAccessConstraints,
     )
 
+  const { editableSplits, lockedSplits } = useMemo(() => {
```
aeolianeth, line 111: **suggestion** maybe can type these, then we won't need `[] as Split[]`

```suggestion
  const { editableSplits, lockedSplits } : { editableSplits: Split[], lockedSplits: Split[] } = useMemo(() => {
```

src/components/v2/V2Create/forms/FundingForm/index.tsx
```diff
@@ -63,21 +71,23 @@ type FundingFormFields = {
...
-  const {
-    currencies: { ETH },
-  } = useContext(CurrencyContext)
-
   const dispatch = useAppDispatch()
 
   const [fundingForm] = Form.useForm<FundingFormFields>()
 
   const [splits, setSplits] = useState<Split[]>([])
+  // Must differentiate between splits loaded from redux and
```
aeolianeth, line 86: **praise** awesome

src/components/v2/V2Create/forms/FundingForm/index.tsx
```diff
@@ -134,7 +160,8 @@ export default function FundingForm({ onFinish }: { onFinish: VoidFunction }) {
   }, [fundingForm, fundingCycleData, fundAccessConstraint, payoutGroupedSplits])
 
   const onFundingFormSave = useCallback(
-    (fields: FundingFormFields) => {
+    async (fields: FundingFormFields) => {
+      await fundingForm.validateFields()
```
aeolianeth, line 164: is the form not already validated at this point? Using the `onFinish` prop on the form is meant to validate before the `onFinish` handler (this function) is called: https://github.com/jbx-protocol/juice-interface/pull/845/files#diff-54e2ecc8080eb485aa06e45b065668919de95ef981ee2178a191d5971bb524a7L213

could def be broken though knowing `antd`

src/components/v2/V2Create/forms/FundingForm/index.tsx
```diff
@@ -376,20 +422,31 @@ export default function FundingForm({ onFinish }: { onFinish: VoidFunction }) {
...
             />
+            <Form.Item
+              name={'totalSplitsPercentage'}
+              rules={[
+                {
+                  validator: validateTotalSplitsPercentage,
+                },
+              ]}
+            >
+              {/* Added a hidden input here because Form.Item needs
```
aeolianeth, line 445: **suggestion** why not a state variable for `totalSplitsPercentage`?

Or, could we compute `totalSplitsPercentage` in the validator, based on the sum of all `editingSplits` percentages?

src/components/v2/shared/DistributionSplitsSection/DistributionSplitCard.tsx
aeolianeth: THANK YOU for screenshot <3

src/components/v2/shared/DistributionSplitsSection/DistributionSplitCard.tsx
```diff
@@ -0,0 +1,221 @@
...
+        ) : null}
+      </Space>
+
+      {isLocked ? (
+        <LockOutlined style={{ color: colors.icon.disabled }} />
+      ) : (
+        <Button
+          type="text"
+          onClick={e => {
+            onSplitsChanged([
```
aeolianeth, line 200: **suggestion**: I don't think this component should know anything about `splits`: I think there's a solution where we might:
- add an `onDelete` prop that takes a function
- Call `onDelete()` here
- in the parent component, handle `onDelete` with this same logic below

Spreading logic across components is something we want to avoid for maintainability.

aeolianeth (commented):
@johnnyd-eth i staretd reviewing this but had to dip, will get back to it after our call.

src/utils/v2/distributions.ts
```diff
@@ -0,0 +1,96 @@
...
+}
+
+/**
+ * Gets distribution amount from percent of the distribution limit and does not apply
+ * any fee
+ * @param percent {float} - value as a percentage.
+ * @param distributionLimit string (hexString)
+ * @returns {number} distribution amount
+ */
+export function getDistributionAmountFromPercentBeforeFee({
```
aeolianeth, line 45: **suggestions**: these functions are really just generic math functions that don't concern themselves with business logic (i.e. distributionLimit)

This naming feels a bit verbose - wdyt about making this a generic `percentOf` function:

```js
const percentOf: (value: number, percent: number) => parseFloat(
    ((percent / 100) * value).toFixed(8)
  )
```

I know we do it in other places, but we should avoid any `undefined`-related logic in these functions. These functions should be as simple as possible.

aeolianeth (changes requested):

src/components/v2/shared/DistributionSplitsSection/DistributionSplitModal.tsx
aeolianeth: fresh

src/components/v2/shared/DistributionSplitsSection/DistributionSplitModal.tsx
```diff
@@ -0,0 +1,381 @@
...
+          <Form.Item
+            label={t`Amount`}
+            className="ant-form-item-extra-only"
+            extra={
+              feePercentage && percent && !(percent > 100) ? (
+                <>
+                  {editingSplitType === 'address' ? (
+                    <div>
+                      <Trans>
+                        Payee will receive{' '}
```
aeolianeth, line 286: **suggestion**: maybe for brevity/clarity it should be:

```suggestion
```

And we add the `(?)` icon with a tooltip:

`13ETH after 2.5% JBX fee (?)`

The tooltip says something like:

`Payouts to Ethereum addresses incur a 2.5% fee. Your project will receive JBX in return at the current issuance rate.`

src/components/v2/shared/DistributionSplitsSection/DistributionSplitModal.tsx
```diff
@@ -0,0 +1,381 @@
...
+              }}
+            />
+          </Form.Item>
+        )}
+        {editingSplitType === 'project' ? (
+          <FormItems.EthAddress
+            name="beneficiary"
+            defaultValue={beneficiary}
+            formItemProps={{
+              label: t`Token beneficiary`,
```
aeolianeth, line 266:
```suggestion
              label: t`Token beneficiary address`,
```

src/components/v2/shared/DistributionSplitsSection/DistributionSplitModal.tsx
```diff
@@ -0,0 +1,381 @@
...
+              extra: t`The address that should receive the tokens minted from paying this project.`,
```
aeolianeth, line 267:
```suggestion
              extra: t`This address will receive the tokens minted from paying this project.`,
```

src/components/v2/shared/DistributionSplitsSection/DistributionSplitModal.tsx
```diff
@@ -0,0 +1,381 @@
...
+            <TooltipLabel
+              label={<Trans>Percent</Trans>}
+              tip={
+                distributionLimitIsInfinite ? (
+                  <Trans>
+                    Percent this payee will receive of all funds raised.
+                  </Trans>
+                ) : (
+                  <Trans>
+                    Percent of the distribution limit this payee will receive.
```
aeolianeth, line 337:
```suggestion
                    Percentage of the distribution limit this payee will receive.
```

src/components/v2/shared/DistributionSplitsSection/DistributionSplitModal.tsx
```diff
@@ -0,0 +1,381 @@
...
+        ) : null}
+
+        <Form.Item
+          label={
+            <TooltipLabel
+              label={<Trans>Percent</Trans>}
+              tip={
+                distributionLimitIsInfinite ? (
+                  <Trans>
+                    Percent this payee will receive of all funds raised.
```
aeolianeth, line 333:
```suggestion
                    Percentage this payee will receive of all funds raised.
```

src/components/v2/shared/DistributionSplitsSection/DistributionSplitModal.tsx
```diff
@@ -0,0 +1,381 @@
...
+                accessory={<InputAccessoryButton content={currencyName} />}
+              />
+            </div>
+          </Form.Item>
+        ) : null}
+
+        <Form.Item
+          label={
+            <TooltipLabel
+              label={<Trans>Percent</Trans>}
```
aeolianeth, line 329: WDYT about

```suggestion
              label={<Trans>Percentage of distribution limit</Trans>}
```

src/components/v1/ProjectPayoutMods.tsx
```diff
@@ -591,8 +592,8 @@ export default function ProjectPayoutMods({
           </Form.Item>
           <Form.Item
             name="lockedUntil"
-            label="Lock until"
-            extra="If locked, this can't be edited or removed until the lock expires or the funding cycle is reconfigured."
+            label={t`Lock until`}
+            extra={t`If locked, this can't be edited or removed until the lock expires or the funding cycle is reconfigured.`}
```
aeolianeth, line 596:
```suggestion
            extra={<Trans>If locked, this split can't be edited or removed until the lock expires or the funding cycle is reconfigured.</Trans>}
```

src/components/v2/shared/DistributionSplitsSection/DistributionSplitModal.tsx
```diff
@@ -0,0 +1,381 @@
...
+            onAddressChange={beneficiary => {
+              setBeneficiary(beneficiary)
+            }}
+          />
+        ) : null}
+
+        {/* Only show amount input if project distribution limit is not infinite */}
+        {!distributionLimitIsInfinite ? (
+          <Form.Item
+            label={t`Amount`}
```
aeolianeth, line 278:
```suggestion
            label={t`Payout amount`}
```

src/components/v2/shared/DistributionSplitsSection/DistributionSplitModal.tsx
```diff
@@ -0,0 +1,381 @@
...
+            }}
+            onAddressChange={(beneficiary: string) =>
+              setBeneficiary(beneficiary)
+            }
+          />
+        ) : (
+          <Form.Item
+            name={'projectId'}
+            rules={[{ validator: validateProjectId }]}
+            label={t`Project ID`}
```
aeolianeth, line 248:
```suggestion
            label={t`Juicebox project ID`}
```

src/components/v2/V2Create/forms/FundingForm/index.tsx
johnnyd-eth: You are correct. Removed

src/components/v2/V2Create/forms/FundingForm/index.tsx
johnnyd-eth: Yeah the field and a specific validator for it to show the error message where we want it though

src/components/v2/shared/DistributionSplitsSection/DistributionSplitCard.tsx
johnnyd-eth: Yeah there are other reasons we needs splits to be passed down though, mainly validating for 'same address already added' in the payout modal. So i figured not worth another another `onDelete` prop if we need `splits`

src/components/v2/shared/DistributionSplitsSection/DistributionSplitModal.tsx
johnnyd-eth: Yeah love it

src/components/v1/ProjectPayoutMods.tsx
johnnyd-eth: Thought I'd just keep all v1 wording the same. Saw I missed changing some back- they're all 'payout' now

src/utils/v2/distributions.ts
johnnyd-eth: Yeah good call. Changed the function name mentioned and removed undefined's from those functions

aeolianeth:
@johnnyd-eth lfg on this one

PR #876, opened 2022-04-27
update eslint configuration

benschac:

This is the last PR in the diff stack and the spiciest.

**What** 
ESlint had a major breaking change under the hood in RCA (https://github.com/facebook/create-react-app/blob/main/CHANGELOG.md#highlights) from v4 to v5. CRA now uses Eslint 8. Additionally, there are breaking changes in the react-supported ESLint plugins. It's a pretty common problem when upgrading: https://stackoverflow.com/questions/58065765/eslint-jest-globals-environment-key-unknown

- Deleted the old ESlint modules and ran through the ESlint getting started flow. https://eslint.org/docs/user-guide/getting-started.
- Saved the old config and added back as many rules as possible without breaking the project.

**Why**
ESLint needed to be upgraded because of the breaking change to RCA in `react-scripts` v5 using ESLint version 8.

**Additional Notes**
Why this change is spicy: I upgraded ESLint and there are a bunch of linting rules that the old configuration didn't have by default, which now ships with the default configuration:

```
"eslint:recommended",
"plugin:react/recommended",
"plugin:@typescript-eslint/recommended",
 ```
 
 The goal of this PR was to just get the project running without errors. I didn't make any assumptions on rules that the team would or wouldn't want to include in the project because I know have enough context into what y'all would like to have linted and then fixed, so I just added eslint comments to turn off errors I wasn't sure the team wanted to support.

For the team: cc @johnnyd-eth @peripheralist @aeolianeth -- which ESLint ignore comments should we fix and support moving forward?

I can add those fixes in another PR or just fix them here.

N/A, but the project now runs like it normally would without any errors.

## Acceptance checklist

### CODE REVIEW

**Comment by benschac** [comment]
* main
  * **PR #867** <a href="https://app.graphite.dev/github/pr/jbx-protocol/juice-interface/867" target="_blank">Graphite</a> 
    * **PR #868** <a href="https://app.graphite.dev/github/pr/jbx-protocol/juice-interface/868" target="_blank">Graphite</a> 
      * **PR #869** <a href="https://app.graphite.dev/github/pr/jbx-protocol/juice-interface/869" target="_blank">Graphite</a> 
        * **PR #870** <a href="https://app.graphite.dev/github/pr/jbx-protocol/juice-interface/870" target="_blank">Graphite</a> 
          * **PR #871** <a href="https://app.graphite.dev/github/pr/jbx-protocol/juice-interface/871" target="_blank">Graphite</a> 
            * **PR #876** <a href="https://app.graphite.dev/github/pr/jbx-protocol/juice-interface/876" target="_blank">Graphite</a>  (this PR)

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jbx-protocol/juice-interface/876?utm_source=stack-comment).

File: package.json [hunk 1]
```diff
@@ -93,11 +93,15 @@
     "@types/react": "^17",
     "@types/react-dom": "^17",
     "@types/react-router-dom": "^5.1.7",
+    "@typescript-eslint/eslint-plugin": "^5.21.0",
+    "@typescript-eslint/parser": "^5.21.0",
     "browserify-fs": "^1.0.0",
     "browserify-zlib": "^0.2.0",
     "crypto-browserify": "^3.12.0",
+    "eslint": "^8.14.0",
```
benschac, line 101: I'm not 100% if we need to install ESLint here because react-scripts is using ESLint8 under the hood.

benschac: I'm going to test this locally, and see if it can be removed.

.eslintrc.json
```diff
@@ -1,8 +1,34 @@
...
+    },
+    "ecmaVersion": "latest",
+    "sourceType": "module"
+  },
+  "plugins": ["react", "@typescript-eslint", "import", "react-hooks"],
   "rules": {
+    "@typescript-eslint/no-var-requires": "off",
+    "@typescript-eslint/ban-types": "off",
+    "react/no-children-prop": "off",
+    "react/no-unescaped-entities": "off",
```
benschac, line 25: These come by default with `plugin:react/recommended` I just turned them off, because they caused some errors.

.eslintrc.json
```diff
@@ -1,8 +1,34 @@
...
+    "@typescript-eslint/no-inferrable-types": "off",
+    "react/prop-types": "off",
+    "prefer-const": "off",
+    "jsx-a11y/img-redundant-alt": "off",
+    "react/react-in-jsx-scope": "off",
+    "@typescript-eslint/ban-ts-comment": "off",
```
benschac, line 31: These come by default with plugin:react/recommended I just turned them off, because they caused some errors.

johnnyd-eth (approved):
Very nice @benschac thanks for picking this up. All LGTM but I'm gonna defer to @aeolianeth  on the question of ignore comments - think he'll have a better answer than myself. Will also wait to get another set of eyes before we merge

package.json
benschac: removed

aeolianeth:
@benschac thanks for the thoughtfulness with this one!

I'm fine with disabling those new rules at the config level.

I think we should disabling eslint for specific lines though. Most of these look pretty trivial to fix (although I could be wrong) - would it be much of a lift to fix these and remove the `disable-next-line` comments?

benschac:
Yeah, I can add another PR to the stack with the fixes.

src/components/shared/IssueTicketsButton.tsx
```diff
@@ -0,0 +1,137 @@
+import { Button, Form, Input, Space, Tooltip } from 'antd'
```
benschac, line 1: This component made its way in here during a re-base. I need to fix this. Sorry!

benschac:
This needs a bit more work. I wish there were an easy way to demote back into a WIP PR.

PR #919, opened 2022-05-03
Fix beneficiary not being set to active wallet

### LINKED ISSUE #860: Beneficiary bug in V2
### repro

- user connects to wallet A, loads project page
- user connects to wallet B (without refreshing)
- pays
- beneficiary is still set to wallet A

### possible problem

It's because the beneficiary state variable defaults to userAddress but it isn't aware of the wallet changes so doesn't trigger a rerender

### possible solution

- make hook for wallet address so that rerenders are triggered

jmilldotdev:

Closes #860. We can get away without using a custom hook for the wallet address and simply using `useEffect` since we're already reacting to changes in address using the network context, but it could be an interesting future feature. I'm just not entirely sure how to implement it.

## Acceptance checklist

- [ x ] I have evaluated the [Approval Guidelines](https://github.com/jbx-protocol/juice-interface/blob/main/CONTRIBUTING.md#approval-guidelines) for this PR.
- [ x ] I have tested this PR in [all supported browsers](https://github.com/jbx-protocol/juice-interface/blob/main/CONTRIBUTING.md#supported-browsers).

### DIFF
--- a/src/components/v2/V2Project/V2PayButton/V2ConfirmPayModal/index.tsx
+++ b/src/components/v2/V2Project/V2PayButton/V2ConfirmPayModal/index.tsx
@@ -9,7 +9,7 @@ import { NetworkContext } from 'contexts/networkContext'
 import { useCurrencyConverter } from 'hooks/v1/CurrencyConverter'
 import { V2ProjectContext } from 'contexts/v2/projectContext'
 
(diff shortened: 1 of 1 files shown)


johnnyd-eth:
Nice nice LGTM

johnnyd-eth (approved):

PR #937, opened 2022-05-05
Add my projects tab and query

### LINKED ISSUE #875: Add page for projects the connected wallet is owner of.
## Description

_A clear and concise description of what could be better._

## Proposal

_A description of the change you'd like to see._

## Screenshots

_If applicable, add screenshots to help explain your problem._

jmilldotdev:

Closes #875. This is my first time working with a graph query so let me know if this could be written better, or if this should be handled differently in the UI.

image

image

## Acceptance checklist

- [ x ] I have evaluated the [Approval Guidelines](https://github.com/jbx-protocol/juice-interface/blob/main/CONTRIBUTING.md#approval-guidelines) for this PR.
- [ x ] I have tested this PR in [all supported browsers](https://github.com/jbx-protocol/juice-interface/blob/main/CONTRIBUTING.md#supported-browsers).


src/components/Projects/MyProjects.tsx
```diff
@@ -0,0 +1,68 @@
...
+        (projects.length === 0 ? (
+          <div
+            style={{
+              textAlign: 'center',
+              color: colors.text.disabled,
+              padding: 20,
+            }}
+            hidden={isLoading}
+          >
+            <Trans>You haven't created any projects yet.</Trans>
```
aeolianeth, line 43: **praise** Love!

What you think about adding a button here to create a project? Literally

`Create a project` => links to create page

src/components/Projects/MyProjects.tsx
```diff
@@ -0,0 +1,68 @@
...
+export default function MyProjects() {
+  const { userAddress } = useContext(NetworkContext)
+
+  const { data: projects, isLoading } = useMyProjectsQuery(userAddress)
+
+  const {
+    theme: { colors },
+  } = useContext(ThemeContext)
+
+  return (
```
aeolianeth, line 20: **suggestion** could we add a state here that prompts the user to connect their wallet if `userAddress` is falsey (haven't connected a wallet)? WDYT?

aeolianeth (changes requested):

aeolianeth:
@jmilldotdev amazing!!! Thanks for picking this up. Much love.

Left 2 suggestions but otherwise this is great 👏 👏

src/components/Projects/MyProjects.tsx
```diff
@@ -0,0 +1,68 @@
...
+    <React.Fragment>
```
aeolianeth, line 21: **suggestion**: wondering if we should also add a small note above the projects, something like "Juicebox projects you own". Not sure though

jmilldotdev: I modeled after the 'Holdings' tab, there is an info paragraph below that says 'Projects that you have created'.

We could put it above for clarity's sake in both tabs?

src/components/Projects/MyProjects.tsx
jmilldotdev: The one tricky thing is that v1 doesn't have a 'create project' page and this is a shared component. It's easy enough for v2 but for v1 we'd be linking to the main page then scrolling to the create section?

src/components/Projects/MyProjects.tsx
jmilldotdev: Yeah I like it, and we could do the same for the holdings tab.

src/components/Projects/MyProjects.tsx
aeolianeth: Few things:
- we'll still have a `/v1/create` route for v1. V2 will be at `/create`
- No Create UI on the homepage anymore
- Everything is v2-by-default going forward.

I think we can leave this for now though - save ourselves some hassle of dealing with a feature flag here, and add a button once V2 is out 🙏

src/components/Projects/MyProjects.tsx
```diff
@@ -0,0 +1,85 @@
...
+              </div>
+            ))}
+
+          {isLoading && (
+            <div style={{ marginTop: 40 }}>
+              <Loading />
+            </div>
+          )}
+
+          <p style={{ marginBottom: 40, marginTop: 40, maxWidth: 800 }}>
```
aeolianeth, line 77: **suggestion** could we only show this when there are some projects? Other than that, looks great 👏

src/components/Projects/MyProjects.tsx
aeolianeth: @jmilldotdev nice nice, yep thats great, i think keep it as you have it.

Just left one comment about hiding it on the empty state: https://github.com/jbx-protocol/juice-interface/pull/937/files#r867260015

src/components/Projects/MyProjects.tsx
jmilldotdev: By 'this', do you mean the tab itself? The weird thing is that fetching projects/holdings is an asynchronous thing, so we would either have to display the tabs by default and remove them if none are found, or add them to the UI after the query returns what it has found.

aeolianeth: @jmilldotdev Oh no! Sorry, should have been more clear.

I meant this "Projects that you have created." message. Maybe we can only show this message when there are > 0 projects. WDYT?

jmilldotdev: Done, also did the same for the holdings tab

aeolianeth (approved):
@jmilldotdev, massive stuff my guy! Thanks so much again for this contribution. Totally LGTM, i had a few suggestions that I applied in this commit in the interest of efficiency: https://github.com/jbx-protocol/juice-interface/pull/937/commits/e259543d87b1f8d8aff8e97d169f6c16d4c5c9d3.

Approving and merging 🚀

PR #952, opened 2022-05-07
Route updates for V2

aeolianeth:

- make `/create` be V2 create when V2 feature flag is enabled
- if  V2 feature flag enabled, add a `/v1/create` route for v1
- if V2 feature flag disabled, add a `/v2/create` route for v2


johnnyd-eth (approved):
LGTM

PR #967, opened 2022-05-08
Refactor ETHAddress form item

aeolianeth:

_Provide a description of what this PR does. Link to any relevant GitHub issues,
Notion tasks or Discord discussions._

changes._

## Acceptance checklist


src/components/shared/formItems/EthAddress.tsx
```diff
@@ -24,65 +27,49 @@ export default function EthAddress({
     theme: { colors },
   } = useContext(ThemeContext)
 
-  const [addressForENS, setAddressForENS] = useState<string>()
-  const [displayValue, setDisplayValue] = useState<string>()
+  const [value, setValue] = useState<string>()
+  const [addressForENSName, setAddressForENSName] = useState<string>()
+  const [ENSName, setENSName] = useState<string>()
```
wraeth-eth, line 32: think naming should always be consistent, so always using camel for vars - bigger fan of `ensName` rather than `ENSName`, thoughts?

src/components/shared/formItems/EthAddress.tsx
```diff
@@ -24,65 +27,49 @@ export default function EthAddress({
...
 
   const onInputChange = useCallback(
+    // value can be ENS name *or* ETH address (0x00...)
     async (value: string) => {
-      setDisplayValue(value)
+      setValue(value)
+      onAddressChange(value)
+
+      setAddressForENSName('')
```
wraeth-eth, line 40: should these be set to `undefined`

src/components/shared/formItems/EthAddress.tsx
```diff
@@ -95,15 +82,17 @@ export default function EthAddress({
       }
     >
       <Input
```
wraeth-eth, line 84: suggestion: disable spell check

wraeth-eth (changes requested):

src/components/shared/formItems/EthAddress.tsx
aeolianeth: Yes totally! good call

src/components/shared/formItems/EthAddress.tsx
aeolianeth: probably!

src/components/shared/formItems/EthAddress.tsx
aeolianeth: goooooood

aeolianeth:
Thanks @wraeth-eth for the review!

@johnnyd-eth over to you for review/merge.

Note that there is still some fuckiness here - the input (and forms genreally) have re-rendering issues, so sometimes the loading state will be triggered when it doesn't make sense. This is also a thing currently in the app, so it's not a step backwards.

johnnyd-eth:
I'm not sure if this is not a step backwards - previously, although getting hit by that race condition where the field would clear after typing a few characters, we could always get the ENS working eventually. With this, I'm often getting completely stuck: I'll type in ENS, see the spinning wheel come up, see it go away, then I click submit and it says 'address is required'. Do you not think it'll be possible to get this tight before Tuesday?

Also, will it be possible to keep the ENS name in the field rather than have the whole address replace it when the ENS resolves?

aeolianeth:
@johnnyd-eth that might the be the expected behavior on Rinkeby. A lot of times, ENS names don't have their reverse records set, so it won't work. Try it with chasemcdude.eth, or try it on `mainnet` and see how you go.

That second point is because of the re-rendering issues I described, which I think is a pre-existing problem.

johnnyd-eth:
Ahh yeah ok, I see it working much better on Mainnet. This is definitely an improvement for now in this case so let's merge for sure. But would love to get this as tight as possible when we can

Issue #1001, opened 2022-05-13
Limit pay button characters and UI width


aeolianeth:
TODO:
1. Add character limit to "pay" button text (anything that makes sense)
2. Fix button width on Project page

### screenshots

juicebox.money/#/v2/p/4

<img width="1138" alt="Screen Shot 2022-05-13 at 3 24 46 pm" src="https://user-images.githubusercontent.com/94939382/168220177-9aaa92e0-7c98-4637-a34c-e31688b272a3.png">

PR #1003, opened 2022-05-13
Limit pay button

wraeth-eth:

Closes https://github.com/jbx-protocol/juice-interface/issues/1001

Also updated antd to get new functionality for inputs.

https://user-images.githubusercontent.com/104132113/168269649-3453c5d7-9fa2-4ae3-824a-4d0f1b0871d0.mov

## Acceptance checklist


package.json
```diff
@@ -22,7 +22,7 @@
     "@uniswap/sdk-core": "3.0.1",
     "@uniswap/v3-sdk": "3.6.3",
     "@walletconnect/web3-provider": "^1.5.4",
-    "antd": "4.16.13",
+    "antd": "4.18.0",
```
johnnyd-eth, line 25: **Question**

johnnyd-eth (approved):
Clean AF. So good. Just wondering if the Antd update should be in this PR?

package.json
wraeth-eth: Thanks for the review! The antd update brings in the change to the input form to use showCount. It was added in 4.7.x.

src/constants/numbers.ts
```diff
@@ -2,3 +2,5 @@ export const WAD_DECIMALS = 18
 
 export const SECONDS_IN_DAY = 24 * 60 * 60
 export const SECONDS_IN_HOUR = 60 * 60
+
+export const PROJECT_PAY_CHARACTER_LIMIT = 16
```
aeolianeth, line 6: @wraeth-eth given https://github.com/jbx-protocol/juice-interface/pull/1007, would it make sense to increase this slightly? Not sure, your call 🙏

wraeth-eth: Oh for sure - if we have plans to wrap it, lets increase maybe to 32? WDYT?

wraeth-eth: Could even go bigger tbh:
image

wraeth-eth: image

aeolianeth: On second thought I think I'm in favour of keeping it at 16. Not sure we want to encourage long buttons, generally not smart from UX pov.

In other words, lgtm!

wraeth-eth: 🙌

aeolianeth (approved):
@wraeth-eth approving and merging!

@johnnyd-eth lets keep an eye out for any ant-d weirdness that might have been introduce from the upgrade.

PR #1010, opened 2022-05-17
Add google site verification via `meta` tag

aeolianeth:
This is to get us access to Google Search Console.

Usually we'd do this using a DNS TXT record, but there's some problem with the way our records are configured currently. This is the issue https://support.google.com/webmasters/thread/73081980/ownership-verification-failed-txt-records-appear-in-dns?hl=en.

I would try to fix it but I don't want to risk taking the site down for an extended period. I think it's fine to add this meta tag for now.


aeolianeth:
cc @johnnyd-eth @peripheralist, I'm going to merge this w/o review, just FYI

PR #1042, opened 2022-05-20
Extend the project description to 3 lines if character limit not exceeded

johnnyd-eth:

Before: 

Screen Shot 2022-05-20 at 12 53 10 pm

After: 

Screen Shot 2022-05-20 at 1 18 33 pm

changes._

## Acceptance checklist


src/components/shared/Paragraph.tsx
```diff
@@ -13,15 +13,26 @@ export default function Paragraph({
 }) {
   const CHARACTER_LIMIT_EXCEEDED =
     (characterLimit && description.length > characterLimit) ||
-    description.split('\n').length > 1
+    description.split('\n').length > 3
```
wraeth-eth, line 16: One thing to possibly consider is if they have one super long line of text. I wonder if we should expand this to check each line for a character limit, wdyt?

Something like:
```ts
const isCharacterLineExceeded(description: string, characterLimit: number | undefined ) {
  if (!characterLimit) return false
  const lines = description.split('\n')
  if (lines > 3) return false
  return lines.every(line => line.length < characterLimit)
}
...
const CHARACTER_LIMIT_EXCEEDED = isCharacterLineExceeded(description, characterLimit)
```

src/components/shared/Paragraph.tsx
```diff
@@ -13,15 +13,26 @@ export default function Paragraph({
...
 
   const [expanded, setExpanded] = useState<boolean>(false)
   const toggleExpanded = () => setExpanded(!expanded)
 
-  // truncate the first line of the description
+  const allLines = description.split('\n')
+  let firstThreeLines = ''
+
+  // Truncate the first 3 lines of the description
+  allLines.forEach((line, index) => {
```
wraeth-eth, line 25: you probably only need to loop over the first three lines with slice(0, 3), and then could join them with a new line:
```ts
const firstThreeLines = allLines
  .slice(0, 3)
  .join('\n')
```

wdyt?

wraeth-eth: You could even possibly put all of this into the memo itself so it only runs when the character limit, or the description changes

wraeth-eth (changes requested):
Looking good - good features for sure! A bit tired at the moment so hopefully my suggestions make sense, lmk what you think

src/components/shared/Paragraph.tsx
johnnyd-eth: @wraeth-eth Not sure I follow this. If one line is super long then it'll just get cut off by the characterLimit right?

src/components/shared/Paragraph.tsx
johnnyd-eth: > you probably only need to loop over the first three lines with slice(0, 3), and then could join them with a new line:
> 
> ```ts
> const firstThreeLines = allLines
>   .slice(0, 3)
>   .join('\n')
> ```
> 
> wdyt?

Beautiful ser

johnnyd-eth: > ```ts
> const firstThreeLines = allLines
>   .slice(0, 3)
>   .join('\n')
> ```

Yep another great call

src/components/shared/Paragraph.tsx
wraeth-eth: Yeah I think you're right - resolving

wraeth-eth (approved):

PR #1135, opened 2022-06-03
Next.js prep: upgrade packages, TS/eslint rules

aeolianeth:

- Upgrades a few packages, notably eslint and frens
- Updates codebase to comply with new eslint rules
- a few other minor QoL improvements in package.json

App still works.

## Acceptance checklist

- [ ] I have evaluated the
      for this PR.


.eslintrc.json
```diff
@@ -1,8 +1,39 @@
...
+    "ecmaFeatures": {
+      "jsx": true
+    },
+    "ecmaVersion": "latest",
+    "sourceType": "module"
+  },
+  "plugins": ["react", "@typescript-eslint", "import", "react-hooks"],
   "rules": {
+    "@typescript-eslint/no-non-null-assertion": "off",
+    "@typescript-eslint/no-var-requires": "off",
```
aeolianeth, line 28: keep for now to avoid messy refactor of some v1 contract loading

.eslintrc.json
```diff
@@ -1,8 +1,39 @@
...
+  "parserOptions": {
+    "ecmaFeatures": {
+      "jsx": true
+    },
+    "ecmaVersion": "latest",
+    "sourceType": "module"
+  },
+  "plugins": ["react", "@typescript-eslint", "import", "react-hooks"],
   "rules": {
+    "@typescript-eslint/no-non-null-assertion": "off",
```
aeolianeth, line 27: keep for now

.eslintrc.json
```diff
@@ -1,8 +1,39 @@
...
+    "@typescript-eslint/no-var-requires": "off",
+    "@typescript-eslint/ban-types": "off",
+    "@typescript-eslint/explicit-module-boundary-types": "off",
+    "react/prop-types": "off",
+    "@typescript-eslint/ban-ts-comment": "off",
+    "react/no-unescaped-entities": "off",
+    "@typescript-eslint/no-inferrable-types": "off",
+    "react/react-in-jsx-scope": "off",
```
aeolianeth, line 35: This breaks the whole app, and react is guaranteed to be in scope in our case. There's probably a way around turning it off, but this feels fine for now.

package.json
```diff
@@ -15,42 +15,46 @@
     "@sentry/react": "^6.18.2",
     "@sentry/tracing": "^6.18.2",
     "@sushiswap/sdk": "5.0.0-canary.116",
-    "@testing-library/jest-dom": "^5.11.4",
-    "@testing-library/react": "^11.1.0",
-    "@testing-library/user-event": "^12.1.10",
+    "@types/ethereum-block-by-date": "^1.4.1",
```
aeolianeth, line 18: `dependencies` now contains everything required to build the app - and nothing more. This should speed up the Fleek build time a bit.

scripts/build.sh
```diff
@@ -0,0 +1 @@
+NODE_ENV=production yarn install && yarn build
```
aeolianeth, line 1: once this PR merges, we can update fleek mainnet and rinkeby to use this script for building. Means we have a single source of truth.

src/components/Landing/QAs.tsx
```diff
@@ -1,15 +1,16 @@
+/* eslint-disable react/jsx-key */
```
aeolianeth, line 1: disabled due to laziness

aeolianeth:
Hey @wraeth-eth as you've already reviewed @benschac's PR, would you mind reviewing this one when you get a chance? Cheers.

.eslintrc.json
wraeth-eth: sounds good - we shoudl make issues for these if we haven't already

package.json
wraeth-eth: LFG!!!!

wraeth-eth: same with eslint and what not

package.json
```diff
@@ -60,7 +64,7 @@
     "prebuild": "yarn compile-styles && yarn i18n:compile",
     "test": "react-scripts test",
     "eject": "react-scripts eject",
-    "lint": "eslint --ext .js,.jsx,.ts,.tsx ./",
+    "lint": "eslint --ext .ts,.tsx ./",
```
wraeth-eth, line 67: i would keep .js and .jsx just in case

src/components/Landing/QAs.tsx
wraeth-eth: would it be better to make this a warn rule, rather than disabling herfe so we are reminded to fix it? (I personally hate warnings so would motivate me to fix 😅 )

src/components/Landing/QAs.tsx
```diff
@@ -1,15 +1,16 @@
...
 import { t, Trans } from '@lingui/macro'
 import ExternalLink from 'components/shared/ExternalLink'
-import { PropsWithChildren } from 'react'
+import { ReactNode } from 'react'
 import { Link } from 'react-router-dom'
 
-export const OverflowVideoLink = ({ children }: PropsWithChildren<{}>) => (
+export const OverflowVideoLink = ({ children }: { children: ReactNode }) => (
```
wraeth-eth, line 7: what do we get with this approach vs `PropsWithChildren`, or is just that PropsWithChild<{}> is antipattern?

wraeth-eth: more I think about it, more I don't like PropsWithChildren<{}>

src/components/Projects/HoldingsProjects.tsx
```diff
@@ -34,8 +34,8 @@ export default function HoldingsProjects() {
           {projects && projects.length > 0 && (
             <React.Fragment>
               <Grid>
-                {projects.map(p => (
-                  <ProjectCard project={p} />
+                {projects.map((p, i) => (
+                  <ProjectCard key={i} project={p} />
```
wraeth-eth, line 38: I am pretty sure from memory that this is a antipattern for react and should use a unique key id, rather than index.

src/components/Projects/MyProjects.tsx
```diff
@@ -44,8 +44,8 @@ export default function MyProjects() {
     <>
       {projects && projects.length > 0 && (
         <Grid>
-          {projects.map(p => (
-            <ProjectCard project={p} />
+          {projects.map((p, i) => (
+            <ProjectCard key={i} project={p} />
```
wraeth-eth, line 48: same applies here

src/components/v2/shared/FundingCycleConfigurationDrawers/__tests__/FundingCycleDrawer.test.tsx
```diff
@@ -4,7 +4,7 @@ import { useFundingCycleDrawer } from '../useFundingCycleDrawer'
 
 describe('FundingCycleDrawer', () => {
   describe('useFundingCycleDrawer', () => {
-    const instanceUnderTest = (onClose: VoidFunction = () => {}) => {
+    const instanceUnderTest = (onClose: VoidFunction = jest.fn()) => {
```
wraeth-eth, line 7: maybe even just
```suggestion
    const instanceUnderTest = (onClose = jest.fn()) => {
```

wraeth-eth (changes requested):
LOVE THIS!!! v excited to see it go in. some comments, lmkwyt

package.json
aeolianeth: @wraeth-eth  yesss I had the same initial thought, but because the build script does typescript compilation and also eslint checks, we need em all in `dependencies`.

So pretty much everything except **dev-only** deps (like source map analyzer and whatnot) is a `dependency` of the app atm.

src/components/Landing/QAs.tsx
aeolianeth: Yess!!! Great catch. 

`PropsWithChildren<{}>` is now an eslint error so that's why it's removed, but you're totally right, we can do React.FC 🚀

src/components/Projects/HoldingsProjects.tsx
aeolianeth: yep totally is, was carryover from the old PR 🙏 . Thanks for catching!!

package.json
aeolianeth: Yeah probably a good call hey :pray:

src/components/Landing/QAs.tsx
aeolianeth: fuck it, fixed it lol

src/components/Landing/QAs.tsx
aeolianeth: done!

.eslintrc.json
```diff
@@ -1,8 +1,38 @@
...
+    "ecmaVersion": "latest",
+    "sourceType": "module"
+  },
+  "plugins": ["react", "@typescript-eslint", "import", "react-hooks"],
   "rules": {
+    "@typescript-eslint/no-non-null-assertion": "off",
+    "react/prop-types": "off",
+    "react/no-unescaped-entities": "off",
+    "react/react-in-jsx-scope": "off",
+    "@typescript-eslint/explicit-module-boundary-types": "off",
```
aeolianeth, line 31: Follow-up issue: https://github.com/jbx-protocol/juice-interface/issues/1138

aeolianeth, line 31: Follow-up issue https://github.com/jbx-protocol/juice-interface/issues/1139

.eslintrc.json
```diff
@@ -1,8 +1,38 @@
...
+
+    "@typescript-eslint/no-var-requires": "warn",
```
aeolianeth, line 33: Follow-up issue: https://github.com/jbx-protocol/juice-interface/issues/1140

.eslintrc.json
```diff
@@ -1,8 +1,38 @@
...
+    "@typescript-eslint/ban-ts-comment": "warn",
```
aeolianeth, line 34: Follow-up issue: https://github.com/jbx-protocol/juice-interface/issues/1141

.eslintrc.json
aeolianeth: Totttalllyy. Great call. done for all relevant ones.

Follow-up for this one Follow-up issue https://github.com/jbx-protocol/juice-interface/issues/1139

aeolianeth:
@wraeth-eth ty for the awesome review! Applied all your suggestions, lemme know 🏓

package.json
wraeth-eth: ok cool cool TIL sounds good

src/components/Landing/QAs.tsx
wraeth-eth: YEAH

src/components/Landing/Faq.tsx
```diff
@@ -7,7 +7,7 @@ export default function Faq() {
   return (
     <Collapse defaultActiveKey={QAs.length ? 0 : undefined} accordion>
       {QAs().map((qa, i) => (
-        <CollapsePanel header={qa.q} key={i}>
+        <CollapsePanel header={qa.q} key={`qa_panel_${i}`}>
```
wraeth-eth, line 10: I think this is a step in the right direction, but will it be strong enough? I don't have the full breakdown to how much data is in each list and how performant each should be but this is a good writeup about it:
https://www.developerway.com/posts/react-key-attribute

I'm a bit hungover so i'm only taking in about 40% of it, but will re-read tomorrow when I am healthy again.

TLDR is I think adding `qa_panel_` to the front is the same in terms of performance as just having `i` as it will treat it as the same prop (if that makes sense). The benefit of using a unique id for each record (UUID, or some other unique id assoc with record) is that irrespective of whether the indexing changes, it will not destroy and recreate elements, but instead just smartly move them around (I think)

src/components/Landing/QAs.tsx
```diff
@@ -18,19 +19,19 @@ export const JBDiscordLink = ({ children }: PropsWithChildren<{}>) => (
...
-// Take-away: If you're storing a list of <Trans>` strings in a list,
+// Take-away: If you're storing a list of <Trans key={uniqueId()}>` strings in a list,
 // make sure you're not rendering it from the same file.
 export default function QAs() {
   return [
     {
       q: <Trans>Who funds Juicebox projects?</Trans>,
       a: [
-        <Trans>
+        <Trans key={uniqueId()}>
```
wraeth-eth, line 29: This I think will still suffer the same problem. Have a look in the article I linked, especially surrounding the example the author gives of `key={Math.random()}`

Definitely think you're on the right track though

wraeth-eth: tbh though - I am wondering if it might be okay for Trans.... 🤔

wraeth-eth: Seems kind of useless having the key for trans as is. Wonder if the docs have anything about bypassing the need for Trans key

wraeth-eth (changes requested):

src/components/Landing/QAs.tsx
aeolianeth: Yeah this a great point, thanks for calling it out.

I ended up refactoring this whole QA thing to avoid this problem entirely.

src/components/Landing/Faq.tsx
aeolianeth: Totally totally, thanks for calling it out!

That was a great article - I didn't know that `key` values don't need to be globally unique; just unique for each specific tree (loop).

In this case, I think using the index is fine. `QAs()` is a static list; the order of items, or the items themselves, won't ever change, so I don't think we need to worry about unique ids for keys.

I've gone over every `key` prop in the app and checked if it's a dynamic (meaning, the elements might change) or static list; in all possible cases I've tried to use unique keys based on the data itself, rather than indexes. But for static lists, I've left those as indexes.

Let me know if I'm misunderstanding something though (or if I've missed anything in the code). Def want to get this right 🙏 .

wraeth-eth: Nah I think that’s it! I’ll need to double check the article I linked but it sounds like the same understanding I had too

src/components/Landing/index.tsx
```diff
@@ -111,7 +111,7 @@ export default function Landing() {
       ].map((data, i) => (
         <Space
           style={{ fontStyle: 'italic', paddingLeft: 8 }}
-          key={i}
+          key={`builtfor_${i}`}
```
wraeth-eth, line 114: I think you're better off just using `i` here. I don't think the extra text will make a diff to performance

src/components/Landing/Faq.tsx
wraeth-eth: Same with this text

src/components/Projects/HoldingsProjects.tsx
```diff
@@ -17,70 +17,67 @@ export default function HoldingsProjects() {
...
-            <React.Fragment>
-              <Grid>
-                {projects.map(p => (
-                  <ProjectCard project={p} />
-                ))}
-              </Grid>
-            </React.Fragment>
-          )}
+    <>
+      {projects && projects.length > 0 && (
```
wraeth-eth, line 36: I'm a big fan of `array.length` sans >0, how do you guys usually operate? The performance diff are severely marginal, but I know when I read `array.length` it means array has something. When parsing `projects.length > 0` I have to do one extra cognitive step, if that makes sense. I don't really mind either, but it's good for consistency (my vote is on sans > 0 though ;) )

src/components/Projects/HoldingsProjects.tsx
```diff
@@ -17,70 +17,67 @@ export default function HoldingsProjects() {
...
+        <Grid>
+          {projects.map(p => (
+            <ProjectCard key={`${p.id}_${p.cv}`} project={p} />
```
wraeth-eth, line 39: **praise**: phat

src/components/Projects/HoldingsProjects.tsx
```diff
@@ -17,70 +17,67 @@ export default function HoldingsProjects() {
...
+          </div>
+        ) : (
+          <div
+            style={{
+              textAlign: 'center',
+              color: colors.text.disabled,
+              padding: 20,
+            }}
+          >
+            {projects.length} {projects.length === 1 ? t`project` : t`projects`}{' '}
```
wraeth-eth, line 65: can there ever be more than 1 project? If so, this should be `projects.length > 1 ?...`/`projects.length ?...`

src/components/shared/ProjectRiskNotice.tsx
```diff
@@ -38,8 +38,8 @@ export default function ProjectRiskNotice({
         </Trans>
       </p>
       <ul>
-        {warnings.map(text => (
-          <li>{text}</li>
+        {warnings.map((text, i) => (
+          <li key={`warning_${i}`}>{text}</li>
```
wraeth-eth, line 42: wont comment on more of these but unsure if extra text buys us anything

wraeth-eth (changes requested):

src/components/Projects/HoldingsProjects.tsx
aeolianeth: I don't feel strongly about this, although there is a slight distinction between them in React-land: https://reactjs.org/docs/jsx-in-depth.html?#booleans-null-and-undefined-are-ignored. 

A quick grep of the codebase reveals we use both approaches fairly equally. I'd generally agree with you but in JSX, I think it's safer to opt for the strict boolean eval offered by `> 0`.

In any case, I'll leave this as-is for now.

src/components/shared/ProjectRiskNotice.tsx
aeolianeth: Yeah they def don't, thanks for calling out. My original understanding was that keys should be "globally" unique, but that's not correct 🙌 . Changed!

wraeth-eth (approved):

src/components/Projects/HoldingsProjects.tsx
aeolianeth: Not quite sure what you mean here. Here's the cases I see:

- project.length == 0 => `0 projects`
- project.length == 1 => `1 project`
- project.length == 2 => `2 projects` (and so on)

In any case, this code wasn't changed for this PR so we can handle any issues in a follow-up 🙌

src/components/Landing/Faq.tsx
aeolianeth: done! https://github.com/jbx-protocol/juice-interface/pull/1135#discussion_r889677226

Issue #911, opened 2022-05-02
Make button "connect to wallet" for any button that initiates a transaction.


aeolianeth:
If no wallet is connected, any button that initiates a transaction doesn't do anything when clicked. This is a confusing UX.

### proposal

Make every button that initiates a transaction open the "connect wallet" modal if no wallet is connected.

### possible implementation

Might be able to build a single TransactorButton component that handles this logic internally.

aeolianeth:
Need to use this on Distribute modal 

<img width="682" alt="Screen Shot 2022-05-16 at 9 04 07 am" src="https://user-images.githubusercontent.com/94939382/168498953-3dac7dcd-a19a-4e86-8b8e-269bf971fff2.png">

PR #1146, opened 2022-06-05
Improve UX paying project with wallet disconnected

### LINKED ISSUE #1014: Improve pay process when wallet disconnected
## Description

From UX standpoint would be nice to disable the pay button and show a tooltip asking user to connect wallet on project page rather than throwing a mysterious transaction reverted error when they attempt to pay.

## Proposal

in general it'd be nice if we had a component like WalletConnectedButton which is basically an antd button with a disabled guard for wallet connected = True, and a tooltip that asks user to connect wallet if it's not connected.

## Screenshots

_If applicable, add screenshots to help explain your problem._

johnnyd-eth:

Slight improvement made in consideration to Jmill's issue here: (closes #1014).

Ideally we'd enable users to connect wallets from the modal, then keep it open when wallet is connected. However, `onSelectWallet` always causes a page reload, which currently causes the modal to close (unexpectedly from user's POV). 

Without overhauling `onSelectWallet()` and doing some fancy shit to prevent reload there, I think it's better UI to at least make users have to back out of the modal and connect the wallet manually. LMK what y'all think though

Before:

Screen Shot 2022-06-05 at 6 14 08 pm

After:

Screen Shot 2022-06-05 at 5 54 49 pm

## Acceptance checklist

- [ ] I have evaluated the
      for this PR.

### DIFF
--- a/src/components/v2/V2Project/V2PayButton/V2ConfirmPayModal/index.tsx
+++ b/src/components/v2/V2Project/V2PayButton/V2ConfirmPayModal/index.tsx
@@ -147,7 +147,8 @@ export default function V2ConfirmPayModal({
       title={t`Pay ${projectMetadata.name}`}
       visible={visible}
       onOk={pay}
-      okText={t`Pay`}
+      okText={userAddress ? t`Pay` : t`Connect wallet to pay`}
+      okButtonProps={{ disabled: !userAddress }}
       onCancel={onCancel}
       confirmLoading={loading}
       width={640}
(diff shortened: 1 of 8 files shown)


jmilldotdev (approved):

aeolianeth (approved):

aeolianeth (approved):
@johnnyd-eth approving this but I would like us to get a more wholistic solution in place.

I've scheduled this issue for this FC https://github.com/jbx-protocol/juice-interface/issues/911

PR #1148, opened 2022-06-05
Improve project search UI

johnnyd-eth:

- Moves the V2 project notice below the search results
- Hides it when projects still loading
- Improves project count message padding-top when 0 projects

Before:

https://user-images.githubusercontent.com/96150256/172064533-f12b14a4-62ab-4cac-ab9f-3289b28b01a2.mp4

After (excuse my terrible internet):

https://user-images.githubusercontent.com/96150256/172064549-6f27dde3-bfae-4b4b-bc13-433821eb0c5c.mp4

## Acceptance checklist

- [ ] I have evaluated the
      for this PR.

### DIFF
--- a/src/components/Projects/index.tsx
+++ b/src/components/Projects/index.tsx
@@ -210,20 +210,6 @@ export default function Projects() {
         />
       </div>
 
-      {!!searchText && (
-        <div
-          style={{
-            marginBottom: 20,
-            textAlign: 'center',
-            color: colors.text.secondary,
-          }}
-        >
-          <Trans>
-            <InfoCircleOutlined /> Search results don't include V2 projects yet.
-          </Trans>
-        </div>
-      )}
-
       {selectedTab === 'all' ? (
         <React.Fragment>
           {concatenatedPages && (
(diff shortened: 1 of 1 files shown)


src/components/Projects/index.tsx
```diff
@@ -287,6 +274,18 @@ export default function Projects() {
           <TrendingProjects count={12} trendingWindowDays={7} />
         </div>
       ) : null}
+      {Boolean(searchText) && !isLoading && (
```
wraeth-eth, line 277: **question** is there a particular preference for this in the project? I am a bigger fan of the double bang (I think the airbnb standard also prefers it). I have seen both used in project

wraeth-eth (approved):
LGTM @johnnyd-eth ! Nice job, makes for a nicer UI. Only thing i'd raise is maybe having the v2 warning be shown while loading, but for most users the loading time is very short and won't impact them

src/components/Projects/index.tsx
aeolianeth: Heh, I'm the opposite - Boolean() clear

aeolianeth: Ultimately: it's whatever the linter allows. Lets lint for this if we feel strongly

PR #1170, opened 2022-06-08
Fix 404 initial load

aeolianeth:
Fixes an issue where the 404 would "flash in" on the initial project load (closes https://github.com/jbx-protocol/juice-interface/issues/1167)

Fixes it by defaulting v2 contract loader "loading" state to `true`.

Also adds a skeleton loader for the Reserved Tokens number.


aeolianeth:
@johnnyd-eth  🏓

johnnyd-eth:
Beautiful fix here ser. LGTM

PR #1207, opened 2022-06-11
Improve wallet disconnect transaction modal state

### LINKED ISSUE #911: Make button "connect to wallet" for any button that initiates a transaction.

johnnyd-eth:

Closes #911 

To me, having the OK button be enabled and the okText as 'Connect wallet to ...' and then opening the wallet select would be ideal if not for the fact that this causes a page reload and therefore closes the modal on the user unexpectedly once they connect their wallet. (reload occurs because setting wallet writes to local storage). 

So then, I think the next best option is just disabling the button with the 'Connect wallet to ...' text, and make the user back out to connect their wallet. 100% open to feedback or other ideas here though.

For every transaction modal that users can get to without a wallet connected, I've implemented this logic. 

https://user-images.githubusercontent.com/96150256/173198766-10551d5a-309f-40a0-9c97-9868c50e81de.mp4

## Acceptance checklist

- [ ] I have evaluated the
      for this PR.

### DIFF
--- a/src/components/shared/TransactionModal.tsx
+++ b/src/components/shared/TransactionModal.tsx
@@ -1,5 +1,6 @@
 import { t, Trans } from '@lingui/macro'
 import { Modal, ModalProps } from 'antd'
+import { NetworkContext } from 'contexts/networkContext'
 
 import { ThemeContext } from 'contexts/themeContext'
 import { PropsWithChildren, useContext } from 'react'
@@ -7,6 +8,7 @@ import { PropsWithChildren, useContext } from 'react'
 type TransactionModalProps = PropsWithChildren<
   ModalProps & {
     transactionPending?: boolean
+    connectWalletText?: string
   }
 >
 
@@ -51,11 +53,20 @@ const PendingTransactionModalBody = () => {
  * are replaced with a juicy loading state.
  */
 export default function TransactionModal(props: TransactionModalProps) {
+  const { userAddress } = useContext(NetworkContext)
+  const okText = userAddress
+    ? props.okText
+    : props.connectWalletText ?? t`Connect wallet`
   const modalProps = {
     ...props,
     ...{
       confirmLoading: props.transactionPending || props.confirmLoading,
       cancelText: t`Close`,
+      okButtonProps: {
+        ...props.okButtonProps,
+        disabled: !userAddress,
+      },
+      okText: okText,
     },
   }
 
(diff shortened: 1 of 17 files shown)


src/components/shared/TransactionModal.tsx
```diff
@@ -1,12 +1,14 @@
...
 import { Modal, ModalProps } from 'antd'
+import { NetworkContext } from 'contexts/networkContext'
 
 import { ThemeContext } from 'contexts/themeContext'
 import { PropsWithChildren, useContext } from 'react'
 
 type TransactionModalProps = PropsWithChildren<
   ModalProps & {
     transactionPending?: boolean
+    connectWalletText?: string
```
aeolianeth, line 11: **praise** love the prop idea!

aeolianeth (changes requested):
@johnnyd-eth this probably works for me.

But a reload need not occur - there's a solution to prevent this. I agree it's out of scope of this PR.

There's a few other buttons that we should port this functionality to, which was my original intention with this issue:
- reconfigure modal
- edit project details drawer
- transfer unclaimed tokens
- add to balance

So that we don't duplicate logic, I suggested we have a `TransactorButton` component that we use anywhere we create a transaction. let me know what you think

johnnyd-eth:
> @johnnyd-eth this probably works for me.
> 
> But a reload need not occur - there's a solution to prevent this. I agree it's out of scope of this PR.
> 
> There's a few other buttons that we should port this functionality to, which was my original intention with this issue:
> 
> * reconfigure modal
> * edit project details drawer
> * transfer unclaimed tokens
> * add to balance
> 
> So that we don't duplicate logic, I suggested we have a `TransactorButton` component that we use anywhere we create a transaction. let me know what you think

@aeolianeth Yeah good call, I did miss the 'transfer' and 'add to balance' buttons - I created a `TransactorButton` component for them. I think the cleanest solution is still just what I've done with `TransactorModal` since `antd` modals can't take buttons as props.

As for the other instances you've mentioned, they can't be seen or accessed with the wallet disconnected so I didn't think they were worth doing anything for.

johnnyd-eth:
Screen Shot 2022-06-12 at 9 34 28 am

Also snuck in a fix to this layout

aeolianeth:
@johnnyd-eth ah yeah understood.

Merging! Thanks for working on this, really nice UX imrpovement.

aeolianeth (approved):

PR #1085, opened 2022-05-27
Upgrade JB contracts + project relaunch FC UI

aeolianeth:
## What does this PR do and why?

TBA

## Screenshots or screen recordings

<details>
<summary>Existing V2 project page</summary>

localhost_3000_ (1)

</details>

<details>
<summary>Relaunch FC modal</summary>

This is an example of what the project will see when they are relaunching.

LEFT: the original FC data that launched the project

RIGHT: The data that will be used to relaunch the project. This should be the exact same as the original FC.

| original FC config | relaunch FC config |
| --- | --- |
| Screen Shot 2022-05-27 at 4 08 51 PM | Screen Shot 2022-05-27 at 4 28 31 PM |

</details>

**Relaunch UI demo**

https://user-images.githubusercontent.com/94939382/170651025-a3ba9e25-a75d-4884-b958-1cf87f0ed64d.mp4

## Acceptance checklist

- [x] I have evaluated the [Approval Guidelines](https://github.com/jbx-protocol/juice-interface/blob/main/CONTRIBUTING.md#approval-guidelines) for this PR.
- [x] I have tested this PR in [all supported browsers](https://github.com/jbx-protocol/juice-interface/blob/main/CONTRIBUTING.md#supported-browsers).

PR #1213, opened 2022-06-12
Compiler fixes

wraeth-eth:

_Provide a description of what this PR does. Link to any relevant GitHub issues,
Notion tasks or Discord discussions._

changes._

## Acceptance checklist

- [ ] I have evaluated the
      for this PR.

### DIFF
--- a/src/hooks/v1/V1ContractLoader.ts
+++ b/src/hooks/v1/V1ContractLoader.ts
@@ -23,14 +23,19 @@ export function useV1ContractLoader() {
         // Contracts can be used read-only without a signer, but require a signer to create transactions.
         const signerOrProvider = signingProvider?.getSigner() ?? readProvider
 
-        const newContracts = Object.values(V1ContractName).reduce(
-          (accumulator, V1ContractName) => ({
-            ...accumulator,
-            [V1ContractName]: loadContract(
-              V1ContractName,
-              network,
-              signerOrProvider,
-            ),
+        const loadContractKeyPair = async (contractName: V1ContractName) => ({
+          key: contractName,
+          val: await loadContract(contractName, network, signerOrProvider),
+        })
+
+        const newContracts = (
+          await Promise.all(
+            Object.values(V1ContractName).map(loadContractKeyPair),
+          )
+        ).reduce(
+          (acc, { key, val }) => ({
+            ...acc,
+            [key]: val,
           }),
           {} as V1Contracts,
         )
--- a/src/hooks/v1/V1ContractLoader.ts
+++ b/src/hooks/v1/V1ContractLoader.ts
@@ -47,11 +47,13 @@ export function useV1ContractLoader() {
   return contracts
 }
 
-const loadContract = (
+const loadContract = async (
   contractName: keyof typeof V1ContractName,
   network: NetworkName,
   signerOrProvider: JsonRpcSigner | JsonRpcProvider,
-): Contract => {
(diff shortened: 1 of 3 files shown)


src/utils/v1/terminals.ts
```diff
@@ -10,6 +10,9 @@ const loadTerminalAddress = (
   network: NetworkName,
   terminal: V1TerminalName,
 ): string =>
+  // NOTE: This require is harder to easily change in the code base as it means
```
wraeth-eth, line 13: It's also worth nothing that this is v1 so we should try to spend as little time fixing, unless we have to

src/hooks/v1/V1ContractLoader.ts
```diff
@@ -47,11 +47,13 @@ export function useV1ContractLoader() {
   return contracts
 }
 
-const loadContract = (
+const loadContract = async (
```
aeolianeth, line 50: **issue**: we need to also refactor the `loadContracts` function https://github.com/jbx-protocol/juice-interface/pull/1213/files#diff-7d6dcf148ae7c1e6fa95ffb7b7368ab023c49294d6df19028ac0ebc08a11622fR29

Check the V2 contract loader for an example; i recently refactored it to be async.

src/utils/v1/terminals.ts
aeolianeth: yeah it would be great to fix this eventually, as it results in more efficient code-splitting: https://webpack.js.org/guides/code-splitting/

Should we create a follow-up issue for this?

wraeth-eth: I'm happy to fix it here if you think its valuable - will take a read of code-splitting

wraeth-eth: yep - i'll jiust do it now

wraeth-eth: Actuaslly it is a bigger headache than I originally thought. Since its v1 work, lets create an issue

wraeth-eth: https://github.com/jbx-protocol/juice-interface/issues/1214

src/hooks/v1/V1ContractLoader.ts
wraeth-eth: Are you talking about `src/hooks/v1/V1ContractLoader.ts`? It looks like it is already async:
```ts
export function useV1ContractLoader() {
  const [contracts, setContracts] = useState<V1Contracts>()

  const { signingProvider } = useContext(NetworkContext)

  useEffect(() => {
    async function loadContracts() {
      try {
...
```

aeolianeth: Yeah i think we should `await` until all the contracts are loaded, just to be safe.

Check this diff, should be copy-pastable: https://github.com/jbx-protocol/juice-interface/pull/1085/files#diff-bf47168917ef85b811be6c44b2cdecaba47629ea198f83ac7ab493b20b5f076bR42-R56

wraeth-eth: ah i see now - can do

wraeth-eth: have done 👍

src/utils/v1/terminals.ts
aeolianeth: yep nice

aeolianeth (approved):
@wraeth-eth approved!!! TY, real nice.

PR #1238, opened 2022-06-15
Refactor tools drawer

aeolianeth:

- break component up into multiple files
- add owner tab (tab hidden when owner not connected)
- clearer sections
- improve spacing and copy

⚠️ no functionality as changed :)

**Why?**

I want to add the V1V2 migration setup UI to this drawer, but adding it as-is would clog the drawer up too much. This separation is a better UX i believe.

FWIW i think we should have a dedicated settings page soon.

| wallet not connected | owner wallet connected |
| --- | --- |
| Screen Shot 2022-06-15 at 10 24 42 AM | Screen Shot 2022-06-15 at 10 28 31 AM |

## Acceptance checklist

- [x] I have evaluated the
      for this PR.
- [x] I have tested this PR in


src/components/shared/MinimalCollapse.tsx
```diff
@@ -1,12 +1,14 @@
-import { Collapse } from 'antd'
+import { Collapse, CollapseProps } from 'antd'
```
aeolianeth, line 1: totally unrelated change and unused atm.

jmilldotdev (approved):
LGTM, only thing i would say is that there are spacing differences between the original form and the new ones. The spacing is consistent now between the new form components, but i'm not sure which one is 'correct'. I think I prefer the less cramped spacing on the transfer form in the current JBX.money

image

aeolianeth:
@jmilldotdev lurrrrv it. agree with you. Made the change! Also moved some more files around,

wraeth-eth (commented):
Post mortem - nothing large and don't want to overreview as its already merged, but looks damn good! nice job

Just a couple of small comments, feel free to change them or not 👍

PR #1239, opened 2022-06-15
Fix v2 page titles

aeolianeth:
Fixes https://discord.com/channels/775859454780244028/942522503258054686/986443858852339742

Create a hook for setting page titles and use it in the router, v1 dashboard and v2 dashboard.

### DIFF
--- a/src/components/v1/V1Dashboard.tsx
+++ b/src/components/v1/V1Dashboard.tsx
@@ -23,17 +23,14 @@ import useTokenAddressOfProject from 'hooks/v1/contractReader/TokenAddressOfProj
 import useUriOfProject from 'hooks/v1/contractReader/UriOfProject'
 import { useCurrencyConverter } from 'hooks/CurrencyConverter'
 import { V1CurrencyOption } from 'models/v1/currencyOption'
-import { useEffect, useMemo } from 'react'
+import { useMemo } from 'react'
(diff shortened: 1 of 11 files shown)


aeolianeth:
@wraeth-eth could you review this when you get a chance? thanks

src/components/v1/V1Dashboard.tsx
```diff
@@ -96,13 +92,9 @@ export default function V1Dashboard() {
...
   const { data: metadata } = useProjectMetadata(uri)
 
-  useEffect(() => {
-    if (metadata?.name) {
-      document.title = `${metadata.name} | Juicebox`
-    } else {
-      document.title = DEFAULT_SITE_TITLE
-    }
-  }, [metadata])
+  usePageTitle({
```
wraeth-eth, line 95: nit:
I wonder if we should roll an auxilary hook like:
```ts
useJuiceboxPageTitle = (title?: string) => usePageTitle({title: title? `${title} | Juicebox` : undefined})
```
sorta psuedocode - but you get the idea?

wraeth-eth (approved):

src/components/v1/V1Dashboard.tsx
aeolianeth: Yeah good suggestion for sure, might leave for now though. Can def revisit.

PR #1251, opened 2022-06-17
Refactor Reconfigure to break down into smaller parts

wraeth-eth:

Refactors the reconfigure form to break into smaller testable hooks. Makes understanding the responsibilities of the modal a bit easier as well.

## Changes
The code itself is wholly unchanged (I think I changed a `useLayoutEffect` to `useEffect`).

Four new hooks have been added for handling state in the modal:

### useInitialEditingData

This hook is used to read the initial editing data from the project and write to the redux store. The `visible` it takes in is a bit hacky, but ensures that the useEffect to update the data is called when the visibility changes. This could be something a bit more generic to make the hook more reusable, but for now it is fine.

### useEditingProjectData

This hook is used to define the data being read from redux. Potentially could be reused everywhere that this data is used, but for now just exists here.

### useFundingHasSavedChanges

This hook manages the state of the overall changes and the drawer changes of the reconfigure. It returns 1 var and 3 functions used in reconfigure. ~~Realistically, we should update the 3 functions to just be memos... Might do that before I merge.~~ Done

### useReconfigureFundingCycle

The real meat and potatoes of why we're here. This is the actual transaction hook of the reconfigure. Takes in the editingProjectData and an exit call. We could likely rename this exit to something like `onConfirmed`.

## Acceptance checklist

- [ ] I have evaluated the
      for this PR.

### DIFF
--- a/src/components/v2/V2Project/V2ProjectReconfigureModal/hooks/editingProjectData.ts
+++ b/src/components/v2/V2Project/V2ProjectReconfigureModal/hooks/editingProjectData.ts
@@ -1,0 +1,43 @@
+import {
+  useAppSelector,
+  useEditingV2FundAccessConstraintsSelector,
+  useEditingV2FundingCycleDataSelector,
+  useEditingV2FundingCycleMetadataSelector,
+} from 'hooks/AppSelector'
+import {
+  V2FundingCycleMetadata,
+  V2FundingCycleData,
+  V2FundAccessConstraint,
+} from 'models/v2/fundingCycle'
+import {
+  ETHPayoutGroupedSplits,
+  ReservedTokensGroupedSplits,
+} from 'models/v2/splits'
+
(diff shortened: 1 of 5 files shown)


johnnyd-eth:
This looks great @wraeth-eth. I've tested and everything seems to be functioning perfectly as is. This diff has come out quite massive and it's hard to visualise exactly what you've put where. Would you mind adding a little summary in the description of this PR?

wraeth-eth:
Absolutely can do @johnnyd-eth - will update description. This is likely where I should have broken into individual commits, will try and keep that in mind next time

johnnyd-eth:
Very sick @wraeth-eth , thanks for the great description. I do love me some good meat and potatoes. Excellent work 🧠🙌

johnnyd-eth (approved):

Issue #1243, opened 2022-06-15
Can't submit form when sum of percentages is 100%


aeolianeth:
![Image](https://user-images.githubusercontent.com/94939382/173841546-5cee45da-82ad-41e6-8ef4-1cf13f4d362c.png)

https://discord.com/channels/775859454780244028/866040669712678942/986269157479837706

PR #1260, opened 2022-06-20
Fix rounding error in reserved token split

wraeth-eth:

Original bug: **Can't submit form when sum of percentages is 100%**

Identified as being caused by a rounding error where the number would appear as: `100.00000000000001`.

The fix instead makes the calculation before converting to a percentage (in its permyriad) and converts after the fact, avoiding the rounding error.

Closes https://github.com/jbx-protocol/juice-interface/issues/1243.

### At 99.99%

image

### At 100.00%

image

### At 100.01%

image

## Acceptance checklist

- [ ] I have evaluated the
      for this PR.


aeolianeth:
@wraeth-eth awesome! Thanks for a nice pr description too. lezzzz merge

aeolianeth (approved):

PR #949, opened 2022-05-07
Close modal on network change

aeolianeth:
## What does this PR do and why?

Subscribe to network change events. Means that the modal closes when user changes network.

Closes https://github.com/jbx-protocol/juice-interface/issues/407

This doesn't add an explicit "Switch Network" button to the modal, but this can be handled in https://github.com/jbx-protocol/juice-interface/issues/290

## Screenshots or screen recordings

https://user-images.githubusercontent.com/94939382/167236990-8cf38d1f-9d4a-4a6c-9b00-c01c9a511883.mp4

## Acceptance checklist

- [x] I have evaluated the [Approval Guidelines](https://github.com/jbx-protocol/juice-interface/blob/main/CONTRIBUTING.md#approval-guidelines) for this PR.
- [x] I have tested this PR in [all supported browsers](https://github.com/jbx-protocol/juice-interface/blob/main/CONTRIBUTING.md#supported-browsers).

Issue #407, opened 2022-01-20
Website isn't aware of network change events


LuckyKoala:
## Description

The website are not awared of network change event, this will bring inconsistency and inconvience to users:

1. Wallet connected to another network, user go to juicebox, this modal pop up
![image](https://user-images.githubusercontent.com/18607651/150303383-1002b3ee-17e2-488c-94c9-6e726f363201.png)

Currently, users who connected with different network will see `switchNetworkModal`. But this modal can't be closed, so users have to switch network through their wallet manually and then refresh the page.

---

2. Wallet connected to expected network, user go to juicebox, switch to other website, come back with different network connected
2.1 user go to dashboard page of arbitrary project
![image](https://user-images.githubusercontent.com/18607651/150310793-2be4fae4-7478-45c8-8b6b-bbcd4ae6e60f.png)
2.2 user click "Connect", both switchNetworkModal and network switch of Blocknative onboard pops up
![image](https://user-images.githubusercontent.com/18607651/150309541-c6af0814-859b-444d-90f2-60c04539a2c8.png)

## Proposal

1. monitor network change event, display and close `switchNetworkModal` accordingly to improve UX.
2. add connect option to this modal, since Blocknative has an onboard API which will pull up the wallet to suggest network change. (check screenshot below)

## Screenshots

![image](https://user-images.githubusercontent.com/18607651/150304790-eb7583a1-83f9-4a51-ae30-0e83c6a5f236.png)
![image](https://user-images.githubusercontent.com/18607651/150305844-3754e1f7-58db-43d9-b70a-5e9920a8a072.png)

peripheralist:
duplicate of #290, but leaving it open since @LuckyKoala has some great info here

jmilldotdev:
Honestly resolving this seems as easy as removing the switch network modal in `src/App.tsx`. Blocknative seems to handle requesting a network switch and denying the user's transactions with a warning modal that they need to switch out of the box.

Maybe as an additional guard, we could track the connected network and put a warning below the connected address box that the user is not on the correct network.

aeolianeth:
@jmilldotdev Is it possible that, when the user changes their wallet network, we can close the modal? That feels like the minimum viable change.

aeolianeth:
@jmilldotdev all good, found it https://github.com/jbx-protocol/juice-interface/pull/949

aeolianeth:
Reverted the PR: https://github.com/jbx-protocol/juice-interface/pull/962

[...]

PR #1261, opened 2022-06-20
dismiss wallet modal

wraeth-eth:

Fixes **1** in https://github.com/jbx-protocol/juice-interface/issues/407.

Problem:
image

Fixed by adding a null/undefined check to useHandleForProjectId -> formatter before passing to parseBytes32String.

This was initially fixed, but causing a runtime error here: 
https://github.com/jbx-protocol/juice-interface/pull/949
and reverted here:
https://github.com/jbx-protocol/juice-interface/pull/962/files

Doesn't necessarily close issue 407, but solves problem 1.

## Acceptance checklist

- [ ] I have evaluated the
      for this PR.


src/providers/NetworkProvider.tsx
```diff
@@ -63,6 +63,7 @@ export const NetworkProvider: React.FC = ({ children }) => {
     const config: Subscriptions = {
       address: setAccount,
       wallet: selectWallet,
+      network: networkId => setNetwork(NETWORKS[networkId]?.name),
```
aeolianeth, line 66: @wraeth-eth i like your fix here. Wondering if you think a hard page reload might be worthwhile instead of this though 🤔 . Just to be safe.

aeolianeth (commented):
@wraeth-eth niiiiceee, good catch. One question for ya

src/providers/NetworkProvider.tsx
wraeth-eth: definitely on board with a hard page refresh @aeolianeth - just smooths over any other bumps that might occur. Where is the best place to do that - in this function, or somewhere else?

wraeth-eth: Have updated - its not the best UX, but it works. I am uncertain if user will be swapping networks often while on JB too

aeolianeth (approved):
@wraeth-eth noice noice. Had a play, feels good to me!

PR #1291, opened 2022-06-28
IPFS API integration

aeolianeth:

Uses the new api for Pinata interactions

### DIFF
--- a/src/components/shared/forms/ProjectDetailsForm.tsx
+++ b/src/components/shared/forms/ProjectDetailsForm.tsx
@@ -3,7 +3,6 @@ import { t, Trans } from '@lingui/macro'
 
 import { FormItems } from 'components/shared/formItems'
 import { normalizeHandle } from 'utils/formatHandle'
-import { cidFromUrl, unpinIpfsFileByCid } from 'utils/ipfs'
 import { CSSProperties } from 'react'
 import Callout from 'components/shared/Callout'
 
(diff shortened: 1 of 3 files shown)


src/hooks/IpfsCache.ts
```diff
@@ -49,15 +49,6 @@ export function useIpfsCache<T extends IpfsCacheName>(
               : (data.data as unknown as IpfsCacheData[T]),
           )
         }
-
-        // Unpin cache files, including latest if expired
-        // There should never be more than one cache file to unpin. But during high traffic it may be possible for multiple unique cache files to be uploaded by different users simultaneously
-        list.rows.slice(isExpired ? 0 : 1).reduce(
```
aeolianeth, line 55: This has moved to a cron job that runs on the server every 10 minutes.

src/components/shared/forms/ProjectDetailsForm.tsx
```diff
@@ -75,10 +74,7 @@ export default function ProjectDetailsForm({
             name="logoUri"
             initialUrl={form.getFieldValue('logoUri')}
             onSuccess={logoUri => {
-              const prevUrl = form.getFieldValue('logoUri')
-              // Unpin previous file
               form.setFieldsValue({ logoUri })
-              if (prevUrl) unpinIpfsFileByCid(cidFromUrl(prevUrl))
```
aeolianeth, line 81: This will mean files will pile up, but this is a low-volume interaction, so not too bad.

We can solve once we move to Next.

johnnyd-eth (approved):
Beast pick up and fix here @aeolianeth. Fantastic work

PR #1293, opened 2022-06-28
Rework token form copy

aeolianeth:

New token form copy.

Screen Shot 2022-06-28 at 1 47 19 PM

## Acceptance checklist

### DIFF
--- a/src/components/v2/shared/FundingCycleConfigurationDrawers/TokenDrawer.tsx
+++ b/src/components/v2/shared/FundingCycleConfigurationDrawers/TokenDrawer.tsx
@@ -1,4 +1,5 @@
 import { t, Trans } from '@lingui/macro'
+import Callout from 'components/shared/Callout'
 import FundingCycleDrawer from 'components/v2/shared/FundingCycleConfigurationDrawers/FundingCycleDrawer'
 import { useFundingCycleDrawer } from 'components/v2/shared/FundingCycleConfigurationDrawers/useFundingCycleDrawer'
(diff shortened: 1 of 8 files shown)


src/components/v2/shared/FundingCycleConfigurationDrawers/TokenDrawer.tsx
```diff
@@ -30,8 +31,16 @@ export default function TokenDrawer({
         onClose={handleDrawerCloseClick}
       >
         <p>
-          <Trans>Configure your project's tokenomics.</Trans>
+          <Trans>Design how your tokens should work.</Trans>
         </p>
+        <Callout style={{ marginBottom: '2rem' }}>
+          <Trans>
+            Project tokens <strong>aren't ERC-20 tokens</strong> by default.
```
wraeth-eth, line 38: I think maybe you just want to bold *aren't* . This way it stands out more to users

wraeth-eth (approved):
Just one comment, otherwise think its g2g 👍

src/components/v2/shared/FundingCycleConfigurationDrawers/TokenDrawer.tsx
aeolianeth: Hmm, i dont feel strongly but i kinda wanted to emphasize **ERC-20 tokens** too, because I thought that's what users might be looking for/will resonate more strongly.

Give you've approved I'll merge, but let's def follow up if it doesn't feel right.

PR #1296, opened 2022-06-28
Set tx metadata args to constant 0 value

aeolianeth:

Random bytes may cause issue in future. Plus it looks weird in the on-chain logs. Setting to 0 is a better soln.


aeolianeth:
@wraeth-eth could u pls review when available?

wraeth-eth (approved):
Approved ser - merge at your leisure

PR #1315, opened 2022-07-02
Fix misrounding formattedNum

johnnyd-eth:
BEFORE: 

Screen Shot 2022-07-02 at 12 03 42 pm

With the code before this PR, passing precision into `formattedWad()` would result in the bottom one being rounded to .99 which is actually incorrect. 

AFTER:

Screen Shot 2022-07-02 at 12 00 00 pm

## Acceptance checklist

### DIFF
--- a/src/utils/formatNumber.ts
+++ b/src/utils/formatNumber.ts
@@ -1,5 +1,6 @@
 import { BigNumber, BigNumberish } from '@ethersproject/bignumber'
 import { formatUnits, parseUnits } from '@ethersproject/units'
+import { round } from 'lodash'
 
 import { WAD_DECIMALS } from 'constants/numbers'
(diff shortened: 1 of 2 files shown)


src/utils/formatNumber.ts
```diff
@@ -42,11 +43,8 @@ export const parseWad = (value?: BigNumberish) =>
  * fromWad(1000000000000000000);
  *
  */
-export const fromWad = (
-  wadValue?: BigNumberish,
-  decimals: number = WAD_DECIMALS,
```
johnnyd-eth, line 47: Move the rounding functionality out of `fromWad` and into `formattedNum`

aeolianeth (approved):

Issue #1241, opened 2022-06-15
NextJS prep: Organize files to better reflect nextJS structure


aeolianeth:
- Create pages/ directory

PR #1330, opened 2022-07-04
NextJS Prep: create pages/ directory

aeolianeth:
- Moves all pages code to `pages/`. The directory structure matches the route structure. This is to prepare for NextJS, which requires this structure.
- Moves all `components/shared/` to `components/`. Everything in `components/` can be considered shared. We can better organise this directory later. 

closes https://github.com/jbx-protocol/juice-interface/issues/1241


aeolianeth:
@wraeth-eth could you please take another look at this?

I decided to create a new PR because the old one was too out of date.

src/pages/p/index.tsx
```diff
@@ -1,5 +1,5 @@
-import FeedbackFormButton from 'components/shared/FeedbackFormButton'
```
wraeth-eth, line 1: Question: do we need these to be 1:1 mapping to routes (`p`), or could we give this folder a more meaningful name?

wraeth-eth: Also, in this PR, could we take the chance to move code out of index files and have them as pure redirects for files?

wraeth-eth:
Should [src/pages/v2-bug-notice.tsx](https://github.com/jbx-protocol/juice-interface/pull/1330/files#diff-4016df2415dc90d5fd4d1b1ddc0966ff95442c41af50465ec4ecb11749bd43a0) be in top level `pages/`

src/pages/p/index.tsx
aeolianeth: @wraeth-eth  yup, needs to be a 1:1 mapping.

Next routes are built from the file tree.

It also uses `index` files to determine routes. `pages/index.ts` is the homepage at route `/`

wraeth-eth: sounds good. How do you feel about having something like this:
```ts
// test/Test.tsx
export const Test = () => {
  return <></>
}

// index.ts
import {Test} from './Test'

export default Test

```

The reason behind this is I think it is a bit cleaner - index.ts (to me) should be just for re-routing exports. It also allows faster fuzzy searching too

wraeth-eth (approved):

src/pages/p/index.tsx
aeolianeth: Yeah I agree def. Think it's juuust outside scope though so I created this issue: https://github.com/jbx-protocol/juice-interface/issues/1331

Thanks for raising this 🙏

PR #1340, opened 2022-07-05
Bump @jbx-protocol/contracts-v1 to v2.0.0

aeolianeth:
What it says on the tin!

This PR upgrades the V1 contract package version. This new version is itself a reshuffling of dependencies. See https://github.com/jbx-protocol/juice-contracts-v1/pull/20 for the diff.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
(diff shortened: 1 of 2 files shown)


aeolianeth:
@johnnyd-eth could u review this instead? 🙏

johnnyd-eth:
Smoke tested and LGTM

johnnyd-eth (approved):

Issue #1223, opened 2022-06-13
Show images as a delete-able preview in the pay memo input field


johnnyd-eth:

0xbakedpotato:
I'm happy to grab this one. @johnnyd-eth, is there any more context here that I'd need?

johnnyd-eth:
> I'm happy to grab this one. @johnnyd-eth, is there any more context here that I'd need?

Sweet @0xbakedpotato! And I don't think so - we just want images that are attached from the 'Add image' or the new sticker modal to show up as images instead of their URL's mainly. But also would be great to remove with them with an X button

aeolianeth:
@benschac I'm going to give this one to @wraeth-eth because he's working on related work.

benschac:
I'm pretty close to finishing it up. I'd likely have a PR up tomorrow. Also, not married to the task in @wraeth-eth wants to grab it.

PR #1348, opened 2022-07-06
V2 Pay modal improvements

aeolianeth:
## What does this PR do and why?

Simplifies the V2 pay modal

Closes https://github.com/jbx-protocol/juice-interface/issues/1287

## Screenshots or screen recordings

| basic | risks + notice | expanded | not accepted |
| --- | --- | --- | --- |
|  Screen Shot 2022-07-06 at 6 59 45 PM | Screen Shot 2022-07-06 at 6 57 04 PM | Screen Shot 2022-07-06 at 6 57 22 PM | Screen Shot 2022-07-06 at 6 57 11 PM |

## Acceptance checklist

- [ ] I have evaluated the [Approval Guidelines](https://github.com/jbx-protocol/juice-interface/blob/main/CONTRIBUTING.md#approval-guidelines) for this PR.
- [ ] I have tested this PR in [all supported browsers](https://github.com/jbx-protocol/juice-interface/blob/main/CONTRIBUTING.md#supported-browsers).
- [ ] I have tested this PR in dark mode and light mode (if applicable).

PR #1349, opened 2022-07-06
Change Sticker Support on Pay

wraeth-eth:

Closes https://github.com/jbx-protocol/juice-interface/issues/1223

- Significantly reworks the pay form to be more inline with antd expected use.
- Adds a new line of stickers to bottom of form (see pic) and is no longer included in size count.
- User uploaded image is no longer included in size count either

image

## Acceptance checklist

### DIFF
--- a/src/components/IconedImage/IconedImage.tsx
+++ b/src/components/IconedImage/IconedImage.tsx
@@ -16,7 +16,7 @@ export const IconedImage = ({
   onClick: VoidFunction
 }) => {
   return (
-    <div style={{ position: 'relative' }}>
+    <div style={{ position: 'relative', padding: '1rem 0' }}>
       <Image key={url} width={width} src={url} />
       <div
         role="button"
--- a/src/components/IconedImage/IconedImage.tsx
+++ b/src/components/IconedImage/IconedImage.tsx
@@ -1,0 +1,35 @@
+import { Image } from 'antd'
+import { ReactNode } from 'react'
+
+/**
+ * Shows an {@link Image} with a small icon in the top left.
+ */
+export const IconedImage = ({
+  url,
+  width,
+  onClick,
+  icon,
+}: {
+  url: string
+  width: number
+  icon: ReactNode
+  onClick: VoidFunction
+}) => {
+  return (
+    <div style={{ position: 'relative' }}>
+      <Image key={url} width={width} src={url} />
+      <div
+        role="button"
+        style={{
+          cursor: 'pointer',
+          position: 'absolute',
+          top: 0,
+          left: 0,
+        }}
+        onClick={onClick}
+      >
+        {icon}
+      </div>
+    </div>
+  )
+}
--- a/src/components/IconedImage/IconedImage.tsx
+++ b/src/components/IconedImage/IconedImage.tsx
@@ -24,7 +24,7 @@ export const IconedImage = ({
           cursor: 'pointer',
           position: 'absolute',
           top: 0,
-          left: 0,
+          right: 0,
         }}
         onClick={onClick}
       >
--- a/src/components/modals/AttachStickerModal/AttachableSticker.tsx
+++ b/src/components/modals/AttachStickerModal/AttachableSticker.tsx
@@ -13,7 +13,7 @@ export const AttachableSticker = ({
       style={{
         display: 'flex',
         justifyContent: 'center',
-        alignItems: 'middle',
+        alignItems: 'center',
         padding: '1rem 0',
         cursor: 'pointer',
       }}
--- a/src/components/modals/AttachStickerModal/AttachableSticker.tsx
+++ b/src/components/modals/AttachStickerModal/AttachableSticker.tsx
@@ -14,7 +14,7 @@ export const AttachableSticker = ({
         display: 'flex',
         justifyContent: 'center',
         alignItems: 'middle',
-        padding: '20px 0',
+        padding: '1rem 0',
         cursor: 'pointer',
       }}
       className="hover-bg-l2"
--- a/src/components/modals/AttachStickerModal/AttachableSticker.tsx
+++ b/src/components/modals/AttachStickerModal/AttachableSticker.tsx
@@ -1,0 +1,26 @@
+import { PaymentMemoSticker } from './paymentMemoSticker'
+
+export const AttachableSticker = ({
+  sticker,
+  onClick,
+}: {
+  sticker: PaymentMemoSticker
+  onClick: VoidFunction
+}) => {
+  return (
+    <div
+      role="button"
+      style={{
+        display: 'flex',
+        justifyContent: 'center',
+        alignItems: 'middle',
+        padding: '20px 0',
+        cursor: 'pointer',
+      }}
+      className="hover-bg-l2"
+      onClick={onClick}
+    >
+      <img src={sticker.filepath} alt={sticker.alt} height="75px" />
+    </div>
+  )
+}
--- a/src/components/v2/V2Project/V2PayButton/V2ConfirmPayModal/index.tsx
+++ b/src/components/v2/V2Project/V2PayButton/V2ConfirmPayModal/index.tsx
@@ -1,10 +1,8 @@
 import { BigNumber } from '@ethersproject/bignumber'
-import { isAddress } from '@ethersproject/address'
 import { t, Trans } from '@lingui/macro'
-import { Checkbox, Descriptions, Form, Space, Switch } from 'antd'
+import { Descriptions, Space } from 'antd'
 import { useForm } from 'antd/lib/form/Form'
 import FormattedAddress from 'components/FormattedAddress'
-import ImageUploader from 'components/inputs/ImageUploader'
 import { NetworkContext } from 'contexts/networkContext'
 import { useCurrencyConverter } from 'hooks/CurrencyConverter'
 import { V2ProjectContext } from 'contexts/v2/projectContext'
@@ -20,8 +18,6 @@ import {
 } from 'utils/v2/currency'
 import { usePayV2ProjectTx } from 'hooks/v2/transactor/PayV2ProjectTx'
 
-import * as constants from '@ethersproject/constants'
-
 import {
   getUnsafeV2FundingCycleProperties,
   V2FundingCycleRiskCount,
@@ -33,8 +29,8 @@ import { weightedAmount } from 'utils/v2/math'
 
 import TransactionModal from 'components/TransactionModal'
 import ProjectRiskNotice from 'components/ProjectRiskNotice'
-import MemoFormItem from 'components/inputs/Pay/MemoFormItem'
-import { EthAddressInput } from 'components/inputs/EthAddressInput'
+
+import { V2PayForm, V2PayFormType } from '../V2PayForm'
 
 export default function V2ConfirmPayModal({
   visible,
@@ -53,20 +49,15 @@ export default function V2ConfirmPayModal({
     fundingCycleMetadata,
     projectMetadata,
     projectId,
-    tokenAddress,
     tokenSymbol,
   } = useContext(V2ProjectContext)
   const converter = useCurrencyConverter()
   const payProjectTx = usePayV2ProjectTx()
 
   const [loading, setLoading] = useState<boolean>()
-  const [preferClaimed, setPreferClaimed] = useState<boolean>(false)
-  const [customBeneficiaryEnabled, setCustomBeneficiaryEnabled] =
-    useState<boolean>(false)
-  const [memo, setMemo] = useState<string>('')
   const [transactionPending, setTransactionPending] = useState<boolean>()
 
-  const [form] = useForm<{ beneficiary: string }>()
+  const [form] = useForm<V2PayFormType>()
 
   const usdAmount = converter.weiToUsd(weiAmount)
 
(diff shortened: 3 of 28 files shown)


src/components/IconedImage/IconedImage.tsx
```diff
@@ -0,0 +1,35 @@
...
+  url: string
+  width: number
+  icon: ReactNode
+  onClick: VoidFunction
+}) => {
+  return (
+    <div style={{ position: 'relative' }}>
+      <Image key={url} width={width} src={url} />
+      <div
+        role="button"
```
aeolianeth, line 22: **praise** ahhh, thank you.

src/components/v2/V2Project/V2PayButton/V2PayForm/StickerSelection.tsx
```diff
@@ -0,0 +1,46 @@
...
+  const handleImageDeletion = (i: number) => {
+    if (!value) {
+      onChange?.([])
+      return
+    }
+    const editedImages = [...value.slice(0, i), ...value.slice(i + 1)]
+    onChange?.(editedImages)
+  }
+  return (
+    <>
```
aeolianeth, line 26: **suggestion**: think we can remove the fragment here.

src/constants/v2/projectPreferences.ts
```diff
@@ -0,0 +1,3 @@
+export const ProjectPreferences = {
+  STICKER_MAX: 3,
```
aeolianeth, line 2: **suggestion** if this isn't gonna be used anywhere, probably fine to keep this const in the component itself. wdyt?

src/components/v2/V2Project/V2PayButton/V2PayForm/V2PayForm.tsx
```diff
@@ -0,0 +1,158 @@
...
+
+  const [customBeneficiaryEnabled, setCustomBeneficiaryEnabled] =
+    useState<boolean>(false)
+
+  const [attachStickerModalVisible, setAttachStickerModalVisible] =
+    useState<boolean>(false)
+
+  const hasIssuedTokens = tokenAddress !== constants.AddressZero
+
+  const stickerUrls = useWatch('stickerUrls', form)
```
aeolianeth, line 40: ohhhhhhh. huge.

src/components/v2/V2Project/V2PayButton/V2ConfirmPayModal/index.tsx
```diff
@@ -212,67 +213,7 @@ export default function V2ConfirmPayModal({
...
-                    wallet. Leave unchecked to have your token balance tracked
-                    by Juicebox, saving gas on this transaction. You can always
-                    claim your ERC-20 tokens later.
-                  </Trans>
-                </label>
-              </Space>
-            </Form.Item>
-          )}
-        </Form>
+        <V2PayForm form={form} />
```
aeolianeth, line 216: **praise** it had to be done. Nice work!

src/components/v2/V2Project/V2PayButton/V2PayForm/V2PayForm.tsx
```diff
@@ -0,0 +1,158 @@
...
+
+  return (
+    <>
+      <Form form={form} layout="vertical">
+        <Form.Item
+          label={t`Memo (optional)`}
+          className={'antd-no-number-handler'}
+          extra={t`Add an on-chain memo to this payment.`}
+        >
+          <Form.Item noStyle name="memo">
```
aeolianeth, line 53: **question**: i think we still have [`MemoFormItem.tsx`](https://github.com/jbx-protocol/juice-interface/blob/290fe0226f09f13c3f59fa0d1ab039641c1450cf/src/components/inputs/Pay/MemoFormItem.tsx). Is it worth trying to move any new code into that component, and using an instance of it here?

src/components/v2/V2Project/V2PayButton/V2ConfirmPayModal/index.tsx
```diff
@@ -106,10 +101,26 @@ export default function V2ConfirmPayModal({
...
+    if (stickerUrls) {
+      memoMetadata = { ...memoMetadata, stickerUrls }
+    }
+
+    const memo =
+      (textMemo ?? '') +
+      (Object.keys(memoMetadata).length
+        ? '\n' + JSON.stringify(memoMetadata)
+        : '')
+
```
aeolianeth, line 119: **thought**: I know we talked about this, but im wondering if this is overkill for now.

WDYT about appending stickers and the image to the memo instead, with no JSON?

```
<memo text> <sticker urls> <uploaded image>
```

With this, we can still differentiate the uploaded image from stickers when we re-work the payment activity element.

src/components/modals/AttachStickerModal/AttachableSticker.tsx
```diff
@@ -0,0 +1,26 @@
...
+  onClick: VoidFunction
+}) => {
+  return (
+    <div
+      role="button"
+      style={{
+        display: 'flex',
+        justifyContent: 'center',
+        alignItems: 'middle',
+        padding: '20px 0',
```
aeolianeth, line 17: **suggestion** might `rem` this

src/components/v2/V2Project/V2PayButton/V2PayForm/V2PayForm.tsx
aeolianeth: Also TIL `noStyle` prop

aeolianeth (changes requested):
@wraeth-eth killer PR.

I left a few minor comments but on the whole, lgtm.

src/components/IconedImage/IconedImage.tsx
```diff
@@ -0,0 +1,35 @@
...
+}: {
+  url: string
+  width: number
+  icon: ReactNode
+  onClick: VoidFunction
+}) => {
+  return (
+    <div style={{ position: 'relative' }}>
+      <Image key={url} width={width} src={url} />
+      <div
```
aeolianeth, line 21: I think we should:
- put this on the right-hand side
- make it a bit bigger
- potentially change color to make it more salient.

I think we could either:

- use antd `Badge` component https://ant.design/components/badge/
- Use similar design to the image uploader. 

Screen Shot 2022-07-06 at 11 32 41 PM

src/components/modals/AttachStickerModal/AttachableSticker.tsx
```diff
@@ -0,0 +1,26 @@
...
+  sticker: PaymentMemoSticker
+  onClick: VoidFunction
+}) => {
+  return (
+    <div
+      role="button"
+      style={{
+        display: 'flex',
+        justifyContent: 'center',
+        alignItems: 'middle',
```
benschac, line 16: small nit: https://developer.mozilla.org/en-US/docs/Web/CSS/align-items

`middle` --> `center`

src/constants/v2/projectPreferences.ts
wraeth-eth: I'll probably use it on reading the stickers from the tx on pay feed

src/components/v2/V2Project/V2PayButton/V2PayForm/V2PayForm.tsx
wraeth-eth: We could yeah - only problem with that class is it doesn't conform to what antd expect (value/onChange) so it is bit of extra work. I think MemoFormItem is only used in v1 too. I am happy to do it though if you think it is valuable to update v1 code

src/components/v2/V2Project/V2PayButton/V2ConfirmPayModal/index.tsx
wraeth-eth: I don't mind - reasoning behind this (especially with the version flag) is it allows us to update the format later and still provide backward compatibility

src/components/v2/V2Project/V2PayButton/V2PayForm/StickerSelection.tsx
wraeth-eth: we can! :)

src/components/v2/V2Project/V2PayButton/V2PayForm/V2PayForm.tsx
wraeth-eth: nvm - it actually does conform to it. I'll look at moving

wraeth-eth: Actually I think it is more complicated than initially seemed. I think if we want to reuse it, lets revisit then. MemoFormItem is only used in v1 pay atm.

src/components/v2/V2Project/V2PayButton/V2ConfirmPayModal/index.tsx
aeolianeth: I think let's roll with the simplest, boring solution. + it saves characters, saving a touch of gas ⛽️.

Do you foresee us having to update the format with any regular frequency?

src/components/v2/V2Project/V2PayButton/V2ConfirmPayModal/index.tsx
```diff
@@ -33,8 +29,8 @@ import { weightedAmount } from 'utils/v2/math'
 
 import TransactionModal from 'components/TransactionModal'
 import ProjectRiskNotice from 'components/ProjectRiskNotice'
-import MemoFormItem from 'components/inputs/Pay/MemoFormItem'
```
aeolianeth, line 36: @wraeth-eth MemoFormItem used here in V2.

src/components/v2/V2Project/V2PayButton/V2PayForm/V2PayForm.tsx
```diff
@@ -0,0 +1,158 @@
...
+              <label htmlFor="preferClaimed">
+                <Trans>
+                  Check this to mint this project's ERC-20 tokens to your
+                  wallet. Leave unchecked to have your token balance tracked by
+                  Juicebox, saving gas on this transaction. You can always claim
+                  your ERC-20 tokens later.
+                </Trans>
+              </label>
+            </Space>
+          </Form.Item>
```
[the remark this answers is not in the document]
wraeth-eth: I'm having issues with this:
image

src/components/v2/V2Project/V2PayButton/V2PayForm/V2PayForm.tsx
aeolianeth: We were using MemoFormItem in V2 https://github.com/jbx-protocol/juice-interface/pull/1349/files#r915355021

It would be nice to re-use in V1 and V2 if possible. Otherwise we're duplicating code. We can leave this for a follow-up PR though - really want to get this PR in as-is.

src/components/IconedImage/IconedImage.tsx
wraeth-eth: I think badge is the wrong component here - looks like its used for notifications rather than a close button. I also tried out the blue, but it feels a bit distracting to the form. wdyt about this
image

Its not exactly the same as image uploader, but I think thats ok too

src/components/v2/V2Project/V2PayButton/V2PayForm/V2PayForm.tsx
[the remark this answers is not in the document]
aeolianeth: Oh yeah sorry, you need this too https://github.com/jbx-protocol/juice-interface/pull/1348/files#diff-b8e4a608443da6dd28b7b2a735602d032bba146eef48a0687afc30e1cb09fefc

Screen Shot 2022-07-07 at 10 24 46 AM

src/components/IconedImage/IconedImage.tsx
aeolianeth: Looks perfect! Great call.

src/components/v2/V2Project/V2PayButton/V2PayForm/V2PayForm.tsx
```diff
@@ -0,0 +1,160 @@
...
+        </Form.Item>
+        <Form.Item extra={t`Mint tokens to a custom address.`}>
+          <span style={{ color: colors.text.primary, fontWeight: 500 }}>
+            <Trans>Custom token beneficiary</Trans>
+          </span>
+          <Switch
+            checked={customBeneficiaryEnabled}
+            onChange={setCustomBeneficiaryEnabled}
+            style={{ marginLeft: 10 }}
+          />
```
aeolianeth, line 99: @wraeth-eth sorry one more thing. Can we flip the switch and the text the other way around?

```
<switch> Custom token beneficiary
```

src/components/v2/V2Project/V2PayButton/V2PayForm/V2PayForm.tsx
wraeth-eth: yeah lets do it - don't mind following up in separate PR
https://github.com/jbx-protocol/juice-interface/issues/1354
❤️

src/components/v2/V2Project/V2PayButton/V2ConfirmPayModal/index.tsx
wraeth-eth: True true, hmm... I think the trouble with not having any way to control the format now means we cannot change it later in the future. What happens if we want to allow more expression in the memo? I dunno, wdyt? It buys us more flexibility for a trivial gas increase. If you want it looking better on chain, we could base64 encode it? ;)

src/components/v2/V2Project/V2PayButton/V2PayForm/V2PayForm.tsx
wraeth-eth: image

wraeth-eth: image
dropping the margin to align better wdyt

src/components/v2/V2Project/V2PayButton/V2ConfirmPayModal/index.tsx
aeolianeth: I don't feel strongly. Want to avoid over-engineering and opt for the most boring solution in general IMO.

Let's ship this as-is 👍

src/components/v2/V2Project/V2PayButton/V2PayForm/V2PayForm.tsx
aeolianeth: Yep nailed it

aeolianeth (approved):
@wraeth-eth so tight, much love.

Merging.

PR #1366, opened 2022-07-09
Safe access for Web APIs

jmilldotdev:

For our NextJS migration, provides safe access to all usages of `localStorage`, `window`, and `navigator` found in our codebase. All still seems to be stable, but this should probably be thoroughly reviewed.

https://nextjs.org/docs/migrating/from-create-react-app#safely-accessing-web-apis

## Acceptance checklist

### DIFF
--- a/src/components/Navbar/MenuItems.tsx
+++ b/src/components/Navbar/MenuItems.tsx
@@ -98,13 +98,15 @@ export function TopLeftNavItems({
         text={t`FAQ`}
         route={undefined}
         onClick={() => {
-          if (onClickMenuItems) onClickMenuItems()
-          window.location.hash = '/'
-          setTimeout(() => {
-            document
-              .getElementById('faq')
-              ?.scrollIntoView({ behavior: 'smooth' })
-          }, 0)
+          if (typeof window !== 'undefined') {
+            if (onClickMenuItems) onClickMenuItems()
+            window.location.hash = '/'
+            setTimeout(() => {
+              document
+                .getElementById('faq')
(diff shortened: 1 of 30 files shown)


src/utils/windowUtils.ts
```diff
@@ -0,0 +1,18 @@
+export const reloadWindow = () => {
```
aeolianeth, line 1: **praise** great call with these utils 👏

src/components/Navbar/MenuItems.tsx
```diff
@@ -98,13 +98,15 @@ export function TopLeftNavItems({
...
         route={undefined}
         onClick={() => {
-          if (onClickMenuItems) onClickMenuItems()
-          window.location.hash = '/'
-          setTimeout(() => {
-            document
-              .getElementById('faq')
-              ?.scrollIntoView({ behavior: 'smooth' })
-          }, 0)
+          if (typeof window !== 'undefined') {
```
aeolianeth, line 101: **nitpick (non-blocking)** super minor, but I find it more readable to avoid nesting code in this block and return early using the inverse logic.

```js
if (typeof window === 'undefined') {
    return
}
// ...code
```

Like this: https://github.com/jbx-protocol/juice-interface/pull/1366/files#diff-8c58fee43e99a62af46f20eba09c7918f17b7d4fc16e6822e5fa566fb7c49bb8R59-R61

aeolianeth (approved):
@jmilldotdev  Thanks so much for this. I had a click around and everything seems fine. This should set us up well for Next.

Merging!

Issue #1467, opened 2022-07-21
Can't pay some projects


tomquirk:
## Summary

Ongoing investigation. Can't pay 

- https://www.juicebox.money/@galaxybrain.nftstory
- https://www.juicebox.money/v2/p/103

PR #1474, opened 2022-07-21
Force set useDataSourceForPay to false

tomquirk:
Remediation for https://github.com/jbx-protocol/juice-interface/issues/1467

Affected projects can reconfig their project with no modifications, and `useDataSourceForPay` will revert from `true` to `false. 

This means their project's payments will work again.


johnnyd-eth (approved):
LGTM

PR #1551, opened 2022-07-26
Add basic validation to github archive endpoint

tomquirk:
As per title.

I think rogue issues are being created: https://github.com/jbx-protocol/juice-interface/issues/1549

This PR should prevent that.

### DIFF
--- a/src/pages/api/github/archive-project.page.ts
+++ b/src/pages/api/github/archive-project.page.ts
@@ -30,12 +30,16 @@ const handler = async (
   try {
     const { archived, projectId, metadata, handle, cv } = req.body
 
+    if (!projectId || !metadata || !handle) {
+      throw new Error()
+    }
+
     const headers = {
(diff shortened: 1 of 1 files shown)


tomquirk:
@jmilldotdev may you pls review? 🙏

jmilldotdev:
LGTM, though one concern is that anyone could spoof these requests. I doubt we'd ever see this happen but it could be worth thinking about how we could have a project owner sign a nonce to submit this request.

jmilldotdev (approved):

tomquirk:
> LGTM, though one concern is that anyone could spoof these requests for any project without a traditional concept of authed requests. I doubt we'd ever see this happen but it could be worth thinking about how we could have a project owner sign a nonce to submit this request.

definitely, great callout. We do make the project owner submit a tx (so there's gas cost) so we could probably do something more legit there.

PR #1556, opened 2022-07-26
Add ipfs fallback tools

jmilldotdev:

Adds some utilities for getting IPFS resources with fallback to the default gateway, and rework veNFT asset fetching to use these utilities.

## Acceptance checklist


src/components/FallbackImage.tsx
```diff
@@ -0,0 +1,28 @@
...
+  useEffect(() => {
+    setImgSrc(src)
+  }, [src])
+
+  return (
+    <Image
+      {...rest}
+      src={imgSrc ? imgSrc : ''}
+      onError={() => {
+        setImgSrc(fallbackSrc)
```
tomquirk, line 22: **question** antd Image has a `fallback` prop. I've never used it, but is it useful here potentially?

https://ant.design/components/image/

src/components/FallbackImage.tsx
```diff
@@ -0,0 +1,28 @@
+import { useEffect, useState } from 'react'
+import { Image, ImageProps } from 'antd'
+
+interface FallbackImageProps {
+  src: string | undefined
+  fallbackSrc: string | undefined
+  rest: ImageProps
+}
+
+const FallbackImage = ({ src, fallbackSrc, rest }: FallbackImageProps) => {
```
tomquirk, line 10: **suggestion** could we rename `rest` to `...props`? That way we don't need to pass in `rest` prop: can just pass all the props along side `src` and `fallbackSrc`

src/utils/ipfs.ts
```diff
@@ -54,11 +54,29 @@ export const logoNameForHandle = (handle: string) => `juicebox-@${handle}-logo`
 export const metadataNameForHandle = (handle: string) =>
   `juicebox-@${handle}-metadata`
 
-export const ipfsCidUrl = (hash: string) =>
-  `https://${IPFS_GATEWAY_HOSTNAME}/ipfs/${hash}`
+export const ipfsCidUrl = (hash: string, useFallback = false) => {
```
tomquirk, line 57: **suggestion** for readability purposes, miight be better to have an object prop that includes `useFallback`.

That way instead of `ipfsCidUrl(hash, true)` (i dont know what `true` means without looking at the function internals), it would read `ipfsCidUrl(hash, { useFallback: true })

tomquirk (commented):
Looks good @jmilldotdev just a couple comments

src/utils/ipfs.ts
jmilldotdev: too much python

will do

src/components/FallbackImage.tsx
jmilldotdev: using next image component fallback instead

src/components/FallbackImage.tsx
jmilldotdev: Good call, I was looking at Next Image docs for some reason

jmilldotdev:
responded

src/components/veNft/VeNftCarouselItem.tsx
```diff
@@ -34,24 +36,29 @@ const VeNftCarouselItem = ({
...
+  )
 
   return (
     <Col span={8}>
       {variant && (
         <>
           <Image
-            src={getCarouselImage()}
+            src={getCarouselImage({ useFallback: false })}
+            fallback={getCarouselImage({ useFallback: true })}
```
tomquirk, line 61: Thats dope. Non-blocking, but I guess we could add some better "empty image" state here. Like a grey banana emoji or something.

tomquirk (approved):

PR #1607, opened 2022-07-29
Only show network warning modal wen connected

tomquirk:

It's a bandaid, but it's something.


src/pages/_app.page.tsx
```diff
@@ -36,9 +36,10 @@ const AppWrapper: React.FC = ({ children }) => {
   ]
 
   useEffect(() => {
-    if (!signerNetwork) return
+    if (!signerNetwork || !userAddress) return
```
tomquirk, line 39: If the user isn't connected, bail early.

When the user is connected AND also on the wrong network, then we'll still show the annoying modal.

tomquirk:
@jmilldotdev wdyt?

jmilldotdev (approved):
LGTM

PR #1649, opened 2022-08-03
[1/2] Remove string origins from translation files

tomquirk:
As per title!

These origin strings serve no functional purpose. They lead to bigger diffs and higher potential for merge conflicts.


jmilldotdev (approved):
HELL
YES

PR #1677, opened 2022-08-04
Move some landing page strings into own file

tomquirk:
Want to make it easier for anyone to update the landing page copy. We can point folks to this specific file and encourage them to make changes and submit a PR.


johnnyd-eth (approved):
Hell yes. Very good idea

Issue #1636, opened 2022-08-01
Sticker Upload in V1 projects adds sticker URL to input memo form


benschac:
## Summary

- Go to v1 project.
- Add sticker
- Old sticker URL Memo is shown in the input field, not matching functionality in v2 projects

## Steps to reproduce

1. Go to https://www.juicebox.money/p/juicebox
2. Click on pay, then customize with sticker 

## What is the _current_ bug behavior?
- Old sticker URL Memo is shown in the input field, not matching functionality in v2 projects

## What is the expected _correct_ behavior?
- Add a sticker on a v2 project, that is the correct behavior.
## Relevant logs and/or screenshots
<img width="1276" alt="CleanShot 2022-08-01 at 10 00 35@2x" src="https://user-images.githubusercontent.com/2502947/182165113-aa0c8bea-4920-43e5-87aa-84b2dddf7149.png">

## Environment

- **Device/OS**: [e.g. iOS]
- **Browser:** [e.g. chrome, safari]
- **Hardware wallet**: [e.g. Ledger w/ Metamask]

PR #1705, opened 2022-08-09
Sticker upload for v1 projects

NakedFool:

Link to an issue - https://github.com/jbx-protocol/juice-interface/issues/1636

The intention of this PR is for sticker upload on v1 projects to have the same functionality as in v2 projects.

Most of the work is done in `V1ConfirmPayOwnerModal.tsx` since I had to update form functionality and prepare a payment memo for `pay` function.

If you have any questions or suggestions, please let me know.

https://user-images.githubusercontent.com/18723426/183588442-d04bbfb7-1dd6-47ba-82ab-4bb08acbdd1d.mov

## Acceptance checklist


src/components/v1/V1Project/modals/V1ConfirmPayOwnerModal.tsx
```diff
@@ -195,43 +229,87 @@ export default function V1ConfirmPayOwnerModal({
             name="memo"
             label={t`Memo (optional)`}
             className={'antd-no-number-handler'}
-            extra={t`Add an on-chain memo to this payment.`}
+            style={{ marginBottom: 0 }}
           >
-            <MemoFormInput />
+            <Form.Item
```
johnnyd-eth, line 234: **Issue:** we want to avoid nested `<Form.Item`s like this. Can we make it work with a single `memo` item?

src/components/v1/V1Project/modals/V1ConfirmPayOwnerModal.tsx
```diff
@@ -1,33 +1,48 @@
...
+import { StickerSelection } from 'components/Project/StickerSelection'
+
+import { buildPaymentMemo } from 'utils/buildPaymentMemo'
+
+import { ThemeContext } from 'contexts/themeContext'
 
 import { V1_CURRENCY_ETH, V1_CURRENCY_USD } from 'constants/v1/currency'
+import { ProjectPreferences } from 'constants/v1/projectPreferences'
+
+export interface V1PayFormType {
```
johnnyd-eth, line 40: **Nitpick:** don't need `export` here for now

src/components/v1/V1Project/modals/V1ConfirmPayOwnerModal.tsx
```diff
@@ -195,43 +229,87 @@ export default function V1ConfirmPayOwnerModal({
...
+                <Trans>Receive ERC-20</Trans>
+              </Checkbox>
             </Form.Item>
           )}
         </Form>
+        <AttachStickerModal
+          visible={attachStickerModalVisible}
+          onClose={() => setAttachStickerModalVisible(false)}
+          onSelect={sticker => {
+            if (typeof window === 'undefined') {
```
johnnyd-eth, line 301: **Suggestion:** would be great to separate this code out into a `handleStickerSelect` function

src/components/v1/V1Project/modals/V1ConfirmPayOwnerModal.tsx
```diff
@@ -195,43 +229,87 @@ export default function V1ConfirmPayOwnerModal({
             name="memo"
             label={t`Memo (optional)`}
             className={'antd-no-number-handler'}
-            extra={t`Add an on-chain memo to this payment.`}
+            style={{ marginBottom: 0 }}
           >
-            <MemoFormInput />
```
johnnyd-eth, line 200: **Question:** why remove this? 

Lines `297:312` are repeated 3 times throughout the codebase I believe. I think it would be great if we could leverage the `<MemoFormInput` in both v1 and v2 pay modals to avoid as much repeated code as possible

johnnyd-eth:
Really good stuff here @NakedFool. Love that you made the effort to extract some code into shared functions. Just a few other comments from me!

src/components/v1/V1Project/modals/V1ConfirmPayOwnerModal.tsx
NakedFool: That is a reasonable assumption and question, i'll try to be concise and explain to you why I went this route.

`MemoFormInput` is acting as a shared component and in order to function in pay modal like form inputs are acting in v2 it would need some modifications.

Currently, it is outputting `value` and `value` is formed both from `Input.TextArea` and from `Sticker` components, therefore outputted value is a mix of text and URLs. 
The way `memo` is built in v2 components is that it is using `buildPaymentMemo` function and in the end, it is combining value based on text value, sticker URLs, and image URLs from all separate form items and passing it to `pay` function.

We are trying to achieve the same process in v1 and with the current logic from `MemoFormInput` it is not possible, however, what I can do is, I can change how `MemoFormInput` works and separate logic there, so it can output `text` and `stickerURLs` but it would require changes on all files that are using `MemoFormInput`.

That is absolutely not a problem for me, but I do not want to go far away before we are both on the same page and you understand what I am trying to achieve. 
If you find the above confusing, I'm open to jumping on a quick call and explaining everything to you ❤️

johnnyd-eth: Ok thanks for the explanation @NakedFool , too easy!

src/components/v1/V1Project/modals/V1ConfirmPayOwnerModal.tsx
johnnyd-eth: I jumped in and fixed this :)

johnnyd-eth (approved):

src/components/v1/V1Project/modals/V1ConfirmPayOwnerModal.tsx
NakedFool: Oh, i thought that comment was related to the one below, sorry!

PR #1755, opened 2022-08-15
dry up transactor try catch

benschac:

creating new PR pointed at main

Same PR, I just needed to create a new one because I accidentally merged the last one into `tiny-invariant` and not main 😅

## Acceptance checklist


benschac:
* main
  * **PR #1755** <a href="https://app.graphite.dev/github/pr/jbx-protocol/juice-interface/1755" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jbx-protocol/juice-interface/1755?utm_source=stack-comment).

src/hooks/v1/transactor/SafeTransferFromTx.ts
```diff
@@ -12,7 +13,15 @@ export function useSafeTransferFromTx(): TransactorInstance<{
   const { projectId, owner } = useContext(V1ProjectContext)
 
   return ({ newOwnerAddress }, txOpts) => {
-    if (!transactor || !projectId || !contracts?.Projects) {
+    invariant(transactor && contracts?.Projects && projectId && owner)
```
tomquirk, line 16: **question** I might need more coffee, but is this `invariant` call supposed to be within the `try` block? That way, when `invariant` throws, we're catching it?

tomquirk (commented):
sorry for the delay here @benschac it slipped my radar!

Just left 1 question

src/hooks/v1/transactor/SafeTransferFromTx.ts
benschac: yesss, good catch.

src/utils/transactor.ts
```diff
@@ -0,0 +1,69 @@
...
+  try {
+    invariant(transactor && contract, 'missing required')
+    const reqArgs = args.filter(arg => !optionalArgs.includes(arg))
+    const isMissingParameter =
+      args.length - optionalArgs.length !== reqArgs.length
+
+    if (isMissingParameter) {
+      throw new Error('missing parameter')
+    }
+
```
benschac, line 34: I rebased and got a bunch of other commits in this branch.


I pushed this branch to more explicitly show work.


I was trying to get the error logic into a function that we could easily use across the board. 

While something like this works @wraeth-eth made a great point that once we start using juice-sdk a lot of this code is going to go away and we're not going to need it.


In the short term, I'm just going to keep fixing these hanging pending states more explicitly in transactor hook calls.


In the long run, I'm sure we'll figure out a better way of write this logic with juice-sdk.

benschac (commented):
the one comment I left is the one worth checking. Sorry about the rebase noise.

src/utils/transactor.ts
benschac: cc: @jbx-protocol/frontend-core

PR #1772, opened 2022-08-21
Activity Share Button - Refactor Activity Event Elements - [1/a]

wraeth-eth:

First stage of adding share button to activity.

This brings all the different elements into a singly unifying component which will later be used to make changes globally to all the components.

No functional code changes here, just refactoring.

NA

## Acceptance checklist


wraeth-eth:
* main
  * **PR #1772** <a href="https://app.graphite.dev/github/pr/jbx-protocol/juice-interface/1772" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jbx-protocol/juice-interface/1772?utm_source=stack-comment).

src/components/activityEventElems/ActivityElement/activityElementEvent.ts
```diff
@@ -0,0 +1,5 @@
+import { PayEvent } from 'models/subgraph-entities/vX/pay-event'
+
+export type ActivityElementEvent = Pick<PayEvent, 'timestamp' | 'txHash'> & {
```
peripheralist, line 3: if this is a generic activity type, maybe we shouldn't tie it to a specific event (PayEvent)? and just define the properties explicitly

peripheralist (commented):
THANK YOUUU this rly needed to happen

Some thoughts on the pattern here. Instead of defining unique components that all return the same element with different args (PayEventElem, RedeemEventElem, etc), maybe we just define some parser functions that take a PayEvent,RedeemEvent,etc and output typed args:

Instead of just the `ActivityElementEvent` type, we could type all the args for the ActivityEventElem (which also seems to be the only place `ActivityElementEvent` is used):
```
export type ActivityEventElemProps = {
  header: string
  subject: string | JSX.Element | null
  extra?: string | JSX.Element | null
  event: ...event stuff previously typed ActivityElementEvent
}
```

Then in the ProjectActivity list component: instead of mapping events to unique components that are really the same components with different args, we could map events to `ActivityEventElemProps[]`, which are then all mapped to `ActivityElement` components

Thinking this would cut down on overall code complexity/repetition. Lmk if it doesn't make sense or if I'm missing anything

wraeth-eth:
accidentaly closed

tomquirk:
@wraeth-eth should we look at getting this over the line?

wraeth-eth:
For sure @aeolianeth - sorry, been busy with the create stuff

peripheralist:
hey @wraeth-eth 

referencing my feedback here: https://github.com/jbx-protocol/juice-interface/pull/1772#pullrequestreview-1081167558

i've been playing with implementing this and it does feel _clean_, but ends up further complicating our file structure. we already have event typed objects (DeployedERC20Event, etc) and also event components. adding additional "builder functions" that return a `DeployedERC20EventElemProps` value just feels overly complicated. let's ignore this for now

i've got work to do for adding the FC events anyway. want me to pick this up where you left it (getting other event elems using this system + maybe other stuff) so i can start working on those too?

wraeth-eth:
> hey @wraeth-eth
> 
> referencing my feedback here: [#1772 (review)](https://github.com/jbx-protocol/juice-interface/pull/1772#pullrequestreview-1081167558)
> 
> i've been playing with implementing this and it does feel _clean_, but ends up further complicating our file structure. we already have event typed objects (DeployedERC20Event, etc) and also event components. adding additional "builder functions" that return a `DeployedERC20EventElemProps` value just feels overly complicated. let's ignore this for now
> 
> i've got work to do for adding the FC events anyway. want me to pick this up where you left it (getting other event elems using this system + maybe other stuff) so i can start working on those too?

sounds good @peripheralist ! go for it

src/components/activityEventElems/ActivityElement/ActivityElement.tsx
```diff
@@ -0,0 +1,122 @@
...
+  )
+}
+
+function Header({ header }: { header: string }) {
+  const {
+    theme: { colors },
+  } = useContext(ThemeContext)
+  return (
+    <div style={smallHeaderStyle(colors)}>
+      <Trans>{header}</Trans>
```
peripheralist, line 44: @wraeth-eth have we tested if this \<Trans> works here? i see in some places you're passing a `t` object as the header arg

wraeth-eth: I dont think this would work - looks like a noob mistake on my part 3 months ago

wraeth-eth: The right course of action is the t object being passed

peripheralist: noooooooob

PR #1779, opened 2022-08-22
Fix erc20 onClose redirect

tomquirk:

Currently when erc-20 modal is closed, there's a redirect bug.

Screen Shot 2022-08-22 at 11 21 54 AM.

This pr fixes that


johnnyd-eth (approved):
I see. Nice catch and fix 🙌

Issue #1684, opened 2022-08-06
Cancelling redeem tx keeps loading state


johnnyd-eth:
## Summary

https://discord.com/channels/775859454780244028/866040669712678942/1005132408137457774

## Steps to reproduce

1. Go to redeeem some ETH on v2 project
2. Cancel the tx
3. See modal is loading and CAN'T BE STOPPED

## Environment

- **Device/OS**: [e.g. iOS]
- **Browser:** [e.g. chrome, safari]
- **Hardware wallet**: [e.g. Ledger w/ Metamask]

PR #1791, opened 2022-08-23
Added onError for claim and redeem modals

NakedFool:

Fixes - https://github.com/jbx-protocol/juice-interface/issues/1684

The initial problem was that when the user is trying to redeem tokens and if the transaction is canceled, spinner is loading infinitely.
This was also the case when the user is trying to claim tokens or burn tokens, so I applied the same logic in that modal as well.

For testing purposes, I have disabled validation checks so I can test them easily.
I have tested on Burn and Claim modal, Redeem is in the same file as Burn, they are using the same function to execute transactions so, therefore, the same behavior is going to be there.

https://user-images.githubusercontent.com/18723426/186086784-fc7dba2f-ed75-4021-acfd-b80b6123c80b.mov

## Acceptance checklist


src/hooks/Transactor.ts
```diff
@@ -181,11 +181,10 @@ export function useTransactor({
           description = JSON.parse(json).message || message
         } catch (_) {
           description = message
-          options?.onError?.(new DOMException(description))
           emitErrorNotification(t`Transaction failed`, { description })
         }
 
-        options?.onDone?.()
+        options?.onError?.(new DOMException(description))
```
wraeth-eth, line 187: I am hesitant to change the transactor code as it could have unforeseen consequences elsewhere. We are using it in too many places. Although its not the best practice, I think you are better off using the pattern used elsewhere of detecting the boolean flag that is returned from the transactor call itself.

i.e, 
```ts
const txSuccess = redeemTokensTx(...)
if (!txSuccess) {
  setTransactionPending(false)
  setLoading(false)
}
```

wraeth-eth (changes requested):

src/hooks/Transactor.ts
wraeth-eth: FWIW this would be less an issue if we had more test coverage, but I don't think it is worth running this risk right now

johnnyd-eth (approved):
Great work fren! Lezgoooo

src/hooks/Transactor.ts
johnnyd-eth: Good point, is this an essential change for this fix @NakedFool ?

NakedFool: @johnnyd-eth It is ser, i will apply right in the morning.

@wraeth-eth that is good point, i was a bit anxious about changing that functionality in there, i will minimize risks with how you explained it to me.

NakedFool:
I've applied the logic that we talked about above, feel free to review @wraeth-eth  or @johnnyd-eth

src/hooks/Transactor.ts
NakedFool: Branch is now updated with these changes 🍻

benschac:
Hey @NakedFool,

This was a great catch.

We have this problem though out the app, I've been working on fixing a bunch of them. Definitely appreciate the 👀, additionally if you wanna help me fix a bunch of these UX bugs I could use the help.


We have a pattern now that's a little more straight forward, and doesn't involve changing too much code:

https://github.com/jbx-protocol/juice-interface/pull/1755/files -- what the pattern should look like moving forward.

Get out of the pending state: https://github.com/jbx-protocol/juice-interface/blob/cancelling-redeemTx-keeps-loading/src/components/Project/ProjectToolsDrawer/TransferOwnershipForm.tsx#L33 all you'd need to do is just set the state on the onError callback fn.

NakedFool:
> Hey @NakedFool,
> 
> This was a great catch.
> 
> We have this problem though out the app, I've been working on fixing a bunch of them. Definitely appreciate the 👀, additionally if you wanna help me fix a bunch of these UX bugs I could use the help.
> 
> We have a pattern now that's a little more straight forward, and doesn't involve changing too much code:
> 
> https://github.com/jbx-protocol/juice-interface/pull/1755/files -- what the pattern should look like moving forward.
> 
> Get out of the pending state: https://github.com/jbx-protocol/juice-interface/blob/cancelling-redeemTx-keeps-loading/src/components/Project/ProjectToolsDrawer/TransferOwnershipForm.tsx#L33 all you'd need to do is just set the state on the onError callback fn.

Yo @benschac 

I wanted to go with a similar approach but @wraeth-eth said above to me that it is not good to change transactor logic for now since it is called in many places which makes sense.

However, since your PR tackle all places in which the transactor is being called, making those changes there makes sense and I'm more than happy to help you out with that. It looks to me like that PR is closed now. in any case I'd like to work with you on this stuff :)

Maybe this PR is in a good state to merge as it is now, then when we are finished with the PR you linked it will replace the current logic.

benschac:
My bad, I sent you the wrong link of the pattern we use today.


This is the pattern that we use today: https://github.com/jbx-protocol/juice-interface/blob/1ee4f8db276c80610652ed919a55a385989672ab/src/hooks/v2/transactor/TransferProjectOwnershipTx.ts


Moving forward, I was going to use a `try catch` block. But, that doesn't really matter for this. The other link I sent yesterday you can ignore.

You don't need to change the transactor code, just the hook that's calling the transactor, like `TransferProjectOwnershipTx`

then add the on error callback in the hook call https://github.com/jbx-protocol/juice-interface/blob/cancelling-redeemTx-keeps-loading/src/components/Project/ProjectToolsDrawer/TransferOwnershipForm.tsx#L33

NakedFool:
> My bad! I sent you the wrong link of the pattern we use today.
> 
> This is the pattern that we use today: https://github.com/jbx-protocol/juice-interface/blob/1ee4f8db276c80610652ed919a55a385989672ab/src/hooks/v2/transactor/TransferProjectOwnershipTx.ts
> 
> Moving forward, I was going to use a `try catch` block. But, that doesn't really matter for this. The other link I sent yesterday you can ignore.
> 
> You don't need to change the transactor code, just the hook that's calling the transactor, like `TransferProjectOwnershipTx`
> 
> then add the on error callback in the hook call https://github.com/jbx-protocol/juice-interface/blob/cancelling-redeemTx-keeps-loading/src/components/Project/ProjectToolsDrawer/TransferOwnershipForm.tsx#L33

Gotcha, yeah I was a bit confused 😸 

I'll apply it the way you are doing then and if you need a hand with other stuff just let me know and I'll be happy to help you 🤙

NakedFool:
@benschac Hmm i think we have one problem here, it is because we are using `TransactionModal` in these files that I'm applying.

Thing is, but you may know this is that we need `transactionPending` to be `true` while it is in `onDone` callback, with the approach above I would need to set `false` in `onDone` in order for the modal to disappear since it is first going into `onError` and then into `onDone` but that's not nice since if we have successful transactions we wouldn't display pending modal then 😿  

wdyt? am I missing something here? I think main difference between examples you provided from files and mine ones are that I'm having a transaction modal here and I need to follow that state.

benschac:
> we need transactionPending to be true while it is in onDone callback
I think that's a bug.

Made an example of what I'm talking about, handling both API and missing parameter errors and pending states.

https://github.com/jbx-protocol/juice-interface/pull/1845/files


What I'm suggesting is following the pattern of callback functions in the hooks to both trigger an error toast, and control pending states, like the claim button pending state in modal.

benschac:
Following up from discord. Your implementation works as expected with the pending modal shown after the `claim` button is  fixed.

I think there's a bug in the transactor where the pending transaction modal is being dismissed.

NakedFool:
> Following up from discord. Your implementation works as expected with the pending modal shown after the `claim` button is fixed.
> 
> I think there's a bug in the transactor where the pending transaction modal is being dismissed.
> 
> Ignore my comments. Sorry about the back and forth.

Responded in discord but will add here too, so it looks like we need to adjust transactor in order to make things flow easily across many components.

I will take a look!

benschac:
Cross post from discord: https://discord.com/channels/939317843059679252/1012735432515129446/1013077786031829082

NakedFool:
> Cross post from discord: https://discord.com/channels/939317843059679252/1012735432515129446/1013077786031829082

@benschac can you review this one more time when you have a chance ser? 🍻

benschac:
I had to quick questions:
- It looks like `public/vendor/desmos/calculator.min.js` was related to a different PR / might have got mixed into this PR accidently. Could you remove it? Additionally, not sure how big of a deal it is cc: @wraeth-eth @johnnyd-eth


- > For testing purposes, I have disabled validation checks so I can test them easily.
I have tested on Burn and Claim modal, Redeem is in the same file as Burn, they are using the same function to execute transactions so, therefore, the same behavior is going to be there.

Could you give Redeem a test too? / Screen Shot?


Also, @wraeth-eth, didn't mean to take over / step on toes. I was looking at the transactor code a bit and thought I could add some context 😅.

src/components/v2/V2Project/V2ManageTokensSection/V2ClaimTokensModal.tsx
```diff
@@ -58,8 +58,17 @@ export default function V2ClaimTokensModal({
           setTransactionPending(false)
           onConfirmed?.()
         },
+        onError: () => {
+          setTransactionPending(false)
+          setLoading(false)
+        },
```
benschac, line 64:
```suggestion
        onError: (e: DOMException) => {
          setTransactionPending(false)
          setLoading(false)
          emitErrorNotification(e.message)
        },
```

src/hooks/v2/transactor/ClaimTokensTx.ts
```diff
@@ -15,6 +15,22 @@ export function useClaimTokensTx(): TransactorInstance<{
...
+        ? 'userAddress'
+        : !projectId
+        ? 'projectId'
+        : !contracts?.JBTokenStore
+        ? 'contracts.JBTokenStore'
+        : null
+
+      txOpts?.onError?.(
+        new DOMException(
+          `Missing ${missingParam ?? 'parameter` not found'} in v2 transactor`,
```
benschac, line 30: I don't think we need 'parameter`' --> 'parameter' remove the `

NakedFool: nice catch!

tomquirk:
@NakedFool checking in on this one. Can I help with anything to get this merged?

NakedFool:
> @NakedFool checking in on this one. Can I help with anything to get this merged?

code-wise, everything is good; we only need to be sure we need this `calculator.min.js` merged with this PR, or we should delete it.

tomquirk:
@NakedFool ah good catch, no, there shouldn't be a diff there. I've removed it and merged in the latest `main`. Will do a final review now and merge 💪

src/hooks/v2/transactor/ClaimTokensTx.ts
```diff
@@ -15,6 +15,22 @@ export function useClaimTokensTx(): TransactorInstance<{
...
+        ? 'projectId'
+        : !contracts?.JBTokenStore
+        ? 'contracts.JBTokenStore'
+        : null
+
+      txOpts?.onError?.(
+        new DOMException(
+          `Missing ${missingParam ?? 'parameter not found'} in v2 transactor`,
+        ),
+      )
```
tomquirk, line 32: @NakedFool one final thing: could you please update these to use the new-ish `onCatch` function?

https://github.com/jbx-protocol/juice-interface/blob/4a593744b502a7c8b3debcf6a844e6a149986a83/src/hooks/Transactor.ts#L32

tomquirk:
@NakedFool just one last comment!

src/hooks/v2/transactor/ClaimTokensTx.ts
NakedFool: sure! 🍻

NakedFool:
> @NakedFool just one last comment!

I've updated the branch with the latest changes, take a look when you have a chance! <3

tomquirk (approved):
@NakedFool fantastic! THanks for making those final changes. Merging 🚀🚀🚀🚀

Issue #1787, opened 2022-08-23
Remove 'funding distribution' title from distribution card when no funds are being distributed


johnnyd-eth:
## Summary

<img width="536" alt="Screen Shot 2022-08-23 at 12 22 26 pm" src="https://user-images.githubusercontent.com/96150256/186054946-99b9c878-a07f-4a12-9a40-3a64706ffe83.png">

## Environment

- **Device/OS**: [e.g. iOS]
- **Browser:** [e.g. chrome, safari]
- **Hardware wallet**: [e.g. Ledger w/ Metamask]

PR #1792, opened 2022-08-23
Removed funding distribution area if there is not distribution amount

NakedFool:

Fixes - https://github.com/jbx-protocol/juice-interface/issues/1787

The intention of this PR is that if there is no distributable amount, the funding distribution area is removed in the funding cycle section.

image
## Acceptance checklist


johnnyd-eth (changes requested):
I think for now we should just remove the lonely yellow header:

Screen Shot 2022-08-24 at 9 17 04 am

johnnyd-eth (approved):
LGTM!

PR #1811, opened 2022-08-24
Make Payment Address copy consistent

tomquirk:

Use `Payment Address` everywhere.


johnnyd-eth (approved):
Gr8 gr8 gr8 LGTM!

Issue #1781, opened 2022-08-22
Load project handles in payout cards


johnnyd-eth:
## Description

This is in `DistributionSplitsSection/index`

<img width="522" alt="Screen Shot 2022-08-22 at 2 02 19 pm" src="https://user-images.githubusercontent.com/96150256/185848226-7c9f4904-0f2c-413b-9411-d3b959c67b50.png">

_A clear and concise description of what could be better._

## Proposal

_A description of the change you'd like to see._

## Screenshots

_If applicable, add screenshots to help explain your problem._

PR #1814, opened 2022-08-24
Load project handles in payout cards v2

NakedFool:

Fixes - https://github.com/jbx-protocol/juice-interface/issues/1781

I've just got one UI / UX dilemma here, I'm utilizing `V2ProjectHandle` shared func in `DistributionSplitCard`

I've renamed the label in the first row, instead of `Project ID` it is now `Project` since sometimes we get the handle back, only small issue I see here is that if the project has no handle then in the first row it is like this `Project: Project 13`  ( since shared func is structured like that but it will be clear to you what I mean in screenshots )

I can either change our shared func to return a boolean if there is a handle so I can manipulate with labels outside of it or I can repeat logic from the shared func inside `DistributionSplitCard` but I think that is not good since we want to utilize shared func as much as we can.
image
image

## Acceptance checklist

### DIFF
--- a/src/components/v2/shared/DistributionSplitsSection/DistributionSplitCard.tsx
+++ b/src/components/v2/shared/DistributionSplitsSection/DistributionSplitCard.tsx
@@ -23,6 +23,7 @@ import {
 } from 'utils/v2/math'
 
 import { CurrencyName } from 'constants/currency'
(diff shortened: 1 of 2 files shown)


johnnyd-eth (approved):
No stress. This looks good to me 🛳

PR #1816, opened 2022-08-24
Add persistent transactions list / deprecate bnc-notify

peripheralist:

Adds a transaction history state and replaces bnc-notify.

Transactions appear in a collapsible list below the navbar and link to the tx on etherscan. History persists even on refresh, using a localStorage entry keyed using the connected wallet + current chain. Transaction state can be "success", "failed", or "pending". Pending transactions refresh their status automatically until they've been mined.

**Flow:**
- Transaction created in the UI (i.e. Pay modal)
- Transaction added to TxHistory state with "pending" status
- Navbar automatically opens the TransactionsList when a pending tx is added
- Pending tx appears in TransactionsList with a loading icon
- Pending tx is periodically refreshed using a poller until it's been mined
- Once mined, pending tx status is updated to "success" or "failed"
- Polling stops once there are no pending transactions in the TxHistory state
- Transactions will persist up to 1 hour in TxHistory using localStorage, but can also be manually removed

**Note:** Most files changed in this PR are just adding titles to TransactorInstances. The most important changes are:
- txHistoryContext.ts
- Transactor.ts
- TxHistoryProvider.tsx
- TransactionsList.tsx

image
image
image
image

Screen recording shared in discord, file too big for here

## Acceptance checklist


src/hooks/Transactor.ts
```diff
@@ -75,34 +70,6 @@ export function useTransactor({
         return false
       }
 
-      const notifyOpts: InitOptions = {
```
peripheralist, line 78: all bnc-notify stuff removed

peripheralist:
updated styling 
image
image

src/hooks/Transactor.ts
```diff
@@ -4,22 +4,21 @@ import { Contract } from '@ethersproject/contracts'
...
-import { windowOpen } from 'utils/windowUtils'
 
+import { TxHistoryContext } from 'contexts/txHistoryContext'
 import { useWallet } from './Wallet'
 
-type TransactorCallback = (e?: TransactionEvent, signer?: Signer) => void
+type TransactorCallback = (e?: Transaction, signer?: Signer) => void
 
 type TransactorOptions = {
+  title?: string
```
peripheralist, line 21: this allows us to define a title for every TransactorInstance (i.e. "Pay <project name>", "Redeem $<token>", etc)

it will default to using the functionName. so far in this PR a title has only been defined for the pay tx

src/providers/TxHistoryProvider.tsx
```diff
@@ -0,0 +1,134 @@
+import { TransactionResponse } from '@ethersproject/providers'
```
peripheralist, line 1: TxHistoryProvider is the crux of this PR

- maintains state of pending/historical transactions. provides `transactions` via context
- uses a poller to periodically update pending TXs
- provides addTransaction/removeTransaction functions via context for manipulating state

src/providers/TxHistoryProvider.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+const nowSeconds = () => Math.round(new Date().valueOf() / 1000)
+
+const KEY_TRANSACTIONS = 'transactions'
+
+const SHORT_TERM_POLL_INTERVAL_MILLIS = 3 * 1000 // 3 sec
+const LONG_TERM_POLL_INTERVAL_MILLIS = 12 * 1000 // 12 sec
+
+// Arbitrary time to give folks a sense of tx history
+const TX_HISTORY_TIME_SECS = 60 * 60 // 1 hr
```
peripheralist, line 17: refreshing the app will discard any txs older than 1 hour from localStorage

src/providers/TxHistoryProvider.tsx
```diff
@@ -0,0 +1,145 @@
...
+              ),
+            ),
+          pollInterval,
+        ),
+      )
+    }
+
+    // Clean up. Ensures we stop polling once all pending TXs have been refreshed
+    return () => {
+      if (poller) clearInterval(poller)
```
peripheralist, line 107: we're basically destroying and _conditionally_ recreating the poller every time the state of any transaction changes. i think it's the best way to do this, but i also went kinda mad dancing with react state while testing this so who knows

src/providers/TxHistoryProvider.tsx
peripheralist: totally arbitrary. idea is to give folks some sense of their history if they've been in a sesh on the jb app. but you prolly won't care about old transactions if it's the next day or whatev

src/hooks/Transactor.ts
```diff
@@ -127,9 +94,13 @@ export function useTransactor({
       try {
         let result
 
+        const txTitle = options?.title ?? functionName
+
         if (tx instanceof Promise) {
           console.info('AWAITING TX', tx)
           result = await tx
+
+          addTransaction?.(txTitle, result)
```
peripheralist, line 103: each time the transactor creates a transaction, we beam it to the TxHistoryProvider via TxHistoryContext

src/hooks/Transactor.ts
```diff
@@ -39,18 +38,14 @@ export type TransactorInstance<T = undefined> = (
...
 export function useTransactor({
   gasPrice,
 }: {
   gasPrice?: BigNumber
 }): Transactor | undefined {
   const { chain, signer } = useWallet()
-  const chainId = chain ? BigNumber.from(chain.id) : undefined
   const { chainUnsupported, isConnected, changeNetworks, connect } = useWallet()
-
-  const { isDarkMode } = useContext(ThemeContext)
```
peripheralist, line 53: Love removing transactor's dependency on theme

src/hooks/v2/transactor/PayETHPaymentTerminal.ts
```diff
@@ -17,7 +17,11 @@ type PayV2ProjectTx = TransactorInstance<{
   value: BigNumber
 }>
 
-export function usePayETHPaymentTerminalTx(): PayV2ProjectTx {
+export function usePayETHPaymentTerminalTx({
```
peripheralist, line 20: This file diff is an example of how we'll add transactionTitles to every TransactorInstance

Titles will default to the contract function name (see transactor hook) if we don't define them

src/components/Navbar/TransactionsList.tsx
```diff
@@ -0,0 +1,172 @@
...
+  )
+
+  useEffect(() => {
+    // Auto expand if pending tx is added
+    if (hasPendingTxs) setIsExpanded(true)
+  }, [hasPendingTxs])
+
+  return (
+    <div style={{ ...style }}>
+      {!!transactions?.length && (
```
jmilldotdev, line 65: why the `!!` usage?

peripheralist: `transactions?.length` doesn't automatically resolve to boolean, i'm guessing due to the undefined but seems like a ts bug. need to do `transactions && transactions.length`, so this is just shorthand

peripheralist: edit—eh nvm i was wrong. i swear i was seeing that before, but looking now it seems like i can remove the `!!`

src/components/Navbar/TransactionsList.tsx
```diff
@@ -0,0 +1,172 @@
...
+                  <a
+                    style={{
+                      cursor: tx.tx.hash ? 'pointer' : 'default',
+                      color: colors.text.primary,
+                      fontSize: '0.7rem',
+                      lineHeight: '1.5rem',
+                    }}
+                    href={
+                      tx.tx.hash
+                        ? `https://${readNetwork.name}${
```
jmilldotdev, line 126: might be nice to kick this out to a utility function, getEtherscanTxLink or something like that

jmilldotdev (dismissed):
High-level thoughts:

- I believe the transactions dropdown should automatically close once all queued transactions are out of pending state
- Do we need the transaction modals to stay open anymore, or can we automatically close them now that we have something on the page indicating a pending state?
- If we do that, it might be nice to have a sticky header, or at least make the transactions indicator sticky as you scroll down the page while having a pending transaction?
- afaik, the transaction indicator doesn't show up at all on mobile width
- once we have an application DB it will probably be nice to store the entire transaction history for each wallet that has connected

src/components/Navbar/TransactionsList.tsx
peripheralist: yes. luv

src/components/Navbar/TransactionsList.tsx
peripheralist: nvm again, just actually remembered why we need this. in tsx && statements, if the first expression is a number `0`, react just renders the 0 instead of evaluating it as boolean. kinda unexpected, but makes sense i guess. keeping the `!!`

image

src/components/Navbar/TransactionsList.tsx
```diff
@@ -0,0 +1,168 @@
...
+  style?: CSSProperties
+  listStyle?: CSSProperties
+}) {
+  const { transactions, removeTransaction } = useContext(TxHistoryContext)
+  const {
+    theme: { colors },
+  } = useContext(ThemeContext)
+  const [isExpanded, setIsExpanded] = useState<boolean>()
+
+  const TxStatusElem = (status: TxStatus) => {
```
wraeth-eth, line 33: Avoid nesting components - it has problems with re-render optimization

src/hooks/Transactor.ts
wraeth-eth: <3

src/hooks/v1/transactor/RedeemTokensTx.ts
```diff
@@ -40,7 +41,10 @@ export function useRedeemTokensTx(): TransactorInstance<{
         userAddress,
         preferConverted,
       ],
-      txOpts,
+      {
+        ...txOpts,
+        title: tokenSymbol ? t`Redeem $${tokenSymbol}` : t`Redeem tokens`,
```
wraeth-eth, line 46: I vaguely remember a util function that can turn token symbols into plural, but I might have misrememebred something else

src/providers/TxHistoryProvider.tsx
wraeth-eth: Could definitely see value in one day adding a "See all transactions" and we just build a list of them based on the connected wallet, but definitely out of scope

wraeth-eth (approved):
pre-approving

Love it Peri, amazing work ❤️

src/hooks/v1/transactor/RedeemTokensTx.ts
peripheralist: found it, ty!

PR #1831, opened 2022-08-26
[5/4] Include owner split in V2 split export

tomquirk:
Includes the project owner's remainder "split" in the V2 splits CSV file.


johnnyd-eth (approved):
Hell yeah

PR #1868, opened 2022-08-29
Image uploader writes value in `ipfs://` format

### LINKED ISSUE #1726: Rich media Memos should use IPFS links instead of Private Gateway links
## Description

At present, JBM's memo image upload passes Pinata gateway links to the `pay` function. We should not be writing this offchain dependency in memos. 

This will be regrettable when Pinata changes their URL or we switch providers -- or for all other interfaces that want to render pay events but lack access to our API key and will need to write custom logic to render via a different gateway.

## Proposal

1. Memo image upload tool should pin to the jbx.mypinata gateway, but use the raw IPFS URI (`ipfs://Qm..`) in the pay function call.
2. JBM should render all IPFS links via the private gateway, so there is no performance loss

## Screenshots

image

johnnyd-eth:

Fixes #1726. 

- Stores uploaded image links in memos in `ipfs://${cid}` format rather than their full links 
- No user-facing changes

**Pay memo:**
Screen Shot 2022-08-29 at 4 22 59 pm

Screen Shot 2022-08-29 at 4 24 32 pm

**Project payer:**
Screen Shot 2022-08-29 at 4 25 50 pm

Screen Shot 2022-08-29 at 4 29 08 pm

## Acceptance checklist


src/components/inputs/FormImageUploader.tsx
```diff
@@ -15,27 +15,30 @@ enum ByteUnit {
 // TODO: This is a double up of `ImageUploader`. We should combine the two.
 
 export const FormImageUploader = ({
-  value,
```
johnnyd-eth, line 18: Unused code

src/utils/ipfs.ts
```diff
@@ -174,7 +174,7 @@ export async function uploadNftRewardsToIPFS(
 export function formatIpfsLink(ipfsLink: string) {
   const ipfsLinkParts = ipfsLink.split('/')
   const cid = ipfsLinkParts[ipfsLinkParts.length - 1]
-  return `https://${DEFAULT_PINATA_GATEWAY}/ipfs/${cid}`
+  return `https://${IPFS_GATEWAY_HOSTNAME}/ipfs/${cid}`
```
johnnyd-eth, line 177: No reason for this to use pleb gateway. Might result in the newly uploaded image not showing up. Switched to the private gateway :)

src/components/inputs/FormImageUploader.tsx
```diff
@@ -15,27 +15,30 @@ enum ByteUnit {
...
+  const [imageCid, setImageCid] = useState<string>()
 
   const { theme } = useContext(ThemeContext)
 
   const setValue = (cid?: string) => {
-    const newUrl = cid ? ipfsCidUrl(cid) : undefined
-    onChange?.(newUrl)
+    setImageCid(cid)
+    // storing images in `ipfs://` format where possible (see issue #1726)
+    const ipfsUrl = cid ? `ipfs://${cid}` : undefined
```
johnnyd-eth, line 36: Sets form value to `ipfs://` format

src/components/inputs/FormImageUploader.tsx
```diff
@@ -15,27 +15,30 @@ enum ByteUnit {
...
+    onChange?.(ipfsUrl)
   }
 
+  const imageUrl = imageCid ? ipfsCidUrl(imageCid) : undefined
```
johnnyd-eth, line 40: Still uses and shows full URL within this component

wraeth-eth (changes requested):
Have requested some changes, but overall this PR looks great! Really great idea, thanks for picking this up ❤️ @johnnyd-eth !!!

src/utils/ipfs.ts
tomquirk: If someone inputs their own image URL string, wont this break?

wraeth-eth: Do you mean directly into the memo? If so, it should be fine as this is only touching the upload image input code

src/components/inputs/FormImageUploader.tsx
johnnyd-eth: Great pickup @wraeth-eth. I've reinstated `value` and incorporated it with my changes as necessary. The issue you attached is fixed:

https://user-images.githubusercontent.com/96150256/187572017-595b4ba8-4dcf-4891-84e9-c7aa693b5e45.mp4

src/utils/ipfs.ts
johnnyd-eth: I see what you're saying @aeolianeth . I've reverted this change

johnnyd-eth:
@wraeth-eth addressed your latest ser thank you!

wraeth-eth (approved):
LFG

PR #1869, opened 2022-08-29
Make contractToRead shared code [1/n]

wraeth-eth:
## What does this PR do and why?

Initial work to converge the v1 and v2 contract readers (as they share a lot of code).
This also begins the work of create a hook specifically for the contract read value, and allowing the ability to update the read value in a simple way (coming later in stack).

this code can be shared amongst these two files

## Screenshots or screen recordings

_If applicable, provide screenshots or screen recordings to demonstrate the changes._

## Acceptance checklist

- [ ] I have evaluated the [Approval Guidelines](https://github.com/jbx-protocol/juice-interface/blob/main/CONTRIBUTING.md#approval-guidelines) for this PR.
- [ ] I have tested this PR in [all supported browsers](https://github.com/jbx-protocol/juice-interface/blob/main/CONTRIBUTING.md#supported-browsers).
- [ ] I have tested this PR in dark mode and light mode (if applicable).

PR #1870, opened 2022-08-29
Add hook to supply contract read value hook [2/n]

wraeth-eth:

Add a hook to allow flexible updating of a contract read value.

This code will allow the combining of code from v1 and v2 read contracts.

This code is mostly taken from v1 and v2 ContractReader's and supplied now in a shared way. See next item in stack for application.

The rationale here is that `useContractReadValue` can be used anywhere a contract needs to be read. It will also allow the ability to poll contract read values via the `refetchValue` method returned from the hook.

## Acceptance checklist

### DIFF
--- a/src/hooks/ContractReadValue.ts
+++ b/src/hooks/ContractReadValue.ts
@@ -1,0 +1,97 @@
+import { Contract } from '@ethersproject/contracts'
+import { useCallback, useState } from 'react'
+import useDeepCompareEffect from 'use-deep-compare-effect'
+import { callContractRead } from 'utils/callContractRead'
+import { contractToRead } from 'utils/contractToRead'
+
+/**
+ * Reads a `contract` value from `contracts`, using the `functionName`.
+ *
+ * Upon load, the contract is initially read. From there, `refetchValue` must be
+ * called.
+ *
+ * @returns refetchValue - Call to refetch and hydrate the `value`.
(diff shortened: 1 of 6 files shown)


wraeth-eth:
* main
  * **PR #1869** <a href="https://app.graphite.dev/github/pr/jbx-protocol/juice-interface/1869" target="_blank">Graphite</a> 
    * **PR #1870** <a href="https://app.graphite.dev/github/pr/jbx-protocol/juice-interface/1870" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jbx-protocol/juice-interface/1870?utm_source=stack-comment).

src/hooks/ContractReadValue.ts
```diff
@@ -0,0 +1,97 @@
...
+export function useContractReadValue<C extends string, V>({
+  contract,
+  contracts,
+  functionName,
+  args,
+  version,
+  formatter,
+  valueDidChange,
+}: {
+  version: string
```
peripheralist, line 26: i wonder if we really need this `version` arg, since we're only using it for logging. there may also be version-agnostic reader instances (i.e. price of ETH, or generic ERC20 calls)

wraeth-eth: Yeah, agree. I will work up in the stack to converge the two into a single shared function

Issue #1872, opened 2022-08-30
Use DM Mono in Web3-onboard


tomquirk:
We're loading a whole new font for web3-onboard (Sofia Pro). This decreases performance, and also makes the UI less cohesive.

Lets use DM Mono (same font as the rest of the site) for now

<img width="805" alt="Screen Shot 2022-08-30 at 9 46 34 AM" src="https://user-images.githubusercontent.com/12551741/187320945-f67e1486-d135-4450-96e1-d64836015767.png">

wraeth-eth:
I opted away from DM Mono as it doesn't look very good on this form 😅

tomquirk:
> I opted away from DM Mono as it doesn't look very good on this form 😅

https://github.com/jbx-protocol/juice-interface/pull/1874

PR #1874, opened 2022-08-30
Some font optimisations

tomquirk:
- Remove DM Mono weight 300 font. We don't use this anywhere.
- Remove DM Mono italic variant. We used this in 2 spots, but I think performance cost outweighs the stylistic benefit.
- Use DM Mono in web3-onboard, for both performance and vibes. Closes https://github.com/jbx-protocol/juice-interface/issues/1872

Screen Shot 2022-08-30 at 10 19 27 AM

### DIFF
--- a/src/components/common/Head/Head.tsx
+++ b/src/components/common/Head/Head.tsx
@@ -15,9 +15,11 @@ export const Head: VFC = () => {
       <link rel="apple-touch-icon" href="/assets/juice_logo-ol.png" />
       <link rel="icon" href="/favicon.ico" />
       <link rel="manifest" href="/manifest.json" />
-      <link rel="preconnect" href="https://fonts.gstatic.com" />
+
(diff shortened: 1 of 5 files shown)


src/components/common/Head/Head.tsx
```diff
@@ -15,21 +15,26 @@ export const Head: VFC = () => {
...
         async
         src="https://www.desmos.com/api/v1.6/calculator.js?apiKey=dcb31709b452b1cf9dc26972add0fda6"
       ></script>
-      <script
-        src="https://learned-hearty.juicebox.money/script.js"
-        data-site="ERYRRJSV"
-        defer
-      ></script>
+
+      {process.env.NODE_ENV === 'production' && (
```
wraeth-eth, line 31: :pog:

wraeth-eth (approved):

Issue #1733, opened 2022-08-14
Owner percentage doesn't show on infinite distribution 


johnnyd-eth:
## Summary

<img width="516" alt="Screen Shot 2022-08-14 at 11 18 30 pm" src="https://user-images.githubusercontent.com/96150256/184538932-e9f5d702-edcc-415b-9f0d-3fc4de9acd84.png">

This shouldn't be blank. Should show 'Owner 100%' like reserved tokens

## Environment

- **Device/OS**: [e.g. iOS]
- **Browser:** [e.g. chrome, safari]
- **Hardware wallet**: [e.g. Ledger w/ Metamask]

NakedFool:
> 

do you happen to remember which projects are affected by this? some projects I just checked now with infinite dist had owner percentage

NakedFool:
@johnnyd-eth Is this now connected to this fix? https://github.com/jbx-protocol/juice-interface/pull/1806

NakedFool:
Ignore above, just found that this project has that issue  - https://juicebox.money/@dcfcdao
Investigating issue.

PR #1883, opened 2022-08-30
Display owner percentage under funding distribution if there are no splits fix

NakedFool:
Fixes - https://github.com/jbx-protocol/juice-interface/issues/1733
Example - https://juicebox.money/@dcfcdao

Issue is happening when there are no splits, therefore we do not show owner split since we do not pass `reservedRate`

I've added logic to pass `reservedRate` through `PayoutSplitsCard.tsx` since it is not passed to `SplitList.tsx`
Let me know if `reservedRate` is not passed for a reason to `SplitList.tsx` through `PayoutSplitsCard.tsx` 

EDIT: I've updated `UpcomingFundingCycle` with `reservedRate` as well.
**Before**
image
**After**
image

## Acceptance checklist

### DIFF
--- a/src/components/v2/shared/SplitList.tsx
+++ b/src/components/v2/shared/SplitList.tsx
@@ -48,7 +48,7 @@ export default function SplitList({
             />
           </div>
         ))}
-      {ownerSplit ? (
+      {ownerSplit && reservedRate ? (
         <SplitItem
           split={ownerSplit}
           showSplitValue={showSplitValues}
--- a/src/components/v2/shared/SplitList.tsx
+++ b/src/components/v2/shared/SplitList.tsx
@@ -48,7 +48,7 @@ export default function SplitList({
             />
           </div>
         ))}
-      {ownerSplit && reservedRate ? (
+      {ownerSplit ? (
         <SplitItem
           split={ownerSplit}
           showSplitValue={showSplitValues}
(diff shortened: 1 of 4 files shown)


src/components/v2/shared/SplitList.tsx
```diff
@@ -48,7 +48,7 @@ export default function SplitList({
             />
           </div>
         ))}
-      {ownerSplit && reservedRate ? (
+      {ownerSplit && ownerSplit.percent && reservedRate ? (
```
johnnyd-eth, line 51: This can be `ownerSplit?.percent`

johnnyd-eth: I think the better approach to this whole PR is just removing `reservedRate` from this line. Now that we're checking for having no distributions and reserved rate in the `PayoutSplitsCard` and `ReservedTokensSplitsCard` respectively, it's not necessary to have this check here. 

No need to be passing `reservedRate` into the payouts `SplitsList`

johnnyd-eth (changes requested):
.

src/components/v2/shared/SplitList.tsx
NakedFool: You are right, my first commit in PR was removing `reservedRate` but i missed part that there is a check for reserved tokens in `ReservedTokensSplitsCard` and thought that needed to stay 😕 

Things should be good to merge now 👍

johnnyd-eth (approved):
LGTM!

PR #1887, opened 2022-08-31
Only import Desmos when we need it

tomquirk:

- Move all `script` tags to own components
- Remove Desmos from the default head
- Add Desmos script to v1 + v2 create, and v1 + v2 project pages

### DIFF
--- a/src/components/common/SEO/SEO.tsx
+++ b/src/components/common/SEO/SEO.tsx
@@ -46,10 +46,6 @@ export const SEO: FC<Props> = ({
           name="googlebot"
           content={robots ?? 'index,follow'}
         ></meta>
-        <meta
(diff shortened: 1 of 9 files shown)


src/components/common/SEO/SEO.tsx
```diff
@@ -46,10 +46,6 @@ export const SEO: FC<Props> = ({
           name="googlebot"
           content={robots ?? 'index,follow'}
         ></meta>
-        <meta
```
tomquirk, line 49: in DNS now, don't need it anymore

tomquirk:
Final one for you @wraeth-eth <3

wraeth-eth (approved):

wraeth-eth (approved):

PR #1892, opened 2022-09-01
Add window jb object

tomquirk:
add feature flag utils to window object, for easier use.

Example:
Screen Shot 2022-09-01 at 10 22 31 AM


johnnyd-eth (approved):
Nice, all working well for me. Great idea and implementation 💪

PR #1896, opened 2022-09-01
More settings page prep

tomquirk:


src/utils/routes.ts
```diff
@@ -1,4 +1,6 @@
 import { BigNumber, BigNumberish } from '@ethersproject/bignumber'
+import { V2SettingsKey } from 'components/v2/V2Project/V2ProjectSettings/V2ProjectSettings'
+import { NextRouter } from 'next/router'
```
tomquirk, line 4: This diff is to support subroutes (/@handle/settings)

Issue #1909, opened 2022-09-02
AMM Price in tooltip breaks UI when price is too big


NakedFool:
## Summary

When the AMM price of token compared to 1 ETH is too big, tooltip UI breaks

## Steps to reproduce

1. Go to https://www.juicebox.money/@dcfcdao
2. Hover over ` buy BARK on exchange` under `Pay` section
3. See UI behavior
## Relevant logs and/or screenshots

![image](https://user-images.githubusercontent.com/18723426/188125845-b3344d92-7d4d-444f-b174-e1affdbaebd0.png)

## Environment

- **Device/OS**: [e.g. iOS]
- **Browser:** [e.g. chrome, safari]
- **Hardware wallet**: [e.g. Ledger w/ Metamask]

PR #1910, opened 2022-09-02
Added logic to truncate AMM price if it is too big compared to 1 ETH

NakedFool:
Fixes - https://github.com/jbx-protocol/juice-interface/issues/1909

Quick summary of thoughts
- I've initially wanted to expand box, but displaying number this big doesn't really make sense to me 💭 
- I've also wanted to use `text-overflow: ellipsis` but the problem there is that it would put ellipsis on our whole string. 
   for example `190,000,000,000 BARK/ 1 ETH` would become `190,000,000 ...` without user seeing that it is price compared to `1 ETH`
- Created a small function that does what i want, you pass `string` and desired `length` of a string and it will return a truncated string with an ellipsis at the end.

**Before**
image
**After**
image

## Acceptance checklist


src/components/AMMPrices/TokenAMMPriceRow.tsx
```diff
@@ -82,7 +88,7 @@ export default function TokenAMMPriceRow({
             overlayInnerStyle={{ ...fontStyle }}
           >
             <ExternalLink href={exchangeLink} style={{ fontWeight: 400 }}>
-              {`${formattedNum(WETHPrice)} ${tokenSymbol}/1 ETH`}
+              {`${preparePrice(WETHPrice)} ${tokenSymbol}/1 ETH`}
```
johnnyd-eth, line 91: **Nit:** think we could just name this `formatPrice` for consistency

src/components/AMMPrices/TokenAMMPriceRow.tsx
```diff
@@ -56,6 +57,12 @@ export default function TokenAMMPriceRow({
     )
   }
 
+  const preparePrice = (price: string) => {
```
johnnyd-eth, line 60: .

src/utils/formatString.ts
```diff
@@ -0,0 +1,3 @@
+export const truncateString = (str: string, length = 50) => {
```
johnnyd-eth, line 1: **Praise**

johnnyd-eth:
Great find and pickup here @NakedFool. As you mentioned, I would prefer if we still showed an accurate value here rather than ellipses. WDYT about making a `scientificNotation` function so we would show `1.966424e8` (to appropriate preciesion) or something rather than just cutting it off?

johnnyd-eth (changes requested):
.

src/components/AMMPrices/TokenAMMPriceRow.tsx
NakedFool: agreed!

NakedFool:
> Great find and pickup here @NakedFool. As you mentioned, I would prefer if we still showed an accurate value here rather than ellipses. WDYT about making a `scientificNotation` function so we would show `1.966424e8` (to appropriate precision) or something rather than just cutting it off?

Agreed, yeah it feels like this can be more tightened, i'm gonna work on it! <3

johnnyd-eth:
Sounds good! Let me know when this is ready for another look @NakedFool

NakedFool:
> Sounds good! Let me know when this is ready for another look @NakedFool

heyo! I've updated branch with the logic you suggested, lemme quick explain what is going on now there.

As you said we want to display correct number, so therefore i saw that in `utils` there is already function named `truncateLongNumber`

But we do not want to always truncate number as in most cases it won't be that long number, therefore i created utility func `formatOrTruncate` that is either formatting or truncating number based on `limit` that is passed as arg in that func.

I've set limit to be trillion, as numbers above are too big for box and numbers below it fits well, let me know what you think of this approach! 🍻

src/components/AMMPrices/TokenAMMPriceRow.tsx
```diff
@@ -56,6 +56,14 @@ export default function TokenAMMPriceRow({
     )
   }
 
+  const formatPrice = (price: string) => {
+    const p = parseInt(price, 10)
+    const formatLimit = 1e12 - 1 // trillion
```
johnnyd-eth, line 61: **Suggestion: ** add `const` for trillion in `constants/numbers` and use it here :)

johnnyd-eth:
Screen Shot 2022-09-07 at 10 47 14 am

Not sure what's going on with `truncateLongNumber` but this still isn't what we want to show. No idea what the value pictured means. 

I think, since it's potentially so big, we need to show scientific notation when it's above a certain amount. So `1.9735439e12` (however many decimals you think look reasonable).

src/components/AMMPrices/TokenAMMPriceRow.tsx
NakedFool: Thank you for this suggestion ser, i felt a bit dirty leaving it here :/

NakedFool:
> Screen Shot 2022-09-07 at 10 47 14 am
> 
> Not sure what's going on with `truncateLongNumber` but this still isn't what we want to show. Not sure what the value pictured means.
> 
> I think, since it's potentially so big, we need to show scientific notation when it's above a certain amount. So `1.9735439e12` (however many decimals you think look reasonable).

You are right, i've used `toExponential` instead, lemme know what do you think about latest approach :)

johnnyd-eth:
Screen Shot 2022-09-08 at 1 42 43 pm

Sweet @NakedFool that LGTM 🥇

johnnyd-eth (approved):

PR #1940, opened 2022-09-09
Distribution limit bug fix

tomquirk:
- Fixes bug where `parseFloat` on a number string that includes commas is parsed incorrectly.
- Pass unformatted fee into `AmountFormItem`
- Moves v1-specific math utils to `utils/v1/math.ts`
- minor copy change

### DIFF
--- a/src/pages/create/forms/FundingForm/DistributionSplitsSection/AmountFormItem.tsx
+++ b/src/pages/create/forms/FundingForm/DistributionSplitsSection/AmountFormItem.tsx
@@ -1,3 +1,4 @@
+import { BigNumber } from '@ethersproject/bignumber'
 import { t, Trans } from '@lingui/macro'
 import { Form, FormInstance } from 'antd'
 import CurrencySwitch from 'components/CurrencySwitch'
@@ -8,15 +9,16 @@ import TooltipIcon from 'components/TooltipIcon'
 import TooltipLabel from 'components/TooltipLabel'
 import { CurrencyName } from 'constants/currency'
 import { ThemeContext } from 'contexts/themeContext'
-import { round } from 'lodash'
 import { useContext } from 'react'
+import { formatWad, parseWad, stripCommas } from 'utils/formatNumber'
+import { amountSubFee, formatFee } from 'utils/v2/math'
 import { AddOrEditSplitFormFields, SplitType } from './types'
 import { percentageValidator } from './util'
 
 export function AmountFormItem({
   form,
   distributionLimit,
-  feePercentage,
+  fee,
   editingSplitType,
   currencyName,
   isFirstSplit,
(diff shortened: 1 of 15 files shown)


src/pages/create/forms/FundingForm/DistributionSplitsSection/DistributionSplitModal.tsx
```diff
@@ -217,16 +213,22 @@ export function DistributionSplitModal({
     onClose()
   }
 
+  /**
+   * Set new distribution limit
+   */
   useEffect(() => {
     if (distributionLimitIsInfinite || !amount) return
+
+    const newAmount = parseFloat(stripCommas(amount))
```
tomquirk, line 222: This line is the fix.

I'm pretty concerned that this is happening in more places than this, because `FormattedNumberInput` emits the raw string (which includes the commas). I think it should be emitting a number (after all it's a 'number input) or at very least a number string wihtout commas.

src/utils/v1/math.ts
```diff
@@ -0,0 +1,62 @@
+import { BigNumber } from '@ethersproject/bignumber'
```
tomquirk, line 1: Just moving some functions around. Unrelated

src/pages/create/forms/FundingForm/DistributionSplitsSection/AmountFormItem.tsx
```diff
@@ -36,18 +38,19 @@ export function AmountFormItem({
...
-    ? parseFloat(amount) -
-      (parseFloat(amount) * parseFloat(feePercentage ?? '0')) / 100
-    : undefined
-
   function AfterFeeMessage() {
-    return amountSubFee && amountSubFee > 0 ? (
+    if (!fee || !amount || amount === '0') return null
+
+    const feePercentage = formatFee(fee)
+    const amountSubFeeValue = amountSubFee(parseWad(stripCommas(amount)), fee)
```
tomquirk, line 45: Use `amountSubFee` fn.

src/pages/create/forms/FundingForm/DistributionSplitsSection/AmountFormItem.tsx
```diff
@@ -8,23 +9,24 @@ import TooltipIcon from 'components/TooltipIcon'
...
+  fee,
   editingSplitType,
   currencyName,
   isFirstSplit,
   onCurrencyChange,
 }: {
   form: FormInstance<AddOrEditSplitFormFields>
   distributionLimit?: string
-  feePercentage: string | undefined
+  fee: BigNumber | undefined
```
tomquirk, line 29: Parse in the unformatted fee instead.

tomquirk:
@wraeth-eth for vis. Merging as p1

Issue #1926, opened 2022-09-06
Make dates consistent across the app


tomquirk:
## Description

There's some inconsistency with how we display dates and times in the app.

## Proposal

- [ ] Show UTC time on hover on any date/time value
- [ ] Include the timezone for any date/time value

## Screenshots

| element | description | screenshot |
| --- | --- | --- |
| funding cycle time remaining | should have UTC time for FC end date on hover | <img width="502" alt="Screen Shot 2022-09-06 at 10 45 52 AM" src="https://user-images.githubusercontent.com/12551741/188527126-98b26555-5f86-4d0e-954d-48c63b7bb587.png"> |
| start/end FC | should have UTC time on hover. Should include TZ suffix. | <img width="461" alt="Screen Shot 2022-09-06 at 10 45 46 AM" src="https://user-images.githubusercontent.com/12551741/188527207-fb0cf93e-2231-4dbe-a303-ecf57c101fee.png"> |

PR #1942, opened 2022-09-09
Improve date consistency across FC

NakedFool:

Fixes - https://github.com/jbx-protocol/juice-interface/issues/1926

The intention of this PR is to improve date consistency across the funding cycle section

I've just got one question for you @tomquirk 
image

What is the expected situation here since in `start` and `end` we already present the date in UTC format, should we on hover also present the same format or vice versa? I understand suffix addition but couldn't understand hover addition 🍻 

https://user-images.githubusercontent.com/18723426/189296437-b5e4c45c-ec15-43fe-9a3c-7b5abd5a6b20.mov

## Acceptance checklist


tomquirk:
@NakedFool I think `start`/`end` are displayed in the user's local time. In that screenshot I posted, 8:12am is my local time (i'm UTC +9:30). 

I think we have two options:
1. 2022-09-05 8:12am +9:30
2. 2022-09-04 10:42pm UTC

I like the local time. But for consistency, perhaps we should go with the second option and display UTC time.

What do you think?

NakedFool:
> @NakedFool I think `start`/`end` are displayed in the user's local time. In that screenshot I posted, 8:12am is my local time (i'm UTC +9:30).
> 
> I think we have two options:
> 
> 1. 2022-09-05 8:12am +9:30
> 2. 2022-09-04 10:42pm UTC
> 
> I like the local time. But for consistency, perhaps we should go with the second option and display UTC time.
> 
> What do you think?

Ooh, you are right, my bad.

I'd go with option 2, however, we should only show the UTC date on hover in `start` and `end` right?

I'm attaching a recording of the current workaround just so you don't have to pull the latest changes, lmk what you think 

https://user-images.githubusercontent.com/18723426/189839182-d13bd08c-b276-4f1a-a2f3-b7da962d6a31.mov

NakedFool:
@tomquirk the only issue i'm seeing now is UI based, as you can see on the screenshot there isn't enough space in `start` so `UTC` goes in the next row. wdyt?  

image

tomquirk:
@NakedFool I'm pushing a merge commit! 🙏

src/components/v1/shared/FundingCycle/FundingCycleDetails.tsx
```diff
@@ -156,13 +156,23 @@ export default function FundingCycleDetails({
 
         {fundingCycle.duration.gt(0) && (
           <Descriptions.Item label={<Trans>Start</Trans>}>
-            {formattedStartTime}
+            <Tooltip title={formatDateToUTC(fundingCycle.start.mul(1000))}>
```
tomquirk, line 159: **suggestion (non-blocking)**: maybe we can create a `secondsToMilliseconds` utility for all these in a follow-up.

src/utils/format/formatDate.tsx
```diff
@@ -10,9 +10,14 @@ export const formatDate = (
   format = 'YYYY-MM-DD h:mma',
 ) => moment(BigNumber.from(dateMillis).toNumber()).format(format)
 
+export const formatDateToUTC = (
```
tomquirk, line 13: **praise** nice one!

src/utils/format/formatDate.tsx
```diff
@@ -10,9 +10,14 @@ export const formatDate = (
...
+  dateMillis: BigNumberish,
```
tomquirk, line 14: **praise** thanks for including the unit in the variable name.

tomquirk (approved):
@NakedFool thanks so much for this, all looks great. Merging!

src/components/v1/shared/FundingCycle/FundingCycleDetails.tsx
```diff
@@ -1,14 +1,14 @@
 import { parseEther } from '@ethersproject/units'
 import { Trans } from '@lingui/macro'
-import { Descriptions } from 'antd'
+import { Descriptions, Tooltip } from 'antd'
```
tomquirk, line 3: **praise** thanks also for making the changes to V1!!!

Issue #1912, opened 2022-09-03
Front Page payment feed doesn't show v2 project handles


jmilldotdev:
## Summary

<img width="590" alt="image" src="https://user-images.githubusercontent.com/33093632/188281456-17987e79-6a1a-4965-b774-cd0a1d0f3267.png">

## Environment

- **Device/OS**: [e.g. iOS]
- **Browser:** [e.g. chrome, safari]
- **Hardware wallet**: [e.g. Ledger w/ Metamask]

NakedFool:
Very odd issue, correct `projectId` is being passed to `useProjectHandle` but every handle that is returned is `undefined`

And for some reason `readContract` in `callContractRead` function is `undefined` as well.

I'm looking into this.

Issue #1939, opened 2022-09-09
Activity feed not using project handle 


tomquirk:
JuiceboxDAO has a handle (`@juicebox`) but we're rendering `Project #1` in the feed. Should be `@juicebox`

<img width="617" alt="Screen Shot 2022-09-09 at 9 30 34 AM" src="https://user-images.githubusercontent.com/12551741/189245843-54e0029d-2d38-4cf1-9413-d2b4fc3dae3c.png">

NakedFool:
Duplicado ser? https://github.com/jbx-protocol/juice-interface/issues/1912

I was working on this but then switched to https://github.com/jbx-protocol/juice-interface/issues/1926 to clear my head in the meantime

jmilldotdev:
Whatever the solution is to #1912 would fix this also i assume

tomquirk:
Thanks @NakedFool yep, totally a duplicate of https://github.com/jbx-protocol/juice-interface/issues/1912

Closing!

PR #1954, opened 2022-09-11
Show v2 handles in payment feed

NakedFool:

Fixes https://github.com/jbx-protocol/juice-interface/issues/1912 https://github.com/jbx-protocol/juice-interface/issues/1939

The issue here was that in `useV2ContractReader` there was no context from `V2UserContext` returning back from `Payments` component,  therefore we need to wrap the parent component with `V2UserProvider`

I wrapped `TrendingSection` with `V2UserProvider` but I'm not sure if it is the right place to wrap.

**Before**
image

**After**
image

## Acceptance checklist

### DIFF
--- a/src/components/v1/shared/V1ProjectHandle.tsx
+++ b/src/components/v1/shared/V1ProjectHandle.tsx
@@ -5,20 +5,23 @@ import { CSSProperties } from 'react'
 
 export default function V1ProjectHandle({
   projectId,
+  handle,
   style,
 }: {
   projectId: BigNumberish
+  handle?: string | null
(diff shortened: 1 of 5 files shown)


tomquirk:
@NakedFool ahh awesome catch!

This got me thinking... is the project handle included in the subgraph response?

Turns out, yes it is! So fortunately we can avoid this contract call altogether and use the handle from the subgraph instead ☺️ . The 4 most recent commits implement that <3

Going to go ahead and merge. Thanks for jumping on this issue 🙏

tomquirk (approved):

PR #1955, opened 2022-09-11
Add Token to Metamask button

### LINKED ISSUE #1118: "Add token to wallet" button
## Problem to solve

Token holders have to manually add a project's ERC-20 to their wallet (e.g. metamask).

## Proposal

We can add a button to the project page to do this.

## More details

https://docs.metamask.io/guide/registering-your-token.html#example

```js
const tokenAddress = '0xd00981105e61274c8a5cd5a88fe7e037d935b513';
const tokenSymbol = 'TUT';
const tokenDecimals = 18;
const tokenImage = 'http://placekitten.com/200/300';

try {
  // wasAdded is a boolean. Like any RPC method, an error may be thrown.
  const wasAdded = await ethereum.request({
    method: 'wallet_watchAsset',
    params: {
      type: 'ERC20', // Initially only supports ERC20, but eventually more!
      options: {
        address: tokenAddress, // The address that the token is at.
        symbol: tokenSymbol, // A ticker symbol or shorthand, up to 5 chars.
        decimals: tokenDecimals, // The number of decimals in the token
        image: tokenImage, // A string url of the token logo
      },
    },
  });

  if (wasAdded) {
    console.log('Thanks for your interest!');
  } else {
    console.log('Your loss!');
  }
} catch (error) {
  console.log(error);
}
```

jmilldotdev:

Closes #1118. Adds a button in the V2 Manage Tokens section which adds a symbol to the connected wallet.

Currently only supports adding token to Metamask. Had to add the `@metamask/providers` package and make a global declaration of `window.ethereum`. Currently the flow is that we both check that Metamask is installed and that it is connected to the site before showing the button.

Also, as far as I know, there is no way to check whether a user already has a token registered in their Metamask wallet, so there's no nice way to disable the button if they've already registered it.

Also wondering about the placement of the button, and if there's somewhere less intrusive it could go?

image

image

## Acceptance checklist


tomquirk:
@jmilldotdev this looks gr8.

For the UX, what about a real subtle (+) next to the token symbol instead?

tomquirk:
@jmilldotdev two options, I think I prefer the first one but WDYT?

Screen Shot 2022-09-13 at 4 26 34 PM
Screen Shot 2022-09-13 at 4 26 29 PM

jmilldotdev:
second is definitely less intrusive but not at all obvious what it's going to do

tomquirk:
@jmilldotdev Hmm yeah. It kinda doesn't look like a button either. Wdyt about option 1 for now? Can always change it 🚀.

tomquirk:
Idk if option 1 even needs the plus icon tbh.

PR #1959, opened 2022-09-12
[1/2] Only load Prices contract in useEtherPrice

tomquirk:

This PR is in prep for removing V1UserProvider from the global app wrapper.

We only need the `Prices` contract for this single hook. So instead of loading it in the user context, we can load it only for this specific hook. This will allow us to remove V1UserProvider from the global app context and only use it when we need it (follow-up PR)

## Acceptance checklist

- [-] I have tested this PR in dark mode and light mode (if applicable).

### DIFF
--- a/src/models/v2/contracts.ts
+++ b/src/models/v2/contracts.ts
@@ -6,7 +6,6 @@ export enum V2ContractName {
   JBETHPaymentTerminal = 'JBETHPaymentTerminal',
   JBFundingCycleStore = 'JBFundingCycleStore',
   JBOperatorStore = 'JBOperatorStore',
-  JBPrices = 'JBPrices',
   JBProjects = 'JBProjects',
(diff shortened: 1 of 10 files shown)


src/models/v2/contracts.ts
```diff
@@ -6,7 +6,6 @@ export enum V2ContractName {
   JBETHPaymentTerminal = 'JBETHPaymentTerminal',
   JBFundingCycleStore = 'JBFundingCycleStore',
   JBOperatorStore = 'JBOperatorStore',
-  JBPrices = 'JBPrices',
```
tomquirk, line 9: We don't use JBPrices, and are thus loading in its ABIs unnecessarily. Removed.

tomquirk:
@wraeth-eth friendly bump on these 🙏

wraeth-eth (approved):
Bullish

PR #1965, opened 2022-09-13
Safe part 1

tomquirk:

- Loads queued transactions from Gnosis and links to them 
- Provides framework for adding expanding sections for the previews of specific transactions
- Test at mainnet: `v2/p/233/safe`

https://user-images.githubusercontent.com/96150256/190039705-cef0bb8a-527b-4f3c-83a4-136776a377b0.mp4

## Acceptance checklist


src/components/v2/V2Project/V2ProjectSafeDashboard/SafeTransaction.tsx
```diff
@@ -0,0 +1,133 @@
...
+const ReconfigureFundingCyclesOfTransaction = ({
+  transaction,
+}: SafeTransactionComponentProps) => {
+  // const [expanded, setExpanded] = useState<boolean>(false)
+
+  const { contracts } = useContext(V2UserContext)
+
+  if (!contracts || !transaction.data) return null
+
+  // const decodedData = contracts['JBController'].interface.parseTransaction({
```
johnnyd-eth, line 54: Will leave this here for upcoming PR

johnnyd-eth:
@tomquirk this is ready for a look! The only issue I'm seeing in the test case is that the Safe API seems to be sending back some already-rejected tx's, even though we're passing `queued: true` to the request. This seems to be an error on their part. 

I dug a bit but a solution wasn't obvious, so I think we chase the up in a follow-up.

src/components/v2/V2Project/V2ProjectSafeDashboard/SafeTransaction.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+export function SafeTransaction({
+  transaction,
+}: {
+  transaction: SafeTransactionType
+}) {
+  const {
+    theme: { colors },
+  } = useContext(ThemeContext)
+  const [hover, setHover] = useState<boolean>(false)
```
tomquirk, line 98: **suggestion** shouldn't ever set hover state with JS.

I like what you've done with the styling though. It looks like we already have a class for this anyway 💪 

`.hover-bg-l2`

src/components/v2/V2Project/V2ProjectSafeDashboard/SafeTransaction.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+  const TransactionContent = useMemo(() => {
+    if (!method) return GenericSafeTransaction // only the header
+    return TRANSACTION_METHOD_COMPONENTS_MAP[method] ?? GenericSafeTransaction
+  }, [method])
+
+  if (!method) return null
+
+  return (
+    <ExternalLink
```
tomquirk, line 123: Oh nice, i like it 👍

src/styles/utilities.scss
```diff
@@ -30,10 +30,18 @@
...
 
+.hover-box-shadow, .hover-box-shadow-sm {
+  transition: box-shadow 0.3s;
+}
+
 .hover-box-shadow:hover {
   box-shadow: 3px 0px 11px 0px var(--boxShadow-primary)
 }
 
+.hover-box-shadow-sm:hover {
```
tomquirk, line 41: big moves. love it

src/utils/safe.ts
```diff
@@ -0,0 +1,9 @@
+import { SafeTransactionType } from 'components/v2/V2Project/V2ProjectSafeDashboard/V2ProjectSafeDashboard'
+
+// Links to a Gnosis safe multisig
+export const generateSafeUrl = (address: string) =>
+  `https://gnosis-safe.io/app/eth:${address}/transactions/queue`
+
+// Links to specific transaction on Gnosis Safe app
+export const generateSafeTxUrl = (transaction: SafeTransactionType) =>
```
tomquirk, line 8: **praise** beautiful

src/utils/safe.ts
```diff
@@ -0,0 +1,9 @@
+import { SafeTransactionType } from 'components/v2/V2Project/V2ProjectSafeDashboard/V2ProjectSafeDashboard'
```
tomquirk, line 1: **suggestion** WDYT about moving this file to `lib/`?

I think lib is a better spot for service-specific helpers (like safe, snapshot etc.)

src/components/v2/V2Project/V2ProjectSafeDashboard/V2ProjectSafeDashboard.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+      {isLoading && <div style={{ marginTop: 20 }}>Loading...</div>}
+
+      {!isLoading && (
+        <Tabs
+          style={{ marginTop: 20 }}
+          tabBarStyle={{ border: 'unset', marginBottom: 15 }}
+        >
+          <TabPane tab={<Trans>Queued</Trans>} key="1">
+            {queuedSafeTransactions?.results.map(
```
tomquirk, line 77: **question** Is the transaction state (e.g. if it's rejected) in the transaction data? If it is, maybe we could filter them out of this array.

Maybe (untested, just a guess):

```
queuedSafeTransactions.filter(tx => !tx.isExecuted)...
```

tomquirk (commented):
@johnnyd-eth looks so great! Thanks heaps for picking up the slack on this. Real exciting.

I left a few nitpicks. One other thing: can we make the tabs look the same as the Projects page?

Screen Shot 2022-09-14 at 9 01 40 PM

src/utils/safe.ts
johnnyd-eth: Good call

johnnyd-eth:
@tomquirk latest commit:
1. Creates new `Tab` component to use in for the safe tabs and `projects` page tabs.
2. Filters queued safe transactions by `safeTxGas` (removes any that aren't 0). As far as I can see, executed or rejected transactions always have some gas, and queued transactions are always 0. 

- I have tested some extra safe tx's and some such as [this](https://gnosis-safe.io/app/eth:0x07f356Fdb3EA88Fad8607A58434b35F528Ce9b1b/transactions/multisig_0x07f356Fdb3EA88Fad8607A58434b35F528Ce9b1b_0x348bad414c42f91c512c893cab6b73b79975c38ab8402ff09f980849aa566f32) are coming through with null data. 
- I'm not showing these. 
- I've reached out to Safe about what we can do about these. Think we should just merge what we've got for now.

tomquirk:
@johnnyd-eth LFG!!! I made one last commit, check it out if you're interested: https://github.com/jbx-protocol/juice-interface/pull/1965/commits/57ce2de00e43970ab7121bfa184b6cf72e943d47

Taking note of a couple of UX things that we should implement later:
- this list should have an initial limit of, say 10 txs. Then we have a **load more** button at the bottom of the list
- ... i think that's it actually.

Merging 🎉 🎉 🎉 🎉 

---

Note: One thing I've learnt recently is to generally avoid `closure` components; components that are defined inside other components. Can lead to buggy stuff with re-rendering. I moved ProjectTab outside of the ProjectTabs scope, for example.

PR #1966, opened 2022-09-13
Brute-force V3 PoC

tomquirk:


src/contexts/v3/projectContext.ts
```diff
@@ -0,0 +1 @@
+export { V2ProjectContext as V3ProjectContext } from '../v2/projectContext'
```
tomquirk, line 1: Reuse V2 context in V3. If it turns out we need a custom V3 context, all we need to do is remove this line, define a `V3ProjectContext` const and export that instead.

src/hooks/v3/V3ContractLoader.ts
```diff
@@ -0,0 +1,46 @@
...
+        const network = readNetwork.name
+
+        // Contracts can be used read-only without a signer, but require a signer to create transactions.
+        const signerOrProvider = signer ?? readProvider
+
+        const contractLoaders = await Promise.all(
+          Object.values(V3ContractName).map(contractName =>
+            loadContract(contractName, network, signerOrProvider),
+          ),
+        )
```
tomquirk, line 26: This File is mostly a clone of V2 file, except for this line.

We can probably abstract this loadContracts hook further for better re-use/DRY

src/hooks/v3/V3ProjectState.ts
```diff
@@ -0,0 +1 @@
+export { useV2ProjectState as useV3ProjectState } from '../v2/V2ProjectState'
```
tomquirk, line 1: Reuse V2 hook in V3. If it turns out we need a custom V3 hook, all we need to do is remove this line, define a useV3ProjectState fn and export that instead.

PR #1969, opened 2022-09-13
Reorganize filez

tomquirk:

- Moves any veNFT stuff out of `v2/` directories.
- Moves any NFT reward stuff out of `v2/` directories.
- Moves any `split` stuff out of `v2/` directories.
- Moves format-related utils under `utils/format/` directory
- Rename metadata domain to `JUICEBOX_MONEY_PROJECT_METADATA_DOMAIN`
- Move nft max reward tier const to util file
- Rename `loadContract` to `loadV2Contract`, and move to `v2/` dir

The motivation for all of this is to make v3 easier.
- NFT Rewards and veNFT aren't tied to a specific JB version necessarily. Inherently they aren't applicable to v1, but are for v1+N
- Splits aren't a V1 concept, so it would never make sense (or work) to use Split code in V1.
- Other stuff is just cleanup


tomquirk:
cc @jmilldotdev @johnnyd-eth

tomquirk:
@jmilldotdev mind reviewing this pls?

jmilldotdev (approved):

PR #1978, opened 2022-09-15
Project metadata context refactor

tomquirk:
This PR pulls `projectMetadata` out of V2ProjectContext into its own context: ProjectMetadataContext.

This data structure is used across V1 and V2 (and V3). The motivation is to reduce the number of components dependent on version-specific context (like V2ProjectContext).

We can probably put a lot more in this context too, like projectId, owner address, and other things. Playing it safe and small for now.

Towards V3.


src/components/activityEventElems/PayEventElem.tsx
```diff
@@ -17,16 +16,6 @@ import {
   smallHeaderStyle,
 } from './styles'
 
-// Maps a project id to an internal map of payment event overrides.
-const payEventOverrides = new Map<number, Map<string, string>>([
```
tomquirk, line 21: Unrelated removal. Don't need it.

src/components/formItems/ProjectReserved.tsx
```diff
@@ -1,7 +1,7 @@
 import { t, Trans } from '@lingui/macro'
 import { Form } from 'antd'
 import { ThemeContext } from 'contexts/themeContext'
-import { round } from 'lodash'
+import round from 'lodash/round'
```
tomquirk, line 4: Couldn't help myself

wraeth-eth (approved):

PR #1980, opened 2022-09-15
Create NftRewards context

tomquirk:
Towards V3.

Create an NftRewardsContext. This enables us to move NftRewards outside of the v2 directory. Yay!


tomquirk:
cc @johnnyd-eth

PR #1981, opened 2022-09-15
Remove veNFT from V2ProjectContext

tomquirk:
Towards V3.

This PR further removes dependence on V2ProjectContext by removing VeNFT from V2ProjectContext.

I haven't tested any of the VeNFT functionality but assume it should be fine.


tomquirk:
@jmilldotdev admittedly this is a little rushed and there's probably a nicer Context approach, but I'd really like to get venft out of v2projectcontext. If this is okay with you and functionality still works, let me know. Cheers

jmilldotdev (approved):

PR #2002, opened 2022-09-17
redirect to projects page

benschac:

going to `/p/projects` straight from the url bar gives the end user an error on local, and an infinite load on prod.

This change redirects the user to `/projects`
current: 
CleanShot 2022-09-17 at 11 04 41@2x

after: 
CleanShot 2022-09-17 at 11 03 40

## Acceptance checklist

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -17,7 +17,7 @@ module.exports = {
   async redirects() {
     return [
       {
-        source: '/p',
+        source: '/projects',
         destination: '/projects',
         permanent: true,
       },


wraeth-eth (changes requested):
Thanks for the PR @benschac .

I can see the error of going to `/p/projects` but the redirect being changed here isn't for that; this redirect is when a user goes to: `juicebox.money/p` with no follow-up for the project id, redirecting them to the `/projects` page. 

`/p/projects` is trying to find a v1 project with projectId = 'projects' which could exist (technically), but should be actually showing a 404 instead.

PR #2011, opened 2022-09-19
Fix tx noti etherscan link

tomquirk:
- breaks up TransactionList a bit
- Create single util for etherscan links (and fix the bug)
- Use util in TransactionList and EtherscanLink components


src/utils/etherscan.ts
```diff
@@ -1,7 +1,11 @@
 import { readNetwork } from 'constants/networks'
 import { NetworkName } from 'models/network-name'
 
-export const etherscanLinkForTxHash = (hash: string) =>
-  `https://${readNetwork.name}${
-    readNetwork.name !== NetworkName.mainnet ? '.' : ''
-  }etherscan.io/tx/${hash}`
```
tomquirk, line 7: This was producing a buggy link: `mainnetetherscan.io`.

PR #2024, opened 2022-09-20
Add callback support to transaction history poller

tomquirk:
The new transaciton poller didn't support the previous bnc-notify callbacks. This breaks any UI that relies on these callbacks (most of them).

This PR updates the transaction store to include callbacks for each tx. Callbacks are onSuccess and onCancelled.

Please test project creation and `pay` transactions. Note that, locally, page redirection is slow af.

### DIFF
--- a/src/providers/TxHistoryProvider.tsx
+++ b/src/providers/TxHistoryProvider.tsx
@@ -22,7 +22,7 @@ const pollTransaction = async (
 ): Promise<TransactionLog> => {
   // Only do refresh logic for pending txs
   // tx.hash shouldn't ever be undefined but it's optional typed :shrug:
-  if (!txLog.tx.hash || txLog.status !== TxStatus.pending) {
+  if (!txLog.tx?.hash || txLog.status !== TxStatus.pending) {
     return txLog
   }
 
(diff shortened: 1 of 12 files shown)


src/providers/TxHistoryProvider.tsx
```diff
@@ -81,41 +119,15 @@ export default function TxHistoryProvider({
...
-            : await readProvider.getTransaction(txLog.tx.hash)
-
-          let status = TxStatus.pending
-          try {
-            // Tx has been mined
-            status = TxStatus.success
-          } catch (_) {
-            // ethers provider throws error when a transaction fails
-            status = TxStatus.failed
-          }
```
tomquirk, line 104: I don't understand what this was doing prviously. Won't the `catch` never be possible? 🤔

wraeth-eth: It's possible code has been removed

src/providers/TxHistoryProvider.tsx
```diff
@@ -81,41 +119,15 @@ export default function TxHistoryProvider({
...
-            status,
-          }
+        transactions.map(txLog => {
+          return pollTransaction(txLog)
         }),
       )
 
       console.info(
-        'TxHistoryProvider::poller::settings transactions',
+        'TxHistoryProvider::poller::updating transactions state',
```
wraeth-eth, line 128: c++ style 😎

wraeth-eth (approved):

PR #2040, opened 2022-09-23
Make IssueErc20TokenModal compatible with all contract versions

tomquirk:

Updates `IssueErc20TokenModal` internals to use a different tx hook depending on the `cv` in context.

### DIFF
--- a/src/components/IssueErc20TokenButton.tsx
+++ b/src/components/IssueErc20TokenButton.tsx
@@ -25,9 +25,7 @@ export function IssueErc20TokenButton({
     setModalVisible(false)
 
     // remove newDeploy=true query parameter
-    router.replace(router.pathname, {
-      search: '',
-    })
+    router.replace(router.asPath, undefined, { shallow: true })
   }
 
   return (
--- a/src/components/IssueErc20TokenButton.tsx
+++ b/src/components/IssueErc20TokenButton.tsx
@@ -2,7 +2,6 @@ import { SettingOutlined } from '@ant-design/icons'
 import { Trans } from '@lingui/macro'
 import { Button, Tooltip } from 'antd'
 
-import { TransactorInstance } from 'hooks/Transactor'
 import { useRouter } from 'next/router'
 import { useState } from 'react'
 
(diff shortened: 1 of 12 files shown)


src/components/IssueErc20TokenButton.tsx
```diff
@@ -14,10 +13,8 @@ export type IssueErc20TokenTxArgs = {
 }
 
 export function IssueErc20TokenButton({
-  useIssueErc20TokenTx,
   onCompleted,
 }: {
-  useIssueErc20TokenTx: () => TransactorInstance<IssueErc20TokenTxArgs>
```
tomquirk, line 20: Move this logic to component internals so we dont need to prop drill

src/components/modals/IssueErc20TokenModal.tsx
```diff
@@ -1,23 +1,48 @@
...
 
 import { emitErrorNotification } from 'utils/notifications'
 import { IssueErc20TokenTxArgs } from '../IssueErc20TokenButton'
 import TransactionModal from '../TransactionModal'
 
+/**
+ * Return the appropriate issue erc20 token hook for the given contract version [cv].
+ * @returns
+ */
+const useIssueErc20TokenTx = ():
```
tomquirk, line 20: A similar outcome can be achieved with Context, but figured a simple hook dependent on ProjectMetadataContext is nice enough for now.

src/hooks/v3/transactor/V3IssueErc20TokenTx.ts
```diff
@@ -0,0 +1,56 @@
...
+  const { projectId, cv } = useContext(ProjectMetadataContext)
+
+  return ({ name, symbol }, txOpts) => {
+    try {
+      invariant(
+        transactor && projectId && contracts?.JBController && name && symbol,
+      )
+      return transactor(
+        contracts.JBTokenStore,
+        'issueFor',
```
tomquirk, line 28: In V3 we issue tokens using a different method on a different contract. So, this new hook does that :)

Issue #2029, opened 2022-09-20
Move the Transfer Unclaimed Tokens to Manage Tokens modal


tomquirk:
Move the **Transfer Unclaimed Tokens** section in the **Tools** drawer to **Manage Tokens** modal.

Current **Transfer Unclaimed Tokens** section in the **Tools** drawer:
<img width="575" alt="Screen Shot 2022-09-21 at 9 23 59 AM" src="https://user-images.githubusercontent.com/12551741/191384825-674c092a-577d-4314-9be2-4535ddcb6a47.png">

Current **Manage Tokens** modal:
<img width="603" alt="Screen Shot 2022-09-21 at 9 24 04 AM" src="https://user-images.githubusercontent.com/12551741/191384899-9041df0a-b876-40dd-89c7-7d861bfbf4f5.png">

PR #2043, opened 2022-09-23
Move transfer unclaimed tokens section from tools to manage

NakedFool:

Fixes - https://github.com/jbx-protocol/juice-interface/issues/2029

The intention of this PR is to move the `Transfer unclaimed tokens` section from `Tools` to `Manage`

Additional stuff I did while working on this:

- Added validation inside transfer unclaimed modals, you cannot proceed now if the amount is zero or the address is not validated
-  Added error handling in v2 ( same stuff I did previously with hooks )

The only thing that I am not sure about is if it is written well

image

### Before
image

### After
image
image

## Acceptance checklist


johnnyd-eth:
Nice one @NakedFool! Great initiative picking up that validation too.

One comment is that I think the `RichButton` can be completely hidden when the user's unclaimed balance is 0.

src/hooks/v2v3/transactor/TransferUnclaimedTokensTx.ts
```diff
@@ -17,12 +21,26 @@ export function useTransferUnclaimedTokensTx(): TransactorInstance<{
...
   const { contracts } = useContext(V2V3ContractsContext)
   const { tokenSymbol } = useContext(V2V3ProjectContext)
-  const { projectId } = useContext(ProjectMetadataContext)
+  const { projectId, cv } = useContext(ProjectMetadataContext)
 
   const { userAddress } = useWallet()
 
   return ({ amount, to }, txOpts) => {
-    if (!transactor || !projectId || !contracts?.JBTokenStore) {
+    try {
```
johnnyd-eth, line 29: **Fresh**

src/components/v1/V1Project/modals/TransferUnclaimedModal.tsx
```diff
@@ -0,0 +1,113 @@
...
+      centered
+    >
+      <Form form={form} layout="vertical">
+        <p>
+          <Trans>
+            Your unclaimed token balance:{' '}
+            {formatWad(unclaimedBalance, { precision: 0 })}
+          </Trans>
+        </p>
+        <Form.Item name="amount" label="Amount">
```
johnnyd-eth, line 90: ```t`Amount````

src/components/v1/V1Project/modals/TransferUnclaimedModal.tsx
```diff
@@ -0,0 +1,113 @@
...
+import { formatWad, fromWad, parseWad } from 'utils/format/formatNumber'
+import { tokenSymbolText } from 'utils/tokenSymbolText'
+
+interface TransferUnclaimedModalProps {
+  visible?: boolean
+  onCancel?: VoidFunction
+  onConfirmed?: VoidFunction
+}
+
+export const TransferUnclaimedModal: FC<TransferUnclaimedModalProps> = ({
```
johnnyd-eth, line 22: This is definitely valid and good thinking, but I think it's probably best if we stick to our current component naming convention:


```
export function TransferUnclaimedModal({
    visible,
    onClose,
    onConfirmed,
}: {
  visible: boolean
  onClose: VoidFunction
  onConfirmed: VoidFunction
}) 
```

**Note:** 
- Since none of these props are optional we don't need/want the `?:` typing 
- `onClose` might be a better name than `onCancel` for closing modals, and is consistent for our modals throughout the app I believe.

src/components/v1/V1Project/modals/TransferUnclaimedModal.tsx
```diff
@@ -0,0 +1,113 @@
...
+      onCancel={() => {
+        form.setFieldsValue({
+          amount: undefined,
+          to: undefined,
+        })
+        onCancel?.()
+      }}
+      okText={t`Transfer ${tokenTextShort}`}
+      okButtonProps={{
+        disabled: parseWad(amount).eq(0) || !isAddress(address),
```
johnnyd-eth, line 79: This is a decent approach, but I think, mainly for the sake of consistency with other forms throughout the app, it might be even better to go with this approach:
- pass validator functions as props into the two `<Form.Item`s. (e.g. see `validateTotalDistributions`)
- put an `await form.validateFields()` at the start of the `transferTokens` function
- Never disable the button 

The advantage of this approach is that we can show users why the form is not valid if they try to submit. It's not a big deal in this case since the form is so small, but I think it's a good practice to get into the habit of for future forms you will build. 

Let me know if this doesn't make sense, or if you have any trouble implementing.

johnnyd-eth:
One last thing: 
- it seems the only difference between the V1 and V2 instances of these modals are the `transferUnclaimedTokensTx`. 
- So I think we could make a single, shared `<TransferUnclaimedModal ` component and have it take the transaction as a prop. 
- Then pass the version-specific transactions into either instance of the component

Make sense?

src/hooks/v2v3/transactor/TransferUnclaimedTokensTx.ts
NakedFool: I've applied the same behavior for `v1` version of this function in the latest commit.

src/components/v1/V1Project/modals/TransferUnclaimedModal.tsx
NakedFool: great catch ser! <3

src/components/v1/V1Project/modals/TransferUnclaimedModal.tsx
NakedFool: I agree with you on both points, I've applied the first one as it really doesn't make sense for them to be optional as you said.

However, the second point would keep us inconsistent since all other modals are using `onCancel`, I can make another issue after this PR to rename all of these modals to use `onClose` , wdyt ?

src/components/v1/V1Project/modals/TransferUnclaimedModal.tsx
NakedFool: It is definitely healthier to allow users to click and if items are not filled show the message, thanks for this!

I've applied this behavior in the latest commit, but there is still inconsistency throughout the app

- Redeem and Burn modal disables the button
- Claim and Transfer unclaimed tokens enable button but display a message to a user.

I can create an issue after this PR and make all these modals consistent, wdyt?

NakedFool:
> One last thing:
> 
> * it seems the only difference between the V1 and V2 instances of these modals are the `transferUnclaimedTokensTx`.
> * So I think we could make a single, shared `<TransferUnclaimedModal ` component and have it take the transaction as a prop.
> * Then pass the version-specific transactions into either instance of the component
> 
> Make sense?

Yes, 110% on this!

I've created a shared `TransferUnclaimedTokensModal` component and now we need to pass only the tx function throughout the `ManageTokensModal`

All in all, I've applied changes in the latest commit, however, see the comments above and stuff that we can resolve in other PRs.

src/components/v1/V1Project/modals/TransferUnclaimedModal.tsx
johnnyd-eth: Sure, issues for these would be good with the `Maintenance` label. Not huge priorities though I'd put them on the back burner personally

src/components/v1/V1Project/modals/TransferUnclaimedModal.tsx
johnnyd-eth: For sure, that would be great.

src/components/v2v3/V2V3Project/V2V3ManageTokensSection/index.tsx
```diff
@@ -320,9 +323,12 @@ export default function V2ManageTokensSection() {
         redeemDisabled={redeemDisabled}
         tokenSymbol={tokenSymbol}
         tokenAddress={tokenAddress}
+        tokenUnclaimedBalance={unclaimedBalance}
+        transferUnclaimedTokensTx={useTransferUnclaimedTokensTx}
         RedeemModal={V2RedeemModal}
         ClaimTokensModal={V2ClaimTokensModal}
         MintModal={V2MintModal}
+        TransferUnclaimedTokensModal={TransferUnclaimedTokensModal}
```
johnnyd-eth, line 331: **Issue:** I don't think this prop is necessary. Since this component is shared it can just be imported within the shared `ManageTokensModal`

johnnyd-eth:
All functioning perfectly @NakedFool, thanks for addressing all my comments. Just one more comment then we'll get this one in

src/components/v2v3/V2V3Project/V2V3ManageTokensSection/index.tsx
NakedFool: Oooh, you are right, my bad!

NakedFool:
> All functioning perfectly @NakedFool, thanks for addressing all my comments. Just one more comment then we'll get this one in

Fixed it in the latest commit ser! ❤️

src/components/v1/V1Project/modals/TransferUnclaimedModal.tsx
NakedFool: Yeah no problem, just wanna address these so we do not forget in future 🍻

johnnyd-eth:
LFG merging @NakedFool

johnnyd-eth (approved):

PR #2064, opened 2022-09-29
NFT drawer unsaved changes modal

johnnyd-eth:

As in title. 

## Acceptance checklist


src/components/v2v3/V2V3Project/V2V3Project.tsx
```diff
@@ -170,7 +176,21 @@ export function V2V3Project({
 
   // Change URL without refreshing page
   const removeQueryParams = () => {
-    router.push(v2v3ProjectRoute({ projectId }), undefined, { shallow: true })
+    const newQuery: Record<string, string> = {}
```
wraeth-eth, line 179: What does this do?

src/components/v2v3/shared/FundingCycleConfigurationDrawers/NftDrawer/NftMarketplaceCustomizationForm.tsx
```diff
@@ -20,13 +23,33 @@ export function NftMarketplaceCustomizationForm({
...
+
+  useEffect(() => handleFormChange(), [handleFormChange])
+
   return (
-    <Form layout="vertical" form={form} initialValues={initialValues}>
+    <Form
+      layout="vertical"
+      form={form}
+      initialValues={initialValues}
+      onValuesChange={handleFormChange}
```
wraeth-eth, line 51: I'm about to be annoying here, and feel free to not enact this change (especially if this is working), but from a clean perspective, I think we should lift this logic (checking if form has updated) to its parent as it is (imo) not really a concern of the form itself. 

The parent component (`NftDrawer`) is actually the one that cares about the state changing.

We can make an issue for this (or solve here, depending on how you feel) but I think what would be better is that `NftMarketplaceCustomizationForm` mirrors the result of `Form.onValuesChange` and returns that.

`Form.onValuesChange` returns a subset of what values have actually changed.

This is actually a non-trivial change from what you have so I actually recommend not making this change here - lets a add a low priority issue instead - I can make it, but i'll wait to see what you think as you might disagree ;-)

src/utils/nftRewards.ts
```diff
@@ -181,3 +181,21 @@ export async function uploadNftCollectionMetadataToIPFS({
   })
   return res.data.IpfsHash as string
 }
+
+// Determines if two NFT reward tiers are equal
+export function tiersEqual({
```
wraeth-eth, line 186: It's at times like this, I actually miss class based languages with equality operators :-(

src/components/v2v3/shared/FundingCycleConfigurationDrawers/NftDrawer/NftMarketplaceCustomizationForm.tsx
wraeth-eth: Also happy to jump on a call to discuss this if it helps as I feel like I have kind of just jumbled a few incoherent sentences together lol

wraeth-eth (approved):

src/components/v2v3/shared/FundingCycleConfigurationDrawers/NftDrawer/NftMarketplaceCustomizationForm.tsx
johnnyd-eth: I like this idea @wraeth-eth, it'll be a good cleanup task. It will also apply to all the other drawers, correct? The new issue could tackle them all I guess

src/components/v2v3/V2V3Project/V2V3Project.tsx
johnnyd-eth: This was a leak from another PR. Out of the diff now

PR #2109, opened 2022-10-03
Refactor contract loader

tomquirk:
- move contract loaders to own files
- feature flag relevant contract loaders
- add goerli ballot addresses


tomquirk:
@wraeth-eth would you be so kind?

PR #2124, opened 2022-10-04
Context-aware cv in subgraph queries

tomquirk:

Make sure all subgraph queries either:
- use the cv provided in the current context
- include both V2 and V3 in their cv_in queries

## Acceptance checklist

### DIFF
--- a/src/models/cv.ts
+++ b/src/models/cv.ts
@@ -1,6 +1,7 @@
 import { V1TerminalVersion } from 'models/v1/terminals'
 export type V2CVType = '2'
 export type V3CVType = '3'
+export type V2V3CVType = V2CVType | V3CVType
 
 // CV = contracts version
-export type CV = V1TerminalVersion | V2CVType | V3CVType
+export type CV = V1TerminalVersion | V2V3CVType
--- a/src/models/cv.ts
+++ b/src/models/cv.ts
@@ -1,7 +1,7 @@
 import { V1TerminalVersion } from 'models/v1/terminals'
-export type V2CVType = '2'
-export type V3CVType = '3'
-export type V2V3CVType = V2CVType | V3CVType
+export type CV2 = '2'
+export type CV3 = '3'
+export type CV2V3 = CV2 | CV3
 
 // CV = contracts version
-export type CV = V1TerminalVersion | V2V3CVType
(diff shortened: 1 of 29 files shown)


src/pages/projects/index.page.tsx
```diff
@@ -93,8 +93,8 @@ function Projects() {
     const _cv: CV[] = []
     if (includeV1) _cv.push(CV_V1)
     if (includeV1_1) _cv.push(CV_V1_1)
-    if (includeV2) _cv.push(CV_V2)
-    return _cv.length ? _cv : [CV_V1, CV_V1_1, CV_V2]
+    if (includeV2) _cv.push(CV_V2, CV_V3)
```
peripheralist, line 96: ohh nice addition. i'm thinking maybe we should change the UI for that version picker to be more explicit about including v3, though that can be another PR

src/models/cv.ts
```diff
@@ -1,6 +1,7 @@
 import { V1TerminalVersion } from 'models/v1/terminals'
 export type V2CVType = '2'
 export type V3CVType = '3'
+export type V2V3CVType = V2CVType | V3CVType
```
peripheralist, line 4: lol i'm realizing these types are a bit redundant—cv" stands for "contracts version", so the extra V isn't needed

these could be `CV2Type`, `CV2V3Type` etc maybe. same thing with the constants names. maybe out of scope for this PR

peripheralist (approved):
hard to know for sure from the diff if this is _all_ the cv instances that need to be updated, but all changes so far look solid to me! left a few comments you might want to address, but approving and leaving it up to you 🙏

src/pages/projects/index.page.tsx
tomquirk: yeah totally agree, just went with this for completeness and speed but we should prob add another checkbox 👍

src/models/cv.ts
tomquirk: Yeah haha, gr8 point indeed.

I think CV2V3Type is tight. will change

tomquirk: > V2CVType | V3CVType

Yeah I think I got all these, but will check again 👍

PR #2289, opened 2022-10-19
Consider user preferred theme in theme selection

tomquirk:

If no user-stored preference, prefer their machine setting.

### DIFF
--- a/src/hooks/JuiceTheme.ts
+++ b/src/hooks/JuiceTheme.ts
@@ -1,9 +1,7 @@
-import type { ThemeContextType } from 'contexts/themeContext'
-
-import { useEffect, useState } from 'react'
-
 import { juiceTheme } from 'constants/theme'
 import { ThemeOption } from 'constants/theme/theme-option'
+import type { ThemeContextType } from 'contexts/themeContext'
+import { useEffect, useState } from 'react'
 
 const flattenNestedObject = (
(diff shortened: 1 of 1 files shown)


src/hooks/JuiceTheme.ts
```diff
@@ -19,32 +17,41 @@ const flattenNestedObject = (
...
-    Object.entries(juiceTheme(themeOption).radii).forEach(([key, value]) => {
-      if (!value) return
-      document.documentElement.style.setProperty(
-        '--radius-' + key,
-        value.toString(),
-      )
-    })
-  }
+export function useJuiceTheme(storageKey = 'jb_theme'): ThemeContextType {
+  const initialThemeOption = getInitialThemeOption(storageKey)
```
wraeth-eth, line 49: You need to make sure this is wrapped in a browser only ctx, or server will crash on attempting to read `window`

wraeth-eth (changes requested):

src/hooks/JuiceTheme.ts
tomquirk: @wraeth-eth great catch!!!! Checking for presence of `window` object now:

https://github.com/jbx-protocol/juice-interface/pull/2289/files#diff-0072acd67b85ccdf7fa522877db68c7dbeb8e13af574b56360d96626eda5f05aR21-R23

Do you think this is sufficient?

wraeth-eth (approved):
send it

PR #2302, opened 2022-10-19
Use V3 JBPrices contract to get ether price

tomquirk:
It's about time!


src/hooks/EtherPrice.ts
```diff
@@ -1,59 +1,32 @@
...
-  return contract
-}
-
+/**
+ * Return the current price of ETH in USD.
+ * @example 1234.69
+ */
 export function useEtherPrice() {
-  const contract = usePricesContract()
+  const JBPrices = useLoadV2V3Contract({
```
tomquirk, line 21: Load this contract specifically here, instead of relying on context. Means it'll work on V1 projects, where the V2V3 contracts context isn't available.

Issue #2335, opened 2022-10-24
Etherscan link on pending transaction modal


johnnyd-eth:
## Description

<img width="691" alt="Screen Shot 2022-10-24 at 10 09 15 am" src="https://user-images.githubusercontent.com/96150256/197425137-29ba0f9c-ecd1-47d1-9fee-e1279d20dc6e.png">

Link to the pending transaction on Etherscan from this modal

## Proposal

_A description of the change you'd like to see._

## Screenshots

_If applicable, add screenshots to help explain your problem._

PR #2376, opened 2022-10-26
Add etherscan link for pending txs in transaction modal

NakedFool:

Fixes - https://github.com/jbx-protocol/juice-interface/issues/2335

I'm using `TxHistoryContext` to retrieve transactions, and from there i'm just finding the first one that matches the pending status.

I've created 3 iterations, i'm a fan of the first one but let me know which one we should decide or if we want some different UI iteration.

| First iteration| Second Iteration | Third Iteration
| ----------- | ----------- | ----------- | 
| first iteration     | second iteration| third iteration     |

## Acceptance checklist

### DIFF
--- a/src/components/TransactionModal.tsx
+++ b/src/components/TransactionModal.tsx
@@ -25,6 +25,7 @@ const PendingTransactionModalBody = () => {
   const { transactions } = useContext(TxHistoryContext)
 
   const pendingTx = transactions?.find(tx => tx.status === TxStatus.pending)
+  const pendingTxHash = pendingTx?.tx?.hash
 
   return (
     <div
@@ -48,20 +49,16 @@ const PendingTransactionModalBody = () => {
         </h2>
         <p>
           <Trans>
-            Your transaction has been submitted and is awaiting confirmation{' '}
+            Your transaction has been submitted and is awaiting confirmation.
           </Trans>
         </p>
-        <p>
-          {pendingTx ? (
-            <EtherscanLink
-              value={pendingTx.tx?.hash}
-              style={{ fontSize: '0.85rem' }}
-              type="tx"
-            >
+        {pendingTxHash ? (
+          <p>
+            <EtherscanLink value={pendingTxHash} type="tx">
               <Trans>View in Etherscan</Trans>
             </EtherscanLink>
-          ) : null}
-        </p>
+          </p>
+        ) : null}
       </div>
     </div>
   )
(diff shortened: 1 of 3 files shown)


src/components/TransactionModal.tsx
```diff
@@ -42,8 +48,11 @@ const PendingTransactionModalBody = () => {
         </h2>
         <p>
           <Trans>
-            Your transaction has been submitted and is awaiting confirmation.
-          </Trans>
+            Your transaction has been submitted and is awaiting confirmation{' '}
+          </Trans>{' '}
+          {pendingTx ? (
+            <EtherscanLink value={pendingTx.tx?.hash} type="tx" />
```
tomquirk, line 54: **suggestion** wdyt about a text link "Go to Etherscan" or "View in Etherscan" instead?

NakedFool: yup, i've chosen "View in Etherscan"

updated in the latest commit ser 🍻

NakedFool: image

src/components/TransactionModal.tsx
```diff
@@ -42,9 +49,24 @@ const PendingTransactionModalBody = () => {
         </h2>
         <p>
           <Trans>
-            Your transaction has been submitted and is awaiting confirmation.
+            Your transaction has been submitted and is awaiting confirmation{' '}
           </Trans>
         </p>
+        <p>
+          {pendingTx ? (
+            <ExternalLink
```
tomquirk, line 57: **suggestion** Could we use EtherscanLink here perhaps?

`src/components/EtherscanLink.tsx`

tomquirk:
Thanks @NakedFool looks great! One last comment

src/components/TransactionModal.tsx
NakedFool: Yeah, we can use `<EtherscanLink/>` but that component also returns an icon so it looks like this.
I've added children inside that component if `EtherscanLink` is a type of `tx`
image

It looks cool to me, wdyt?

NakedFool: We can put logic to remove the icon if there are children, but this looks good to me the way it is! :)

tomquirk (approved):
LGTM @NakedFool !!!

Added one commit, but merging. 💪  https://github.com/jbx-protocol/juice-interface/pull/2376/commits/5e5d299a5270f3375543842eeb71a6a510289596

PR #2387, opened 2022-10-27
Read project owner address from context for hasPerm hook

tomquirk:
Before, we were doing a network request each time we called hasPermission. Inneficient! This PR updates it so we use the pre-fetched porjectOwnerAddress from the context.

### DIFF
--- a/src/hooks/v2v3/contractReader/V2ConnectedWalletHasPermission.ts
+++ b/src/hooks/v2v3/contractReader/V2ConnectedWalletHasPermission.ts
@@ -9,9 +9,8 @@ export function useV2ConnectedWalletHasPermission(
   permission: V2OperatorPermission | V2OperatorPermission[],
 ) {
   const { userAddress } = useWallet()
-  const { isPreviewMode } = useContext(V2V3ProjectContext)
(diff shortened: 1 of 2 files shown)


src/hooks/v2v3/contractReader/V2ConnectedWalletHasPermission.ts
```diff
@@ -13,20 +11,21 @@ export function useV2ConnectedWalletHasPermission(
   const { userAddress } = useWallet()
   const { isPreviewMode } = useContext(V2V3ProjectContext)
```
tomquirk, line 12:
```suggestion
  const { isPreviewMode, projectOwnerAddress } = useContext(V2V3ProjectContext)
```

src/hooks/v2v3/contractReader/V2ConnectedWalletHasPermission.ts
```diff
@@ -13,20 +11,21 @@ export function useV2ConnectedWalletHasPermission(
...
   const { projectId } = useContext(ProjectMetadataContext)
-
-  const { data: owner } = useProjectOwner(projectId)
+  const { projectOwnerAddress } = useContext(V2V3ProjectContext)
```
tomquirk, line 14:
```suggestion
```

PR #2389, opened 2022-10-27
Get handle, owner addr from subgraph

tomquirk:
Realized we can save 2 Infura requests by sourcing the project owner address and project handle from the subgraph.


tomquirk:
@peripheralist wdyt about this? Curious if you think this is reliable enough or whether we should be extra safe and prefer to fetch stuff like this from chain.

peripheralist:
subgraph integrity feels fine, but it's also nice to keep exploit vectors to a minimum by getting important data from the chain directly, at least for the dashboard. handle doesn't feel _too_ critical to verify on chain, but i'm on the fence about project owner

tomquirk:
@peripheralist thanks for the review ser, gonna close it for now, think a 'relying on on-chain where possible' is a good philosophy

PR #2417, opened 2022-10-31
Fix Queued safe txs

tomquirk:
- Fix filtering logic on queued safe fetch
- update safe API url
- safe fetch refactor
- type all safe fetches
- smol style changes


src/lib/safe.ts
```diff
@@ -1,9 +1,74 @@
...
+
+  const response = await axiosInstance.get<{ results: SafeTransactionType[] }>(
+    `/safes/${safeAddress}/multisig-transactions`,
+    { params },
+  )
+
+  // Ideally the Safe API would return the transactions filtered properly...
+  return response.data.results.filter(
+    (tx: SafeTransactionType) => tx.blockNumber === null,
+  )
```
tomquirk, line 73: This is the only change to the logic.

Use blockNumber instead of safeTxGas to filter. Maybe the safe API changed...

src/lib/safe.ts
```diff
@@ -1,9 +1,74 @@
...
+export type SafeApiParams = {
+  limit?: number
+  executed?: boolean
+  nonce__gt?: number
+  nonce__gte?: number
+  nonce__lt?: number
+  nonce__lte?: number
+}
+
+const SAFE_API_BASE_URL = `https://safe-transaction-${readNetwork.name}.safe.global/api/v1`
```
tomquirk, line 14: Also the API hostname changed lol

tomquirk: https://safe-transaction-mainnet.safe.global/

src/lib/safe.ts
johnnyd-eth: Legend.

johnnyd-eth (approved):
Nice fixes here @tomquirk. All looks good to me.

Issue #2157, opened 2022-10-07
Prefix Discord field with `https` like NFT 


johnnyd-eth:
## Description

Prefix Discord field in project create flow with `https` like the NFT reward tier website link field.

<img width="536" alt="Screen Shot 2022-10-07 at 11 45 48 am" src="https://user-images.githubusercontent.com/96150256/194448973-a0e9a067-da09-41ab-8a4f-85ce5152b639.png">

Possibly could make a shared component out of this, can't remember what it is right now off the top of my head.

## Proposal

_A description of the change you'd like to see._

## Screenshots

_If applicable, add screenshots to help explain your problem._

PR #2425, opened 2022-11-01
added shared component for prefix inputs

NakedFool:

Fixes - https://github.com/jbx-protocol/juice-interface/issues/2157

I've added shared component `PrefixInput` (the name may be dumb but please suggest if you have a better one 😺 ) that is going to handle all inputs that need prefixes throughout the app.

The first example was in the creation flow, therefore I first adjusted in NFT rewards input that is responsible for external links ( you gonna see in the video ), and then I've added prefix input in details for project links.

one note: CSS is acting weird in project links, @wraeth-eth do you have any idea since this is your baby? 🙏 

Small touchups: 

- created `url.ts` and moved functions over from `SocialLinks.tsx` so we can use them everywhere such as -  `prettyUrl` , `linkUrl`
-  adjusted `prettyUrl` function a bit, now we also remove slash at the end of URL, for example in past we displayed `example.com/` and now we display `example.com`

## **Before**

https://user-images.githubusercontent.com/18723426/199320841-979a325f-a84d-403d-a7ba-59415b4373af.mov

## **After**

https://user-images.githubusercontent.com/18723426/199320869-92b736e2-6679-480b-8ac4-580e7a04bbe1.mov

image

## Acceptance checklist


johnnyd-eth:
This is great @NakedFool. I think maybe we could change the name to `PrefixedInput`. I also think the default value 'https' is probably unnecessary, and we should just require giving it as the `prefix` prop in each instance of the component. Looks great to me otherwise.

johnnyd-eth (changes requested):
.

NakedFool:
> This is great @NakedFool. I think maybe we could change the name to `PrefixedInput`. I also think the default value 'https' is probably unnecessary, and we should just require giving it as the `prefix` prop in each instance of the component. Looks great to me otherwise.

Fixed conflict and applied changes ser!

tomquirk:
@johnnyd-eth @NakedFool shall we get this merged?

NakedFool:
> @johnnyd-eth @NakedFool shall we get this merged?

it looks good to me now, there was a minor conflict with importing stuff but now that's resolved.

johnnyd-eth (approved):
Whoops my bad for lagging on this one! LGTM

PR #2445, opened 2022-11-03
Revert "Revert "Transactor arcx fix""

tomquirk:
Reverts jbx-protocol/juice-interface#2444

Refactor the Transactor a bit. Removes the logic in there to support a local ETH chain, for easier maintainability/readability.

Fix Arcx tx submissions

Also removes some unecessary anys and replaces with unknown

### DIFF
--- a/src/hooks/Transactor.ts
+++ b/src/hooks/Transactor.ts
@@ -48,7 +48,7 @@ function prepareTransaction({
   functionName: string
   contract: Contract
   args: unknown[]
-  options?: TransactionOptions
+  options: TransactionOptions | undefined
 }) {
   const tx =
     options?.value !== undefined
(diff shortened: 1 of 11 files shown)


src/hooks/Transactor.ts
```diff
@@ -82,90 +94,61 @@ export function useTransactor({
...
-          if (!tx.gasLimit) tx.gasLimit = hexlify(120000)
+        console.info('✅ Transactor::submitted', result)
 
-          result = await signer.sendTransaction(tx)
-          const txResponse = result as TransactionResponse
-
-          addTransaction?.(txTitle, txResponse, {
-            onConfirmed: options?.onConfirmed,
-            onCancelled: options?.onCancelled,
-          })
```
tomquirk, line 141: this code is never actually used. It's only used when you have a local eth chain set up, which none of us ever do.

I think its safe to remove simply for the sake of maintainability of this file.

johnnyd-eth: Ok sounds good

johnnyd-eth (approved):
Code looks good and pay issue seems to be fixed. Nice work.

PR #2449, opened 2022-11-04
Random changes to support defifa

tomquirk:
- support many tiers in UI
- safer image fetching
- support limit override in tiers read


tomquirk:
cc @johnnyd-eth

PR #2525, opened 2022-11-18
V3 feature flag cleanup

tomquirk:
⚠️ ⚠️ ⚠️ ⚠️  Depends on https://github.com/jbx-protocol/juice-interface/pull/2524, merge it first


src/components/Create/Create.tsx
```diff
@@ -40,9 +38,8 @@ export function Create() {
   }
 
   return (
-    <V2V3ContractsProvider
-      initialCv={featureFlagEnabled(FEATURE_FLAGS.V3) ? CV_V3 : CV_V2}
-    >
+    // New projects will be launched using V3 contracts.
+    <V2V3ContractsProvider initialCv={CV_V3}>
```
tomquirk, line 42: i think this is the old create flow anyway, but just to be safe

PR #2532, opened 2022-11-20
1/n Derive terminals from fund access constraints at tx time.

tomquirk:
Makes fund access constraints the single source of truth for terminals.


tomquirk:
cc @wraeth-eth just merged this but just an FYI about this change.

PR #2545, opened 2022-11-22
Update payMetada rewardTiers based on amount in pay form

peripheralist:
## What does this PR do and why?

- For NFTRewards projects, when the amount in the pay form changes, update the selected NFT rewards `payMetadata.tierIdsToMint` to include the highest-tier affordable NFT tier id. This removes the `selectedTierIds` state object and uses `payMetadata` instead.
- To allow users to pay a project without minting a NFT, don't reset pay amount when deselecting a NFT tier
- Allow payments of 0 ETH

Manually selecting/deselecting NFT reward tiers will override the pay amount, and allow for minting multiple NFT tiers.

## Screenshots or screen recordings

https://user-images.githubusercontent.com/79433522/203423096-9536a00a-3179-4002-b8e7-dd7e84529542.mov

Update: Allowing user to deselect NFT tier without resetting payment amount
https://user-images.githubusercontent.com/79433522/203425488-be1e7c5f-d6b3-4086-963b-eb2271dea0b6.mov

## Acceptance checklist

- [ ] I have evaluated the [Approval Guidelines](https://github.com/jbx-protocol/juice-interface/blob/main/CONTRIBUTING.md#approval-guidelines) for this PR.
- [ ] I have tested this PR in [all supported browsers](https://github.com/jbx-protocol/juice-interface/blob/main/CONTRIBUTING.md#supported-browsers).
- [ ] I have tested this PR in dark mode and light mode (if applicable).

PR #2546, opened 2022-11-23
Allow 0 eth payments

tomquirk:
Follow-up to https://github.com/jbx-protocol/juice-interface/pull/2545/files. Protocol allows 0-eth payments so we should too.


tomquirk:
cc @peripheralist

peripheralist:
nice @aeolianeth , thanks

somehow i didn't push these commits before merging that PR yesterday 🙄 https://github.com/jbx-protocol/juice-interface/compare/main...fix-nft-dontmint

PR #2548, opened 2022-11-23
Never load NFTs if no datasource

tomquirk:
This is a bug atm: 

scenario:
- user has a v2 and v3 FC
- V3 FC has NFTs
- User is on the `reconfigure FC` page in settings
- User switches from V3 to V2

Bug: V3 NFTs are still loaded and shown in Review and Deploy.

The bug is that, for some unknown reason, `tiersOf` is still returning the V3 tiers even when the datasource being sent in args is `0x00`.

### fix

This PR adds a failsafe to ensure the NFT context will never be populated without a valid (non-zero) datasource. Not great but works.

### Also

- removes some dead unused code related to NFTs (`yarn find-deadcode | grep nft`)


src/hooks/NftRewards.ts
```diff
@@ -50,11 +50,12 @@ async function getRewardTierFromIPFS({
 export default function useNftRewards(
   tiers: JB721TierParams[],
   projectId: number | undefined,
+  dataSourceAddress: string | undefined,
 ): UseQueryResult<NftRewardTier[]> {
   const hasTiers = Boolean(tiers?.length)
 
   return useQuery(
-    ['nft-rewards', projectId],
```
tomquirk, line 57: this is nice to have but doesn't actually fix this specific bug. but we should still have it here.

src/hooks/Transactor.ts
```diff
@@ -35,8 +35,13 @@ function logTx({
       {},
     )
 
+  const log = [
+    `functionName=${functionName}`,
+    `contractAddress=${contract.address}`,
```
tomquirk, line 40: add contractAddress to log for more context

src/providers/v2v3/NftRewardsProvider.tsx
```diff
@@ -6,31 +6,37 @@ import { useNftRewardTiersOf } from 'hooks/JB721Delegate/contractReader/NftRewar
...
    * Load NFT Rewards data
    */
   const { data: nftRewardTiersResponse, loading: nftRewardsCIDsLoading } =
     useNftRewardTiersOf({
-      dataSourceAddress: dataSourceAddress,
+      dataSourceAddress,
+      shouldFetch,
     })
+  // catchall to ensure nfts are never loaded if hasNftRewards is false (there's no datasource).
+  const tierData = shouldFetch ? nftRewardTiersResponse ?? [] : []
```
tomquirk, line 28: this is the actual fix. even if dataSource is 0 or undefined, there's some sketchy race condition or cache issue where `nftRewardTiersResponse` is still defined sometimes.

This line will set it to [] if there's no datasource (failsafe, hopefully).

src/utils/nftRewards.ts
```diff
@@ -73,19 +73,6 @@ export function getNftRewardTier({
   return nftReward
 }
 
-// Sorts array of nft reward tiers by contributionFloor
```
tomquirk, line 76: unused these days.

src/providers/v2v3/NftRewardsProvider.tsx
johnnyd-eth: Nice ok, good failsafe

src/utils/nftRewards.ts
johnnyd-eth: Yep cool

johnnyd-eth (approved):
Nice one here @tomquirk. All LGTM

PR #2570, opened 2022-11-25
Add amount to edit payouts page in v2 / v3 projects

NakedFool:

Closes https://github.com/jbx-protocol/juice-interface/issues/1620

I've added a boolean check `isEditPayoutPage` so we can determine if user is on edit payouts page and in that case display both  `amount` and `percentage` fields if there is `distributionLimit`

From what I've tested, I've done reconfiguration and editing payouts, and everything worked correctly. Let me know if some code-wise changes are needed! 🍻 

My next comment in this PR was working on a bug that persisted before and is fixed now ♥️ 

**Distribution limit is set**

https://user-images.githubusercontent.com/18723426/204762295-f82c9e1e-81ff-4925-9972-9d5191844929.mov

**Distribution limit infinite**

https://user-images.githubusercontent.com/18723426/204762460-35cbcf93-7bb5-49d5-9bf7-0dd86ca02f28.mov

## Acceptance checklist


src/components/v2v3/V2V3Project/V2V3ProjectSettings/pages/PayoutsSettingsPage/V2V3EditPayouts.tsx
```diff
@@ -299,7 +299,13 @@ export const V2V3EditPayouts = ({
         </Button>
       </Space>
       <DistributionSplitModal
+        isEditPayoutPage={true}
```
johnnyd-eth, line 302: Can just be `isEditPayoutPage` (no `={true}`)

src/components/v2v3/shared/FundingCycleConfigurationDrawers/DistributionSplitModal/AmountFormItem.tsx
```diff
@@ -103,21 +104,20 @@ export function AmountFormItem({
...
-                    onCurrencyChange={onCurrencyChange}
-                    currency={currencyName}
-                  />
-                ) : (
-                  <InputAccessoryButton content={currencyName} />
-                )
-              }
-            />
-          </div>
+          <FormattedNumberInput
```
johnnyd-eth, line 107: **Nice**

johnnyd-eth:
> The bug i mentioned also appears in the `main` branch when i edit code.
> 
> ~I've tried setting `value={amount}` on `AmountFormItem` component so i can see what the value is when I edit payouts in reconfiguring FC, I couldn't edit the amount to be higher than 1k for example. Still looking into it, it could be something due to formatted.~
> 
> ### EDIT
> Found it finally! Damn, i'm so happy right now. It was because `FormattedNumberInput` was wrapped with div, this is in component `AmountFormItem`, therefore value wasn't being passed to its input 🍻
> 
> One note to myself is to always read docs first before debugging, spent much time looking at other places while I could read docs of antd library first 🦆

Haha legendary @NakedFool! Love it man

NakedFool:
> Sorry for leaving this review so long ser @NakedFool!
> 
> First thing: noticed a bug when typing in a new a payout amount when the distribution limit is `zero`: Screen Shot 2022-12-01 at 10 09 34 pm
> 
> In this case the percentage should be disabled I think. EDIT: Actually maybe should show `100%`. I think this is what it's doing in V1, just copy whatever it's doing there

Thanks for catching this! I was focused only if there has distribution limit ( including infinite )

I'm keeping now only percentages in case dist limit is zero, it was this way before.

johnnyd-eth:
Other than this, looks great! Sorry for the delay, we'll get this in as soon as that last comment is addressed :)

NakedFool:
> Screen Shot 2022-12-05 at 12 28 21 pm
> 
> We don't need the 'read-only' percent next to the amount input when the percent slider is below it.
> 
> When the percent slider is present, hide the read-only percent, and move the '?' tooltip next to the percent slider.
> 
> WDYT?

Yes! Great idea, we definitely do not need to display the same value twice, added in the last commit. 

Let me know what you think of the approach, I've needed to add `children` prop to `NumberSlider` since if we keep it outside of it, it does not adjust height relative to its parent, therefore can't center it vertically.

johnnyd-eth:
> > Screen Shot 2022-12-05 at 12 28 21 pm
> > We don't need the 'read-only' percent next to the amount input when the percent slider is below it.
> > When the percent slider is present, hide the read-only percent, and move the '?' tooltip next to the percent slider.
> > WDYT?
> 
> Yes! Great idea, we definitely do not need to display the same value twice, added in the last commit.
> 
> Let me know what you think of the approach, I've needed to add `children` prop to `NumberSlider` since if we keep it outside of it, it does not adjust height relative to its parent, therefore can't center it vertically.

I'm confused, can't we just add a flex `alignItems: centre` (or whatever class we're using for that now) to the parent div at `PercentageFormItem.tsx:25`? 
I would rather we not make any modifications to `NumberSlider` in case it affects other instances of it.

NakedFool:
> > > Screen Shot 2022-12-05 at 12 28 21 pm
> > > We don't need the 'read-only' percent next to the amount input when the percent slider is below it.
> > > When the percent slider is present, hide the read-only percent, and move the '?' tooltip next to the percent slider.
> > > WDYT?
> > 
> > 
> > Yes! Great idea, we definitely do not need to display the same value twice, added in the last commit.
> > Let me know what you think of the approach, I've needed to add `children` prop to `NumberSlider` since if we keep it outside of it, it does not adjust height relative to its parent, therefore can't center it vertically.
> 
> I'm confused, can't we just add a flex `alignItems: centre` (or whatever class we're using for that now) to the parent div at `PercentageFormItem.tsx:25`? I would rather we not make any modifications to `NumberSlider` in case it affects other instances of it.

Oooh, this was my bad, I didn't see `mb` class in `NumberSlider`, therefore I assumed they couldn't be aligned vertically. 

It is working correctly now without adding it as children to slider.

johnnyd-eth (approved):
Awesome @NakedFool will merge this now!!  Great job 🍻

PR #2631, opened 2022-12-05
Add quick project search menu

peripheralist:

Adds a quick project search menu that can be accessed using `ctrl/cmd + k`, allowing user to search for projects and select one to navigate to by clicking, or by using arrow keys or `tab` and then `enter`

https://user-images.githubusercontent.com/79433522/205755967-60c0e912-90e1-4d4a-8b04-5158ee821ebf.mov

image

## Acceptance checklist

### DIFF
--- a/src/components/QuickProjectSearch.tsx
+++ b/src/components/QuickProjectSearch.tsx
@@ -35,7 +35,7 @@ export default function QuickProjectSearch() {
     if (!project) return
 
     router.push(
-      project.pv === PV_V2 ? `/@${project.handle}` : `/p/${project.projectId}`,
+      project.pv === PV_V2 ? `/@${project.handle}` : `/p/${project.handle}`,
     )
   }, [router, searchPages, highlightIndex])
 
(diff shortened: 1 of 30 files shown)


src/components/QuickProjectSearch.tsx
```diff
@@ -0,0 +1,171 @@
...
+            break
+          case 'Return':
+          case 'Enter':
+            goToProject()
+            reset()
+            break
+        }
+      } else if (e.key === HOT_KEY && (e.metaKey || e.ctrlKey)) {
+        setSpotlightActive(true)
+        setTimeout(
```
wraeth-eth, line 66: NBD but I think we can use refs to solve this - I read about them a month or so ago and this is their exact use case, but lets worry about it when its a problem (may never be)

wraeth-eth (approved):

src/components/QuickProjectSearch.tsx
peripheralist: good suggestion. i'm about to add a debounce to the typing to limit graph requests, and i'll look at this in the same PR. gonna merge this for now since everything seems functional at least

PR #2657, opened 2022-12-08
Add datasource info to FC card

tomquirk:
This adds minimal data source info to the FC card. Can add warnings/other props later.

| before | after |
| --- | --- |
| Screenshot 2022-12-08 at 4 19 26 PM | Screenshot 2022-12-08 at 4 14 45 PM |


johnnyd-eth (changes requested):
All looks good to me. Although wondering if we should hide the section if there is no dataSource (`AddressZero`)?

tomquirk:
@johnnyd-eth yes great idea

johnnyd-eth (approved):
Just pushed that empty state ✅ will merge

Issue #2611, opened 2022-12-04
Loading state on pay button shrinks the button


tomquirk:
<img width="594" alt="Screenshot 2022-12-05 at 10 02 48 AM" src="https://user-images.githubusercontent.com/12551741/205522581-9654454b-9803-4147-8792-d585925f3828.png">

<img width="495" alt="Screenshot 2022-12-06 at 1 39 09 PM" src="https://user-images.githubusercontent.com/12551741/205802401-1a6c32f4-5459-4f90-a8c1-41b4a75d8f1b.png">

PR #2662, opened 2022-12-08
Fix pay button width when it is loading or insufficient funds

NakedFool:

Closes https://github.com/jbx-protocol/juice-interface/issues/2611

I encountered some challenges while researching this issue. 

Eventually, i found that using inline styling was the most effective solution. The key is to keep in mind that when the button is disabled, it gets wrapped in a span, so any classes applied to it will be appended to the span rather than the button itself.

I tried a few things, but so far i've only been able to make it work with inline styling. Let me know if you have any thoughts!

image

## Acceptance checklist


johnnyd-eth (approved):
I'm happy with this.

Issue #2684, opened 2022-12-11
Banny stickers in pay memo are not attached


NakedFool:
## Summary

Banny stickers are not included in the sticker selection on pay memo.

![image](https://user-images.githubusercontent.com/18723426/206927256-cdbd0650-33f8-48cb-a20c-75379ae5fdb3.png)

## Environment

- **Device/OS**: [e.g. iOS]
- **Browser:** [e.g. chrome, safari]
- **Hardware wallet**: [e.g. Ledger w/ Metamask]

PR #2685, opened 2022-12-11
Fix stickers in pay memo

NakedFool:

Closes https://github.com/jbx-protocol/juice-interface/issues/2684

I've added `nftRewardTiers` as a dependency in a hook, let me know if i'm missing some context here! :)

https://user-images.githubusercontent.com/18723426/206927655-2bc566fd-83c5-44ad-bed9-e44914cce862.mov

## Acceptance checklist

### DIFF
--- a/src/components/v2v3/V2V3Project/V2V3PayButton/V2V3ConfirmPayModal/V2V3PayForm.tsx
+++ b/src/components/v2v3/V2V3Project/V2V3PayButton/V2V3ConfirmPayModal/V2V3PayForm.tsx
@@ -64,8 +64,10 @@ export const V2V3PayForm = ({
     const initialStickerUrls = nftRewardTiers?.map(
       (tier: NftRewardTier) => tier.imageUrl,
     )
+
     form.setFieldsValue({ stickerUrls: initialStickerUrls })
-  })
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [nftRewardTiers])
 
   return (
     <>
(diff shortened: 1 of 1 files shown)


src/components/v2v3/V2V3Project/V2V3PayButton/V2V3ConfirmPayModal/V2V3PayForm.tsx
```diff
@@ -64,8 +64,10 @@ export const V2V3PayForm = ({
     const initialStickerUrls = nftRewardTiers?.map(
       (tier: NftRewardTier) => tier.imageUrl,
     )
+
     form.setFieldsValue({ stickerUrls: initialStickerUrls })
-  })
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
tomquirk, line 69: **question** is there a way that we can not ahve this line? Maybe we need `form` as dependency.

Handing this over to @wraeth-eth for reivew though <3

NakedFool: I've seen some discussion about the benefits of not disabling the exhaustive deps warning in the useEffect hook, it is recommended in most cases to not disable it and I see a reason for that.

I'm still not sure why we would need to include the form object in the dependencies array here, since we're only setting the form value when nftRewardTiers changes. I'm happy to discuss this more and hear your thoughts! 🍻

wraeth-eth (approved):
THanks for this one blaz! tbh, I can't even remember the context of that useEffect, but it works so lets ship it!

PR #2714, opened 2022-12-19
Add Sepana API/search UI

peripheralist:

This PR integrates jbm with [Sepana](https://sepana.io/), including:
- Using sepana queries for project search _(feature flagged. enable with `jb.enableFeatureFlag("sepanaSearch")` in browser console)_
- Adding sepana types, including a `SepanaProject` type representing project data returned by querying Sepana
- Adding sepana NextJS api endpoints
  - `/api/sepana/projects?text=<search-text>&pageSize=<page-size>` returns projects matching a search query with a max page size. all projects will be returned if search text is empty. (this can be later optimized for sorting/filtering)
  - `/api/sepana/update` updates the Sepana db with the latest project data. this logic 1) queries the subgraph for any new projects, or projects that have changed since they were last stored, and 2) stores the updated project data (including metadata resolved from ipfs) to the sepana db. this endpoint is open and will be called periodically by our Upstash scheduled job (writes to the sepana db are authenticated with key stored here as a .env var)
  - `/api/sepana/health` this checks the integrity of the data stored in Sepana with data from the Subgraph, and is called periodically by our Upstash scheduled job
  - `/api/sepana/jobs=<id-1>,<id-2>,...` this gets data for all Sepana jobs with matching IDs. Sepana creates a "job" for every requested change to the database. This is helpful for inspecting a job if it seems to be failing or taking too long.
- Adding helper scripts to /scripts/sepana for deleting records from a Sepana db

### UI/UX
The only UI/UX changes in this PR are to the quick project search:
- using `project.name` in place of `project.handle` if no handle is defined
- including the project volume (volume is used in calculating search relevance, and including it makes the results feel more intuitive)

### Logging
A logger `sepanaAlert()` has been set up to send updates from the Sepana db to a Discord channel via webhook. These include both error reporting and sanity checks (Discord role will only be tagged for problems)

image

image

## Acceptance checklist


filipvvv:
@peripheralist added a DB health checker in a recent commit. Was having some typing problems with keys so used a ts-ignore, would love your advice on that one.

file:///home/filipv/1671826035.png

Getting this error when I try to use your `update` endpoint, but not sure why. Some outstanding questions:

1. How do we call the endpoint every n minutes? Wraeth suggested [Upstash](https://upstash.com/), which looks good to me. Would like to minimize new infra if possible.
2. How do we manage permissions for the endpoint (we probably don't want to put a read/write API key in prod)? Might make sense to move the API key from env vars to the headers of a given request. If this sounds good to you, let me know and I'll add it!

Thanks again for getting this PR together, super exciting.

filipvvv:
>   * `/api/sepana/deleteAll` calling this endpoint deletes all records in the Sepana db, and has been used by testing.
>     
>     * **TODO:** this endpoint should be protected, and should support deleting single entities. There's no clear need for this, but may become necessary if there are any future complications with the db.

@peripheralist not sure if this comment is up to date, but this is already managed by permissioned API keys — the delete endpoint can only be called if using a key with admin ACH permissions, and I'm assuming we'll be using one with read + write for production.

peripheralist:
> > * `/api/sepana/deleteAll` calling this endpoint deletes all records in the Sepana db, and has been used by testing.
> >   
> >   * **TODO:** this endpoint should be protected, and should support deleting single entities. There's no clear need for this, but may become necessary if there are any future complications with the db.
> 
> @peripheralist not sure if this comment is up to date, but this is already managed by permissioned API keys — the delete endpoint can only be called if using a key with admin ACH permissions, and I'm assuming we'll be using one with read + write for production.

thanks filip, this is a bit out of date

we'll be using API keys with specific read/write/admin permissions for different API calls as needed. there's no plan to include an admin key in production, so while this endpoint will be live it won't work (the benefit is that we'll have to to run on private builds with an admin key if it's ever needed)

since i missed your earlier message, the API keys will be private—only our Next server will have access to the keys, and all sepana calls will be routed through there. as you prolly know now we are using Upstash to run scheduled jobs, and i've also cleaned up and tricked out the /health endpoint. thanks for getting it started 🙏

src/pages/api/sepana/utils.ts
```diff
@@ -0,0 +1,250 @@
...
+}
+
+/**
+ * Search all Sepana project records, using preconfigured elastic search field weights and script source.
+ *
+ * @param query text to search
+ * @param pageSize max number of results to return
+ * @returns Promise containing project docs from Sepana database matching search params. If no query text is supplied, returns all projects
+ */
+export async function searchSepanaProjects(query = '', pageSize?: number) {
```
peripheralist, line 67: this is where sepana "default search logic" is defined. note that we're including matches to name, handle, and description, but weighting name + handle higher. we're also prioritizing results from those matches that have a higher volume or trending score.

.example.env
```diff
@@ -44,4 +44,10 @@ PRE_RENDER_INFURA_ID=
 INFURA_IPFS_PROJECT_ID=
 INFURA_IPFS_API_SECRET=
 
-BEEHIV_API_KEY=
\ No newline at end of file
+BEEHIV_API_KEY=
```
tomquirk, line 47: we can kill this, mb

```suggestion
```

src/pages/projects/AllProjects.tsx
```diff
@@ -36,9 +42,20 @@ export default function AllProjects({
...
+    useSepanaProjectsSearch(searchText)
+
+  const { data: graphSearchResults, isLoading: isLoadingGraphSearch } =
     useProjectsSearch(searchText)
 
+  const sepanaEnabled = featureFlagEnabled(FEATURE_FLAGS.SEPANA_SEARCH)
+
+  const searchResults = sepanaEnabled ? sepanaSearchResults : graphSearchResults
+
+  const isLoadingSearch = sepanaEnabled
```
tomquirk, line 55: **suggestion** it looks like this'll do 2 requests even though the feature flag means we only need data from 1 of them. Is this intentional? (I can see how it might be, cos then we get some in-prod testing on the api requests and queries, without the user seeing them).

If it's not, WDYT about passing an `enabled` flag to each of useSepanaProjectsSearch and useProjectsSearch, and enabling them depending on feature flag state?

src/pages/api/nextjs/revalidate-project.page.ts
```diff
@@ -41,6 +42,10 @@ export default async function handler(
   console.info(`[Next.js] Revalidating ${path}`)
   await res.revalidate(path)
 
+  // Update sepana database whenever a project needs revalidating
+  // Note: This won't work if it takes too long for project data to become available in the subgraph
+  await axios.get('/api/sepana/update')
```
tomquirk, line 47: In that case, is it worth `try-catching` this? or is it preferred to let it error?

src/pages/api/sepana/deleteAll.page.ts
```diff
@@ -0,0 +1,29 @@
...
+  try {
+    await deleteAllSepanaDocs()
+  } catch (error) {
+    await sepanaAlert({
+      type: 'alert',
+      alert: 'DB_UPDATE_ERROR',
+      body: JSON.stringify(error),
+    })
+
+    res.status(500).json({
```
tomquirk, line 14: **question** does this need a `return` statement?

src/hooks/Projects.ts
```diff
@@ -141,6 +148,36 @@ export function useProjectsSearch(handle: string | undefined) {
...
+    ['sepana-query', text, pageSize],
+    () =>
+      axios
+        .get<SepanaSearchResponse<SepanaProjectJson>>(
+          `/api/sepana/projects?text=${text}${
+            pageSize !== undefined ? `&pageSize=${pageSize}` : ''
+          }`,
+        )
+        .then(res =>
+          res.data.hits.hits.map(
```
tomquirk, line 171: **question** is it possible that any of this data could be undefined? Is it worth optional-chaining here?

src/hooks/Projects.ts
```diff
@@ -141,6 +148,36 @@ export function useProjectsSearch(handle: string | undefined) {
...
+  text: string | undefined,
+  pageSize?: number,
+) {
+  return useQuery(
+    ['sepana-query', text, pageSize],
+    () =>
+      axios
+        .get<SepanaSearchResponse<SepanaProjectJson>>(
+          `/api/sepana/projects?text=${text}${
+            pageSize !== undefined ? `&pageSize=${pageSize}` : ''
```
tomquirk, line 167: **suggestion** `qs` is useful here! https://github.com/ljharb/qs

src/lib/api/ipfs.ts
```diff
@@ -81,24 +81,25 @@ export const editMetadataForCid = async (
...
     responseType: 'json',
     headers: {
       Accept: 'application/json',
       'Content-Type': 'application/json',
     },
   })
-  if (response.data.Data?.['/'].bytes) {
-    response.data = extractJsonFromBase64Data(response.data.Data['/'].bytes)
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  if ((response.data as any).Data?.['/'].bytes) {
```
tomquirk, line 98: **question** is `response.data` of type `T` now? Is there a way to avoid `any`?

src/pages/api/sepana/delete.page.ts
```diff
@@ -0,0 +1,45 @@
...
+  try {
+    await deleteSepanaDoc(id).then(async _res => {
+      await sepanaAlert({
+        type: 'alert',
+        alert: 'DELETED_RECORD',
+        body: `ID: ${id}`,
+      })
+
+      res.status(200).json(_res.data)
+    })
```
tomquirk, line 22: **suggestion** think we can avoid nested `then` here

```suggestion
    const _res = await deleteSepanaDoc(id)
    await sepanaAlert({
      type: 'alert',
      alert: 'DELETED_RECORD',
      body: `ID: ${id}`,
    })

    return res.status(200).json(_res.data)
    })
```

src/pages/api/sepana/delete.page.ts
```diff
@@ -0,0 +1,45 @@
...
+  } catch (error) {
+    await sepanaAlert({
+      type: 'alert',
+      alert: 'DB_UPDATE_ERROR',
+      body: JSON.stringify(error),
+    })
+
+    res.status(500).json({
```
tomquirk, line 30: **suggestion** i think this needs a return

src/pages/api/sepana/utils.ts
```diff
@@ -0,0 +1,250 @@
...
+}
+
+/**
+ * Exhaustively queries all Sepana records.
+ *
+ * @returns Promise containing all project docs from Sepana database
+ */
+export async function queryAllSepanaProjects() {
+  return axios.post<SepanaSearchResponse<SepanaProject>>(
+    `${process.env.NEXT_PUBLIC_SEPANA_API_URL}/search`,
```
tomquirk, line 45: **suggestion** we can create an axiosInstance that has the `baseUrl` set, which will save having to re-add it every time in these utils.

See https://github.com/jbx-protocol/juice-interface/blob/51384d5ed01e903d49805236c4d3b7e3b28e6d64/src/lib/safe.ts#L14-L17 for example

src/pages/api/sepana/utils.ts
```diff
@@ -0,0 +1,250 @@
+import axios from 'axios'
```
tomquirk, line 1: **suggestion** this file might be better placed in `src/lib/sepana.ts` WDYT?

tomquirk (changes requested):
@peripheralist super exciting and thorough work, amazing to see!

I've left a few comments but still haven't gotten to the whole PR yet. 

(P.s. In future it would be helpful to break this work up into multiple PRs to help the code review 🙏 )

src/pages/api/nextjs/revalidate-project.page.ts
peripheralist: it won't _error_, just won't actually write any new data. comment could be more specific

src/pages/api/sepana/deleteAll.page.ts
peripheralist: handlers should be void afaik, important thing is ending the request

src/hooks/Projects.ts
peripheralist: doesn't seem like it, but down to error on the side of caution here

src/lib/api/ipfs.ts
peripheralist: seems like this is something we can remove (cc @wraeth-eth) so i've left it in awkward state. but yes let's plz avoid `any`

src/pages/api/sepana/delete.page.ts
peripheralist: bet

src/pages/api/sepana/delete.page.ts
peripheralist: doesn't look like handlers need to return a value https://nextjs.org/docs/api-routes/response-helpers

src/pages/api/sepana/utils.ts
peripheralist: luv

src/pages/api/sepana/utils.ts
peripheralist: yes, i've just done a bunch of restructuring in another branch to add /lib/sepana

https://github.com/jbx-protocol/juice-interface/pull/2884/files

i'll try putting the same changes in this PR

src/pages/projects/AllProjects.tsx
peripheralist: yeah, currently making 2 queries. i could add a clause in each hook like `if (x) don't fetch`, but i think it'd be useless outside of this use case

i'll see if i can find a conditional that makes sense to avoid double query

src/pages/api/sepana/delete.page.ts
tomquirk: @peripheralist even if we `catch` here, the following `sepanaAlert` on line 37 will still execute right? Is this expected?

tomquirk: even though we've given a response, i believe the function will continue executing as normal.

src/pages/projects/AllProjects.tsx
tomquirk: useQuery has an `enabled` flag which we could use:

https://github.com/jbx-protocol/juice-interface/blob/71bef62f51d00ccb61573d4e7a0ce0c1aeefc6c4/src/hooks/veNft/VeNftContractForProject.ts#L19

tomquirk (commented):
@peripheralist one other thing. Is it an issue that these API routes are public? I.e. someone can call delete all on the sepana db?

src/pages/projects/AllProjects.tsx
peripheralist: ohh, perfect

peripheralist:
> @peripheralist one other thing. Is it an issue that these API routes are public? I.e. someone can call delete all on the sepana db?

Put this in the PR description:
> *NOTE: Delete endpoints require a sepana admin key to be defined in .env (not included in .example.env). So while these endpoints will be live, they won't work unless the admin key is included in production. Motivation to include them is so that if they're needed, we can easily call them from a private build of the app (there's currently no delete UI in the Sepana db console)

this seems okay to me, but also happy to take some extra precautionary steps. worth noting that deleting anything from the database would be a very minor issue. it'd just get repopulated at the next update. as a general rule i don't think we should be using sepana as any source of truth—just a convenience layer

src/pages/api/sepana/delete.page.ts
```diff
@@ -0,0 +1,45 @@
+import { deleteRecord } from 'lib/sepana'
```
wraeth-eth, line 1: These APIs are all publicly facing - do we want to put some auth on them?

src/pages/api/sepana/update.page.ts
```diff
@@ -0,0 +1,230 @@
...
+            project: {
+              ...p,
+              lastUpdated,
+              metadataResolved: false, // If there is an error resolving metadata from IPFS, we'll flag it as `metadataResolved: false`. We'll try getting it again whenever `retryIPFS == true`.
+            } as SepanaProjectJson,
+          })),
+      )
+    }
+
+    const promiseResults = await Promise.all(promises)
```
wraeth-eth, line 147: nit - could consider using `allSettled` to get a list of success and fails rather than the catch

src/pages/api/nextjs/revalidate-project.page.ts
wraeth-eth: From what I read recently, serverless nextjs api should be quick. How long does update usually take? I think we have a cap of 60s according to:
https://vercel.com/docs/concepts/limits/overview#general-limits

src/lib/api/ipfs.ts
wraeth-eth: This was to get around all the ipfs chicanery intro'd by a bug in pinata ipfs upgrade. I think we should open an issue to investigate it (Don't mind doing it) and leave here for now

wraeth-eth: https://github.com/jbx-protocol/juice-interface/issues/2886

src/lib/sepana/index.ts
```diff
@@ -0,0 +1,122 @@
+import { SepanaProjectJson, SepanaQueryResponse } from 'models/sepana'
```
wraeth-eth, line 1: nit - we should try and avoid in index files and use them as barrel-files I think. It makes fuzzy searching a lot simpler

src/lib/sepana/index.ts
```diff
@@ -0,0 +1,122 @@
...
+ */
+export async function queryAll<T extends object>() {
+  return sepanaAxios('read').post<SepanaQueryResponse<T>>(
+    SEPANA_ENDPOINTS.search,
+    {
+      engine_ids: [process.env.SEPANA_ENGINE_ID],
+      query: {
+        match_all: {},
+      },
+      size: 10000, // TODO: Update this before we hit 10k projects
```
wraeth-eth, line 19: We should make a ticket to track this

src/components/QuickProjectSearch.tsx
```diff
@@ -201,28 +219,34 @@ export default function QuickProjectSearch() {
...
-            {searchText && !isLoadingSearch && searchPages?.length === 0 && (
+            {searchText && !isLoadingSearch && searchResults?.length === 0 && (
               <div className="text-center text-grey-400">
                 <Trans>No results</Trans>
               </div>
             )}
           </div>
 
-          <div className="mt-5 flex gap-6 border-t border-r-0 border-b-0 border-l-0 border-solid border-t-smoke-200 bg-smoke-75 py-4 px-5 text-xs dark:border-t-slate-300 dark:bg-slate-600">
+          <div className="mt-5 flex gap-6 bg-smoke-75 py-3 px-5 text-xs dark:border-t-slate-300 dark:bg-slate-600">
```
wraeth-eth, line 249: I think we still want the border here, wdyt? I feel like the colors blend in a weird way without it

**Without**
Screen Shot 2023-01-27 at 12 55 39 pm

**With**
Screen Shot 2023-01-27 at 12 55 35 pm

wraeth-eth (changes requested):
LFG @peripheralist!! ❤️❤️❤️ this!!

I might need some help setting up env so I can test this locally. 🙏 Have left some comments. LMKWYT

src/pages/api/sepana/delete.page.ts
peripheralist: oh i see what you mean now. actually, there shouldn't be any clause after the try/catch, was an error in copying over some code from DeleteAll. fixing

src/pages/api/sepana/delete.page.ts
peripheralist: they are all authenticated using private server keys. there's no harm in calling any except the delete endpoints, which require an admin key that won't be included in the public build (and isn't in .example.env)

i'm open to some added security layer, but worth noting that even deleting the entire DB would hardly be a problem—it would just get repopulated within a few minutes the next time /update gets called, and queries would be broken until then. sepana shouldn't be a source of truth for anything

peripheralist:
> From what I read recently, serverless nextjs api should be quick. How long does update usually take? I think we have a cap of 60s according to:
https://vercel.com/docs/concepts/limits/overview#general-limits

@wraeth-eth no call should ever take more than a few seconds (though populating the db from empty might take 20-30). the line you referenced just meant that if the data isn't available, the function won't find any new data and won't put it in the db. the call won't wait for anything.

src/lib/sepana/index.ts
peripheralist: honestly agree

src/components/QuickProjectSearch.tsx
peripheralist: oh i'm not sure i meant to do this. though fwiw i'm team no-border, fewer lines look cleaner imo

src/pages/api/sepana/update.page.ts
peripheralist: good suggestion, though i need to actually use the catch to store different data. in the case that resolving metadata fails, we store what we do have along with a `metadataResolved: false` flag

peripheralist:
@wraeth-eth getting set up should be as easy as dropping in the api keys (you can get from sepana dashboard or discord resources) and enabling the feature flag. lmk if you get stuck anywhere

peripheralist:
@tomquirk @wraeth-eth thank you both for solid feedback, all addressed 🙏

src/pages/api/nextjs/revalidate-project.page.ts
tomquirk: Great point. maybe worth not even `await`ing this at all? just fire it off?

tomquirk:
@peripheralist we're only calling these endpoints from upstash right?

In that case, we could create a super basic auth for these endpoints.

1. Create a key (cryptographically good etc). This will be the API key used to auth with these endpoints.
2. Set the key as an env var in JBM
3. Create a Next Middleware to validate requests. This will check Authorization header of requests, and check its value matches the env var (our api key)
4. Provide this API key to upstash. If they let us set headers on requests, we can add it there.

I think it might worth doing, just for sake of completeness.

cc @wraeth-eth

peripheralist:
@tomquirk upstash will only be calling /update and /health, and there isn't any risk to calling either

the only endpoint that needs to be protected is delete. the plan is to just not include an admin api key in production, which will disable it

there's no expectation of needing the endpoint under normal circumstances. including it in the app is just a convenience for us in case something goes wrong, and it can be used via a local/private build that includes an admin key

using another key to authenticate with the delete endpoint would be the same level of security we have now with double the risk vector, since compromising either key (stored as a private server var) would compromise the db

tomquirk:
> the plan is to just not include an admin api key in production, which will disable it

@peripheralist ah got it! That clears it up, thanks. If it's not a prod endpoint then perhaps we play it extra-safe and add a `NODE_ENV` check in the route handler (just in case a future dev adds the sepena admin key to prod, thereby  inadvertently 'enabling' public delete functionality).

As an alt suggestion, could we instead create a `scripts/` directory and add this functionality in a script there, instead of as a Nextjs route?

src/lib/sepana/api/http.ts
```diff
@@ -0,0 +1,39 @@
+import axios from 'axios'
```
tomquirk, line 1: **massive praise** for this file

tomquirk (approved):
@peripheralist im stoked to approve this. Thanks a bunch for taking on the feedback. Will leave it to @wraeth-eth for any final comments

peripheralist:
@tomquirk putting delete stuff in /scripts is a great call, will do

also have a few more low key improvements to push for the update endpoint

src/pages/api/sepana/delete.page.ts
wraeth-eth: Sounds GTM!

wraeth-eth (approved):

src/lib/api/ipfs.ts
wraeth-eth: Just had a look; this is still a problem. We'll have to wait for pinata to fix it (if they ever do)

src/pages/api/nextjs/revalidate-project.page.ts
peripheralist: sry thought i responded to this. no endpoints should take more than a couple seconds (even populating the db from empty takes less than 10s, but that'll likely won't ever happen in production)

PR #2739, opened 2023-01-06
Fix payout beneficiary bug

tomquirk:


tomquirk:
cc @johnnyd-eth is tehre anywhere else we need to check? if so pls open PR.

PR #2787, opened 2023-01-16
Fix logo not rendering in project settings

### LINKED ISSUE #2774: Broken image logo on project settings page
## Summary

When viewing the project settings page, existing project logos are appearing as broken images. This issue is preventing users from correctly viewing and managing their project logos.

## Steps to reproduce

1. Navigate to the project settings page
2. Observe that existing project logos are appearing as broken images

## What is the expected _correct_ behavior?

Project logos should be properly displayed on the project settings page.

## Relevant logs and/or screenshots

image

## Discord Link

https://discord.com/channels/775859454780244028/866040669712678942/1063957476480913552

johnnyd-eth:

Closes #2774 

Screen Shot 2023-01-16 at 3 06 13 pm

## Acceptance checklist

### DIFF
--- a/src/components/forms/ProjectDetailsForm.tsx
+++ b/src/components/forms/ProjectDetailsForm.tsx
@@ -1,5 +1,6 @@
 import { t, Trans } from '@lingui/macro'
 import { Button, Form, FormInstance, Space } from 'antd'
+import { useWatch } from 'antd/lib/form/Form'
 import { FormItems } from 'components/formItems'
 import { MinimalCollapse } from 'components/MinimalCollapse'
 import { normalizeHandle } from 'utils/format/formatHandle'
@@ -31,6 +32,8 @@ export default function ProjectDetailsForm({
   loading?: boolean
   onValuesChange?: VoidFunction
 }) {
+  const initialLogoUrl = useWatch('logoUri', form)
+
   return (
     <Form
       scrollToFirstError={{ behavior: 'smooth' }}
@@ -67,7 +70,7 @@ export default function ProjectDetailsForm({
           <FormItems.ProjectDescription name="description" />
           <FormItems.ProjectLogoUri
             name="logoUri"
-            initialUrl={form.getFieldValue('logoUri')}
+            initialUrl={initialLogoUrl}
             onSuccess={logoUri => {
               form.setFieldsValue({ logoUri })
               onValuesChange?.()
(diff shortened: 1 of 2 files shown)


src/components/forms/ProjectDetailsForm.tsx
```diff
@@ -31,6 +32,8 @@ export default function ProjectDetailsForm({
   loading?: boolean
   onValuesChange?: VoidFunction
 }) {
+  const initialLogoUrl = useWatch('logoUri', form)
```
johnnyd-eth, line 35: Also noticed sometimes the logo wasn't loading at all and it appeared like this on initial load:

Screen Shot 2023-01-16 at 3 08 19 pm

This fixes that.

src/components/forms/ProjectDetailsForm.tsx
```diff
@@ -67,7 +70,7 @@ export default function ProjectDetailsForm({
           <FormItems.ProjectDescription name="description" />
           <FormItems.ProjectLogoUri
             name="logoUri"
-            initialUrl={form.getFieldValue('logoUri')}
+            initialUrl={initialLogoUrl}
```
wraeth-eth, line 73: We should probably fix the `ProjectLogoUri` to load its `value` (assuming that is the bug here)

johnnyd-eth: The main bug here was that we weren't converting the `ipfs://` url to `https://` to render the image, this is just something additional that I noticed occasionally. 

`ProjectLogoUri` uses the `ImageUploader` component which uses states. I could be wrong, but I thought this means we would still be required to pass in the initial value as a prop.

wraeth-eth: I'm not familiar with the code here, but when using antd form stuff, all the state should be handled by the form (so the initial values passed into the form. These get trickled down into the form.item with `value`. As long as `value` and `onChange` exist and are as expected per antd, it should be g2g

johnnyd-eth: Yeah it would be great if this was the case, but as it stands now, I think this would require overhauling the `ImageUploader` component, which I think is outside the scope of this issue. Happy to create an issue to improve this component in the future, but don't think it's necessary for this PR. WDYT?

tomquirk (approved):

Issue #2815, opened 2023-01-18
When you have an NFT sticker in a memo, the normal stickers are deleted


NakedFool:
## Summary

If you are purchasing NFT and proceed to payment, in the memo if you try to add regular stickers aside NFT sticker, they will get erased. sometimes instantly and sometimes after a few moments, depending on when the component will re-render.

## Relevant logs and/or screenshots

https://user-images.githubusercontent.com/18723426/213117656-ba16e457-88a7-4347-a8ab-6638bf08791d.mov

## Discord link
https://discord.com/channels/775859454780244028/1065180480913276948/1065180484528787456

Issue #2773, opened 2023-01-15
Payment modal clearing memo and deactivating token beneficiary on rejected pay tx


wraeth-eth:
## Summary

When rejecting a pay transaction in the wallet (for example, to revise the value), the payment modal is currently clearing all memo information and deactivating the token beneficiary. This behavior is not expected and should be fixed.

## Steps to reproduce

1. Open the pay modal
2. Enter memo and token beneficiary information
3. Initiate transaction
4. Click on the "Reject" button to revise the transaction value
5. **Observe error behaviour** (the memo and token beneficiary information is cleared)

## What is the expected _correct_ behavior?
Memo and token beneficiary information should be preserved and remain active even after a rejected pay transaction.

## Relevant logs and/or screenshots
![image](https://user-images.githubusercontent.com/104132113/212571731-4bff1abf-cea7-47c4-8674-d94ef5320672.png)

## Discord Link
https://discord.com/channels/775859454780244028/866040669712678942/1063937287035035750

PR #2817, opened 2023-01-18
Fix memo saving on transaction cancel

NakedFool:
Closes https://github.com/jbx-protocol/juice-interface/issues/2773
Closes https://github.com/jbx-protocol/juice-interface/issues/2815

The issues were discovered while working on each other and they are related. The main fix is that memo data and stickerUrls are now properly saved when a transaction is canceled. Additionally, the radio button that toggles the beneficiary address is now also saved.

The PR also addresses an issue with the inclusion of NFT rewards in the initial stickerUrls array by:

- passing the nftRewardTiers as state to the V2V3PayForm component to prevent re-rendering.

- in the V2V3PayForm hook, it checks if the form already has values. If the form does not have values, it proceeds to set the initialized NFT stickers. If the form already has values, the function exits and the existing values are preserved, allowing them to be saved when the user cancels the transaction.

https://user-images.githubusercontent.com/18723426/213181664-b62fdf86-ebb5-4c20-92ad-a0355d29095c.mov

## Acceptance checklist


tomquirk:
@wraeth-eth mind reviewing?

wraeth-eth:
Hey @NakedFool thanks for the PR. This still has a problem where the memo is empty ("") after clicking pay after rejecting once.

Steps to repeat
1. Open a pay modal
2. Add some stickers to the memo
3. Click Pay
4. Note that the transaction is properly built in the console output
5. Cancel Transaction
6. Immediately click pay again
7. Note that the transaction memo is empty ("") in console output

wraeth-eth:
FWIW I think we should try to limit adding more `useEffect` where possible as these initial bugs are likely due to `useEffect`. It is probably out of scope for this though. If you can get it working with its current solution. If not, I'd recommend looking to incrementally pare back the useEffects used in this modal (while retaining same behavior), then fix the bugs.

wraeth-eth:
@NakedFool I was pretty vague, lemme know if you wanna try and organise some time in cross over to chat about it further if you want!

NakedFool:
> 

Hmmm, how so ser? I've just tried to do it one more time again, attaching a video, soz its is a bit long ~ 1min 🍿 


https://user-images.githubusercontent.com/18723426/214235807-a987c66f-7bba-4d80-8a61-efd06f99099b.mov

NakedFool:
> Hey @NakedFool thanks for the PR. This still has a problem where the memo is empty ("") after clicking pay after rejecting once.
> 
> Steps to repeat
> 
> 1. Open a pay modal
> 2. Add some stickers to the memo
> 3. Click Pay
> 4. Note that the transaction is properly built in the console output
> 5. Cancel Transaction
> 6. Immediately click pay again
> 7. Note that the transaction memo is empty ("") in console output

Hmmm, how so ser? I've just tried to do it one more time again, attaching a video, soz its is a bit long ~ 1min 🍿


https://user-images.githubusercontent.com/18723426/214236279-f6a7463f-875c-424c-9323-fc9a7b3bd220.mov

NakedFool:
> FWIW I think we should try to limit adding more `useEffect` where possible as these initial bugs are likely due to `useEffect`. It is probably out of scope for this though. If you can get it working with its current solution. If not, I'd recommend looking to incrementally pare back the useEffects used in this modal (while retaining same behavior), then fix the bugs.

Yeah, I feel you.

In general, I agree with you that we should utilize as less effects as possible because of their risky behaviors and less error-prone situations in the future, but I think this case was specific and it deserves more attention.

I think the problem here is `V2V3PayForm` component since we are passing `nftRewardTiers` from there, and in a real scenario, that initial useEffect in `V2V3ConfirmPayModal` should only run once when `nftRewardTiers` is passed, but it was re-rendering constantly before due to what I think the way of defining it in `V2V3PayForm`.

Thank you for this review, I'm really down to organize some time and go over this quickly together! 🍻

wraeth-eth:
> > Hey @NakedFool thanks for the PR. This still has a problem where the memo is empty ("") after clicking pay after rejecting once.
> > Steps to repeat
> > 
> > 1. Open a pay modal
> > 2. Add some stickers to the memo
> > 3. Click Pay
> > 4. Note that the transaction is properly built in the console output
> > 5. Cancel Transaction
> > 6. Immediately click pay again
> > 7. Note that the transaction memo is empty ("") in console output
> 
> Hmmm, how so ser? I've just tried to do it one more time again, attaching a video, soz its is a bit long ~ 1min 🍿
> 
>  screen-recorder-tue-jan-24-2023-08-28-59.mov

Ah I know the difference, make sure you don't have an NFT. This also creates some problems if you remove an NFT too.


https://user-images.githubusercontent.com/104132113/214972415-a34aff36-21bf-44b1-b815-5685b78be3a2.mov

NakedFool:
> > > Hey @NakedFool thanks for the PR. This still has a problem where the memo is empty ("") after clicking pay after rejecting once.
> > > Steps to repeat
> > > 
> > > 1. Open a pay modal
> > > 2. Add some stickers to the memo
> > > 3. Click Pay
> > > 4. Note that the transaction is properly built in the console output
> > > 5. Cancel Transaction
> > > 6. Immediately click pay again
> > > 7. Note that the transaction memo is empty ("") in console output
> > 
> > 
> > Hmmm, how so ser? I've just tried to do it one more time again, attaching a video, soz its is a bit long ~ 1min 🍿
> > screen-recorder-tue-jan-24-2023-08-28-59.mov
> 
> Ah I know the difference, make sure you don't have an NFT. This also creates some problems if you remove an NFT too.
> 
>  Untitled.mov

Oo gotcha, I think I've got it working all correctly now, mainly operating on `transactionCanceled` value which is called when a user cancels the transaction.


https://user-images.githubusercontent.com/18723426/215486754-4f8e92d6-0e68-489e-907a-b1030844440c.mov

wraeth-eth (approved):
Approving - still highly suspect on all the useEffect we have near this code though 😬

wraeth-eth:
Found another bug - but it already existed. Will make a new ticket for it.

Good work on this one @NakedFool !

Issue #2835, opened 2023-01-20
Payout to a V1->V3 allocator uses V1 project ID in activity feed


tomquirk:
<img width="471" alt="Screenshot 2023-01-20 at 2 42 44 PM" src="https://user-images.githubusercontent.com/12551741/213615885-95dfe843-facf-43e3-a8d9-720e66c286c5.png">

Issue #2791, opened 2023-01-16
Project handle color is different to eth address color in activity feed


NakedFool:
## Summary

In v1 colors are different, while v3 is fine.

## Relevant logs and/or screenshots
![image](https://user-images.githubusercontent.com/18723426/212687983-5241d8e5-4ff2-4725-be45-c25fe935edf5.png)

## Discord link
https://discord.com/channels/775859454780244028/1064375696622895104/1064375700229988463

PR #2868, opened 2023-01-23
Assign correct project in v1 activity feed

NakedFool:

Closes https://github.com/jbx-protocol/juice-interface/issues/2835

Closes https://github.com/jbx-protocol/juice-interface/issues/2791

I've added `modAllocator` in the keys query, additionally, I've created a function that checks if `allocator` is the same as constant `V1_V3_ALLOCATOR_ADDRESS`

image

## Acceptance checklist

### DIFF
--- a/src/components/v1/V1Project/ProjectActivity/eventElems/TapEventElem.tsx
+++ b/src/components/v1/V1Project/ProjectActivity/eventElems/TapEventElem.tsx
@@ -3,6 +3,8 @@ import { ActivityEvent } from 'components/activityEventElems/ActivityElement'
 import ETHAmount from 'components/currency/ETHAmount'
 import FormattedAddress from 'components/FormattedAddress'
 import V1ProjectHandle from 'components/v1/shared/V1ProjectHandle'
+import V2V3ProjectHandleLink from 'components/v2v3/shared/V2V3ProjectHandleLink'
+import { V1_V3_ALLOCATOR_ADDRESS } from 'constants/contracts/mainnet/Allocators'
 import { ThemeContext } from 'contexts/themeContext'
 import useSubgraphQuery from 'hooks/SubgraphQuery'
 import { TapEvent } from 'models/subgraph-entities/v1/tap-event'
@@ -40,6 +42,7 @@ export default function TapEventElem({
             'modProjectId',
             'modBeneficiary',
             'modCut',
+            'modAllocator',
             {
               entity: 'tapEvent',
               keys: ['id'],
@@ -56,6 +59,13 @@ export default function TapEventElem({
   )
 
   if (!event) return null
+
+  const handleProject = (projectId: number, allocator: string) => {
+    if (allocator.toLowerCase() === V1_V3_ALLOCATOR_ADDRESS.toLowerCase()) {
+      return <V2V3ProjectHandleLink projectId={projectId} />
+    }
+    return <V1ProjectHandle projectId={projectId} />
+  }
 
   return (
     <ActivityEvent
@@ -80,9 +90,7 @@ export default function TapEventElem({
             >
               <div style={{ fontWeight: 500 }}>
                 {e.modProjectId > 0 ? (
-                  <span>
-                    <V1ProjectHandle projectId={e.modProjectId} />
-                  </span>
+                  <span>{handleProject(e.modProjectId, e.modAllocator)}</span>
                 ) : (
                   <FormattedAddress address={e.modBeneficiary} />
                 )}
--- a/src/components/v1/V1Project/ProjectActivity/eventElems/TapEventElem.tsx
+++ b/src/components/v1/V1Project/ProjectActivity/eventElems/TapEventElem.tsx
@@ -3,9 +3,10 @@ import { ActivityEvent } from 'components/activityEventElems/ActivityElement'
 import ETHAmount from 'components/currency/ETHAmount'
 import FormattedAddress from 'components/FormattedAddress'
 import V1ProjectHandle from 'components/v1/shared/V1ProjectHandle'
-import V2V3ProjectHandleLink from 'components/v2v3/shared/V2V3ProjectHandleLink'
+import V2V3ProjectLink from 'components/v2v3/shared/V2V3ProjectLink'
 import { V1_V3_ALLOCATOR_ADDRESS } from 'constants/contracts/mainnet/Allocators'
 import { ThemeContext } from 'contexts/themeContext'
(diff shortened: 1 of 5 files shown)


src/components/v1/V1Project/ProjectActivity/eventElems/TapEventElem.tsx
```diff
@@ -57,6 +60,13 @@ export default function TapEventElem({
 
   if (!event) return null
 
+  const handleProject = (projectId: number, allocator: string) => {
+    if (allocator.toLowerCase() === V1_V3_ALLOCATOR_ADDRESS.toLowerCase()) {
+      return <V2V3ProjectHandleLink projectId={projectId} />
+    }
+    return <V1ProjectHandle projectId={projectId} />
+  }
```
tomquirk, line 68: **suggestion** defining render functions like this within other components ("closures") can lead to rendering bugs. Could we move this outside of the `TapEventElem` component scope?

src/components/v1/V1Project/ProjectActivity/eventElems/TapEventElem.tsx
```diff
@@ -57,6 +60,13 @@ export default function TapEventElem({
 
   if (!event) return null
 
+  const handleProject = (projectId: number, allocator: string) => {
+    if (allocator.toLowerCase() === V1_V3_ALLOCATOR_ADDRESS.toLowerCase()) {
+      return <V2V3ProjectHandleLink projectId={projectId} />
```
tomquirk, line 65: **question** This `V2V3ProjectHandleLink` component relies on `V2V3ContractContext`, which I don't think we provide for V1 projects. Not sure how this can work (but I may be missing something!)

src/components/v1/V1Project/ProjectActivity/eventElems/TapEventElem.tsx
```diff
@@ -57,6 +60,13 @@ export default function TapEventElem({
 
   if (!event) return null
 
+  const handleProject = (projectId: number, allocator: string) => {
+    if (allocator.toLowerCase() === V1_V3_ALLOCATOR_ADDRESS.toLowerCase()) {
```
tomquirk, line 64: **suggestion** a slightly more 'correct' approach might be using ethers `getAddress` https://docs.ethers.org/v5/api/utils/address/#utils-getAddress

src/components/v1/V1Project/ProjectActivity/eventElems/TapEventElem.tsx
tomquirk: Alternatively, we could just render the Project ID and append the V3 version badge:
```
Project #69 <V3>
```

WDYT?

src/components/v1/V1Project/ProjectActivity/eventElems/TapEventElem.tsx
NakedFool: thanks for this tip, didn't know about this utility function!

src/components/v1/V1Project/ProjectActivity/eventElems/TapEventElem.tsx
NakedFool: It is working, I was trying to find where we pass provider but couldn't find it, though we are using the same component in `Mod` component in the v1 folder. should I do the same there or keep it as it is now? 
I can add a badge in any of these cases.

src/components/v1/V1Project/ProjectActivity/eventElems/TapEventElem.tsx
NakedFool: Agreed! 

Should we, in this case, do inline conditioning to avoid render functions, or I can move render function outside of the component scope?

tomquirk: i think inline sounds fine in this case! good call

src/components/v1/V1Project/ProjectActivity/eventElems/TapEventElem.tsx
tomquirk: One point of caution is that `getAddress` will throw if the input is invalid (can't remember the exact cases it fails on). Maybe we should add a `safeGetAddress` util...out of scope tho.

src/components/v1/V1Project/ProjectActivity/eventElems/TapEventElem.tsx
tomquirk: @NakedFool i don't think it's quite working as expected.

V2V3ProjectHandleLink uses `useProjectHandleText`, which uses `useProjectHandle`, which relies on V2V3ContractsContext. If I log `console.log(useContext(V2V3ContractsContext))` in `useProjectHandle`, we can see it is undefined (we're not providing it).

Therefore `useProjectHandle` will never actually return a handle. We can verify this by looking at the Peel V1 project. Our Peel V3 project has a handle `@peel`, so any payout to that project should render `@peel`. But the latest activity item has a distribution to our V3 project, and doesn't render our handle:

Screenshot 2023-01-25 at 11 54 15 AM

FWIW i think is probably okay, but I don't think we should be using the `V2V3ProjectHandleLink` component here. I think
lets render the Project ID and append the V3 version badge in a new component. Something like `V2V3ProjectLink`.

WDYT about this?

Screenshot 2023-01-25 at 12 01 26 PM

NakedFool: Oh oh oh, gotcha! 
Basically what you are saying is why to use `V2V3ProjectHandleLink` if we can never get the handle back for now.

Thanks very much for the thorough explanation, I've created a light component `V2V3ProjectLink` that requires `projectId` but `allocator` is optional, that way we can decide if we want to include the version badge or not.

I've also added it to `Mod` component to keep things consistent.

wraeth-eth:
@tomquirk are you still reviewing this one?

tomquirk:
Thanks @NakedFool this is looking great! I made another commit, just to add `Trans`, and also to address https://github.com/jbx-protocol/juice-interface/issues/2791 (was driving me nuts).


Lets merge 🚀 

P.s. when a PR is ready for a follow-up review, make sure to re-request a review in GitHub. Otherwise the review doesn't know when the PR is ready for review again and doesn't get notified.

Screenshot 2023-02-01 at 11 49 57 AM

tomquirk (approved):

Issue #2879, opened 2023-01-26
Token burning should call `burnTokensOf`


tomquirk:
## Summary

Currently calling `redeemTokensOf` when a user tries to burn their project tokens on JBM.

This is incorrect. We need to call `burnTokensOf`.

PR #2881, opened 2023-01-26
Use burnTokensOf for burn

tomquirk:
Closes https://github.com/jbx-protocol/juice-interface/issues/2879. Insead of calling `redeemTokensOf` in a burn scenario, we now call `burnTokensOf`


johnnyd-eth (approved):
Tested and code looks great.

PR #2882, opened 2023-01-26
Allow NFT upload mp4 in create flow

johnnyd-eth:

**- Architecture changes:**
- Changes all `imageUrl` and `imgUrl` in `NftRewardType`s to `fileUrl` (applies changes in all places necessary, where the type is being used).
- Feature flag to allow for NFT mp4 upload (`nftMp4`)

**- UI changes:**
Before:
Screen Shot 2023-01-26 at 5 17 13 pm

After:
Screen Shot 2023-01-26 at 5 16 39 pm

## Acceptance checklist


src/components/UploadNoStyle/UploadNoStyle.tsx
```diff
@@ -88,6 +104,34 @@ export const UploadNoStyle = (props: UploadNoStyleProps) => {
     [props, setUploadUrl],
   )
 
+  const _body = () => {
```
tomquirk, line 107: **issue** defining functions inside components causes rendering bugs. Need to move outside this component's scope, or make it a variable (not a function)

tomquirk: Also, why not just put this code directly inline?

tomquirk (commented):
@johnnyd-eth this LGTM but shouldn't we first devise a strategy for showing these on the:
- project page
- review and deploy screen

I think this refactor is good though. Maybe we can feature-flag the allowed content type for now?

tomquirk (approved):
@johnnyd-eth this LGTM! thanks for putting behind a feature flag.

Issue #2858, opened 2023-01-23
Retain loading state of "Save project details" button on deploying in Project Settings page


wraeth-eth:
## Summary

When a user clicks the "Save project details" button on the Project Settings page, the button enters a loading state while the changes are being deployed. This is a small UX improvement. It also disallows the user trying to redeploy.

## Steps to reproduce

1. Go to the Project Settings page
2. Make changes to the project details
3. Click the "Save project details" button
4. See the button return to clickable.

## What is the expected _correct_ behavior?

The loading state of the "Save project details" button should be retained until the deploy finishes

PR #2883, opened 2023-01-26
Fix button state when updating project details in settings

NakedFool:

Closes https://github.com/jbx-protocol/juice-interface/issues/2858

The bug here was that we were setting loading state to false in `onDone` callback, which executes immediately. We will catch it and return false in either case because I relocated it to the `onConfirmed` and added the `onError` callback as well.

The tx hook has also been updated to use our most recent approach with invariant and try & catch.

Some takeaways that I want to talk about in this PR:

1. The user might be confused by empty fields since they were just updated, therefore I'm not sure how good it is for us to clean form fields after tx is confirmed. Please let me know your thoughts; although I removed the form reset in this PR, I can simply add it back if you think that is the best course of action in this case or I may miss some more context here.
2. Project header is not updated after the user update project details, this can lead to another confusion, I've seen that it uses `ProjectMetadataContext` but I'm not sure how can we update it. Should I go on and update the context itself?

https://user-images.githubusercontent.com/18723426/214834366-dc6dee1e-6972-4efa-9cb2-242e6465ad53.mov

## Acceptance checklist

### DIFF
--- a/src/components/Create/components/pages/ReviewDeploy/components/DeploySuccess/DeploySuccess.tsx
+++ b/src/components/Create/components/pages/ReviewDeploy/components/DeploySuccess/DeploySuccess.tsx
@@ -1,20 +1,17 @@
-import { ShareAltOutlined } from '@ant-design/icons'
+import { TwitterOutlined, ShareAltOutlined } from '@ant-design/icons'
 import { t, Trans } from '@lingui/macro'
-import { Button, Space } from 'antd'
+import { Button } from 'antd'
 import ExternalLink from 'components/ExternalLink'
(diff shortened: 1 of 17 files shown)


NakedFool:
> @NakedFool this looks fantastic!
> 
> WDYT about refreshing the page when the tx is complete? It's a brute-force solution but as you said, its a bit confusing to the user currently!

@wraeth-eth created a PR in the meantime, that PR is going to be responsible for refetching, gonna rebase on this once it's merged into main.

NakedFool:
Closing this PR; re-opened in https://github.com/jbx-protocol/juice-interface/pull/2890

PR #2930, opened 2023-02-01
[2/n] Move datasource flag

tomquirk:
## What does this PR do and why?

⚠️ depends on https://github.com/jbx-protocol/juice-interface/pull/2922

- Move "Use data source for redeem" flag to NFT page
- Reworks the copy of this to be NFT-centric

### NFT page
Screenshot 2023-02-01 at 12 28 01 PM

### Rules page
Screenshot 2023-02-01 at 12 28 10 PM

### Review section

Screenshot 2023-02-01 at 12 29 35 PM

PR #2932, opened 2023-02-01
[3/n] Add AllowSetController to create flow

tomquirk:

⚠️ depends on https://github.com/jbx-protocol/juice-interface/pull/2930

Adds yet another metadata toggle. Reconfig UI will come in the next PR.

### Form
Screenshot 2023-02-01 at 3 15 01 PM

### Review

Screenshot 2023-02-01 at 3 15 06 PM

## Acceptance checklist

### DIFF
--- a/src/components/Create/components/pages/ReviewDeploy/components/RulesReview/RulesReview.tsx
+++ b/src/components/Create/components/pages/ReviewDeploy/components/RulesReview/RulesReview.tsx
@@ -1,5 +1,4 @@
 import { t } from '@lingui/macro'
-import { Row } from 'antd'
 import FormattedAddress from 'components/FormattedAddress'
 import useMobile from 'hooks/Mobile'
 import { DescriptionCol } from '../DescriptionCol'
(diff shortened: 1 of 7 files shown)


src/components/Create/components/pages/ReviewDeploy/components/RulesReview/RulesReview.tsx
```diff
@@ -21,7 +22,7 @@ export const RulesReview = () => {
       {isMobile ? (
         <MobileRulesReview />
       ) : (
-        <Row gutter={20}>
+        <Row gutter={20} className="gap-y-5">
```
wraeth-eth, line 25: I think its more correct to use two separate rows here rather than let the row wrap around with span pushing

wraeth-eth (changes requested):

src/components/Create/components/pages/ReviewDeploy/components/RulesReview/RulesReview.tsx
tomquirk: In the spirit of anti-antd, i removed `row` and went with a wrapping flex \o/

PR #2962, opened 2023-02-06
Fix graph search

tomquirk:
`entitiesFromSGResponse` was looking for `projectSearchs` (because `s` was always prefixed), which led to no results rendered.

Not sure if best fix but it works.


peripheralist (approved):
perfect, thank you 🙏

Issue #2975, opened 2023-02-07
Update nav bar


tomquirk:
## Description

- [ ] Add 'Create a project' link to nav (far right)
   - [ ] Track button clicks in Fathom. Use `LHWCLH9E` event ID
- [ ] Change 'Projects' to 'Explore'  
  - [ ] Track button clicks in Fathom. Use `YTTX60ZQ` event ID

## Screenshots

![](https://cdn.discordapp.com/attachments/970421807767236708/1072330470978769067/Screen_Shot_2023-02-07_at_9.31.02_am.png)

PR #2979, opened 2023-02-07
Update navbar

NakedFool:

Closes https://github.com/jbx-protocol/juice-interface/issues/2975

I've created smol utility function `recordClickInFathom` so we can use it without code-repetition in the future, lmk if you like it!

Also, I've replaced it in `HeroSection.tsx` so it uses that utility function now.

One more thing that is not related to this PR but I stumbled upon that confusion when I created this PR. 
Initially, I named this branch as `update-navbar` but when I tried to push it I was getting a non-fast-forward error, as the branch is not updated to its remote branch; my confusion was why it was not updated since it was created just now.

Then I realized that it could be some stale branch that was named exactly like this one, it looks like we got over 1000 branches and most of them are already merged, we can begin the process of deleting all those branches?

image

## Acceptance checklist

### DIFF
--- a/src/components/Navbar/SiteNavigation.tsx
+++ b/src/components/Navbar/SiteNavigation.tsx
@@ -43,11 +43,7 @@ export default function SiteNavigation() {
   return (
     // top-nav is antd override
     <nav className="top-nav z-[1] flex h-16 items-center justify-between bg-smoke-25 py-2 px-12 leading-[64px] dark:bg-slate-800">
-      <Menu
-        className="flex w-[500px] flex-row"
-        items={menuItems}
-        mode="inline"
-      />
+      <Menu className="flex flex-row" items={menuItems} mode="inline" />
 
       <div className="flex items-center gap-6">
         <NavLanguageSelector />
--- a/src/utils/fathom.ts
+++ b/src/utils/fathom.ts
@@ -1,0 +1,6 @@
+import { fathom } from '../lib/fathom'
+
+export const recordClickInFathom = (eventId: string) => {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  ;(fathom as any)?.trackGoal(eventId, 0)
+}
--- a/src/utils/fathom.ts
+++ b/src/utils/fathom.ts
@@ -1,6 +1,6 @@
 import { fathom } from '../lib/fathom'
 
-export const recordClickInFathom = (eventId: string) => {
+export const trackFathomGoal = (eventId: string) => {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   ;(fathom as any)?.trackGoal(eventId, 0)
 }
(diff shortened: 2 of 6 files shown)


src/components/Navbar/SiteNavigation.tsx
```diff
@@ -43,11 +43,7 @@ export default function SiteNavigation() {
   return (
     // top-nav is antd override
     <nav className="top-nav z-[1] flex h-16 items-center justify-between bg-smoke-25 py-2 px-12 leading-[64px] dark:bg-slate-800">
-      <Menu
-        className="flex w-[500px] flex-row"
```
NakedFool, line 47: Do we need to specify the width of a menu here? I've removed it since it's flex but I can add '600px' for example if we need it.

With adding a new link to our navbar, on small laptops and tablets it's a bit cumbersome, we can probably think about shrinking it better in some near future ( gonna need designer input probably )

src/utils/fathom.ts
```diff
@@ -0,0 +1,6 @@
+import { fathom } from '../lib/fathom'
+
+export const recordClickInFathom = (eventId: string) => {
```
tomquirk, line 3: **praise** absolutely love this!

**suggestion**: WDYT about naming this `trackFathomGoal` instead, to more closely align with their `fathom.trackGoal` api?

src/components/Navbar/SiteNavigation.tsx
tomquirk: Great point. I think you made the right call to remove it.

**suggestion**: WDYT about hiding the `Create a project` button on screens less than tailwind's `lg` (1024px)? I played around with it and i think it might be okay.

tomquirk (changes requested):
@NakedFool looks fantastic, thanks for this!

While you're here (and we have a nice new helper function), could you track clicks on the little search button in the top nav too (code: `OJRCY4EI`)

tomquirk:
also
> we can begin the process of deleting all those branches?
yes... great point.

src/utils/fathom.ts
NakedFool: agreed! I was renaming it a few times but couldn't think of the best name here, thanks!

src/components/Navbar/SiteNavigation.tsx
NakedFool: SGTM, it's a good quick fix. I've applied the tailwind class to element in desktop items and removed it from mobile.

NakedFool:
> @NakedFool looks fantastic, thanks for this!
> 
> While you're here (and we have a nice new helper function), could you track clicks on the little search button in the top nav too (code: `OJRCY4EI`)

Yep all good! added fathom trackings to quick project search as well, let me know how PR feels now overall 🍻

tomquirk:
@NakedFool thanks for making these changes, lets merge!

tomquirk (approved):

Issue #2949, opened 2023-02-02
Update 500 page with better art


tomquirk:
https://discord.com/channels/775859454780244028/874855844863414293/1070755021240676392

Can use Juicetool as inspiration: https://github.com/jbx-protocol/juicetool/blob/main/pages/500.tsx

Image size needs to be compressed a bit, probably.

![Image](https://user-images.githubusercontent.com/12551741/216456123-5d4e7213-5c15-4299-a975-b19a097bcdf4.png)

![Image](https://user-images.githubusercontent.com/12551741/216455619-07b00cb5-f008-46ae-8884-2147f031fb7e.png)

Issue #2948, opened 2023-02-02
Update 404 page with better art


tomquirk:
Related to https://github.com/jbx-protocol/juice-interface/issues/2949

![](https://cdn.discordapp.com/attachments/874855844863414293/1062075264324087848/404error.png)

PR #2991, opened 2023-02-08
Add custom error pages ( 404 and 500 )

NakedFool:

Closes https://github.com/jbx-protocol/juice-interface/issues/2948 
Closes https://github.com/jbx-protocol/juice-interface/issues/2949

I've created custom error pages following this documentation - https://nextjs.org/docs/advanced-features/custom-error-page

Initially, I wanted to create folder `/errors` and include folders `/404` and `/500` inside but that is not possible currently, however now we have `/500` and `/404` folders inside `/pages` directory.

Let me know also what you think of the layout, I think no more text is needed as it is pretty clean, the image speaks for itself and we have a small paragraph that indicates what could be wrong.

One more thing, I guess we want to include translations here so I wrapped it with `LanguageProvider`, I do not think there is any outside magic that can do it differently, but I could be wrong!

Both images are compressed from ~ 600kb to around ~200kb

image
image
## Acceptance checklist


src/pages/404/index.page.tsx
```diff
@@ -0,0 +1,37 @@
...
+    <LanguageProvider>
+      <div className="flex flex-col items-center justify-center p-6 md:p-40">
+        <img
+          src={error404.src}
+          alt="Image"
+          className="mb-6 h-60 w-60 object-center md:h-80 md:w-80"
+        />
+        <Trans>
+          <p className="mb-6 text-sm text-grey-600 md:text-base">
+            This page could not be found.
```
tomquirk, line 22: **suggestion** perhaps we should `Trans` this line and the link text separately, wdyt?

tomquirk (changes requested):
@NakedFool yessss!!! Looks so great, nice work. Just one comment.

src/pages/404/index.page.tsx
NakedFool: gotcha, changing! I gotta ask what is the difference? asking out of curiosity so I know for future implementations 🍻 

Is it so the translation pot breaks it into the next line for translation so it does not appear as one message which it isn't?

tomquirk: @NakedFool yes exactly! Whether to seperate the strings is a case-by-case decision. In this case, I could imagine us using `Go to homepage` text in multiple spots, so we may as well only translate that _once_ and use the same translation everywhere. It's also totally valid to wrap the whole thing in a `Trans` too. Developer discretion!

tomquirk (approved):

PR #2992, opened 2023-02-08
PayEventElem handle caller != beneficiary

peripheralist:

Adds a case to PayEventElem to show both caller and beneficiary when they are not the same address.

- light refactoring to ActivityEventElem puts the `Subject` and caller/beneficiary on the same line, to allow aligning by baseline (needed because now in the case where caller != beneficiary, right side text uses a smaller font size)
- adds a "title" property to FormattedAddress
- adds missing "caller" key to a few project activity events, so we can show on all events

image
image
image

## Acceptance checklist

### DIFF
--- a/src/components/activityEventElems/ActivityElement/ActivityElement.tsx
+++ b/src/components/activityEventElems/ActivityElement/ActivityElement.tsx
@@ -1,4 +1,5 @@
 import EtherscanLink from 'components/EtherscanLink'
+import { ArrowRightOutlined } from '@ant-design/icons'
 import FormattedAddress from 'components/FormattedAddress'
 import { ProjectVersionBadge } from 'components/ProjectVersionBadge'
 import { ThemeContext } from 'contexts/Theme/ThemeContext'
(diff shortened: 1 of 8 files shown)


src/components/activityEventElems/ActivityElement/ActivityElement.tsx
```diff
@@ -9,16 +10,36 @@ import { contentLineHeight, smallHeaderStyle } from '../styles'
...
+
+  return beneficiary &&
+    caller &&
+    beneficiary.toLowerCase() !== caller.toLowerCase() ? (
     <div
       style={{
-        display: 'flex',
-        justifyContent: 'space-between',
-        alignContent: 'space-between',
+        lineHeight: contentLineHeight,
```
wraeth-eth, line 27: Any chance we can get rid of this?

wraeth-eth (changes requested):

src/components/activityEventElems/ActivityElement/ActivityElement.tsx
peripheralist: actuallyyy i think with the change i just made this is unnecessary now. will verify

peripheralist: done

tomquirk (approved):
lgtm @peripheralist nice addition!

PR #2994, opened 2023-02-09
Smol improvements to subgraph types/utils

peripheralist:

- Cleans up a recent quick patch by @aeolianeth to fix handling of singular/plural subgraph entities in queries and return objects
- Adds support for typing subgraph query `keys` that are nested entities
    - Example: when defining `keys` to query "payEvents" only `project` can be used in the structure `{ entity: <entity E>, keys: <keyof E>[] }`, because it is the only property of PayEvent that is also an entity name

image

## Acceptance checklist


peripheralist:
update on this:

not sure if some of this will work since it depends on a property name being an entity name. for example, this works for `participant.project` where "Project" is an entity name, but not with `project.printPremineEvents`, where "printPremineEvents" is something we came up with, and not the entity name which is "MintTokenEvent"

gonna keep thinking on this. will PR the singular/plural part next week if i can't solve the other typing problems

Issue #3005, opened 2023-02-09
Add more Fathom events


tomquirk:
Track the following in Fathom

| page | element | notes |
| --- | --- | --- |
| Each create flow page | ‘next’ button click | so we can see where/when people are dropping off |
| Homepage | `Explore Projects` button | - | 
 | Homepage | `Create a Project` button | - |
| Projects page | `Create a project` Button | - |
| Project page | `Pay` button click | want to see how many people click but don’t follow through
| Nav bar | `Discord` link |  see if its getting clicks vs the rest  |

### Implementation ideas

- create a new file somewhere `fathomEvents`
- Create constants for each event ID. Variable name should represent the event (e.g. `HOMEPAGE_CREATE_PROJECT_CTA`)

### creating events in Fathom
Use @tomquirk 's Fathom account, go to https://app.usefathom.com/sites/ERYRRJSV/events

PR #3017, opened 2023-02-11
Add more fathom events

NakedFool:

Closes https://github.com/jbx-protocol/juice-interface/issues/3005

I've created `fathomEvents` for storing all eventId constants that we now use, I've decided to categorize constants per page, for example, we have `TOP_NAV` which contains all CTAs that are only in the top nav of the application, that way in the future if we decide to add more events we can easily see where to add and it is more readable, lmk what you think of this approach!

One more thing, I think we need to remove the old event `Click Design Your Project CTA` since it used to be an event for tracking creating a project on the hero homepage, which we now replaced with a new eventId, lmk if I'm being correct here! 🍻 

EDIT: 
Hmm, I think something happened in the meantime, I was initially creating this branch from `main` but I'm seeing a lot of conflicts right now, let me see how to handle it.

## Acceptance checklist


tomquirk:
@NakedFool heads up, going to push a rebase!

src/components/Navbar/navigationItems.tsx
```diff
@@ -130,7 +130,10 @@ export const desktopMenuItems = ({
     key: 'discord',
     label: (
       <Link href="https://discord.gg/wFTh4QnDzk">
-        <a className="flex cursor-pointer items-center font-medium text-black hover:opacity-70 dark:text-slate-100">{t`Discord`}</a>
+        <a
+          className="flex cursor-pointer items-center font-medium text-black hover:opacity-70 dark:text-slate-100"
+          onClick={() => trackFathomGoal('PKIBZYTS')}
```
tomquirk, line 135: **suggestion** could we use the goal constant variable from the constants file here and throughout?

NakedFool: I think we all good now! 🍻

tomquirk (approved):
Love your work @NakedFool , approving!

PR #3032, opened 2023-02-14
IPFS Retries for sepana projects + sepana db schema

peripheralist:

- Introduces a new pattern for limiting the number of times we retry resolving IPFS metadata for projects on sepana. Especially important for projects with invalid `metadataUri`s, or unpinned metadata
- Updates to the project schema in the sepana db:
    - internal vars are prefixed with `_`
    - `_lastUpdated` is now a millis timestamp instead of a block number
    - adds new `_version` property useful for future changes to the schema, when we need to force projects to update
- general cleanup of sepana update routine

## Acceptance checklist

### DIFF
--- a/src/lib/sepana/api/generic.ts
+++ b/src/lib/sepana/api/generic.ts
@@ -54,6 +54,16 @@ export async function getRecord<T extends object>(id: string) {
  * @param records Projects to write to Sepana database
  */
 export async function writeSepanaRecords(records: Json<SepanaProject>[]) {
+  const missingIds = records.filter(r => r.id === undefined || r.id === null)
+
+  if (missingIds.length) {
+    throw new Error(
+      `${missingIds.length} records are missing an id: ${missingIds
+        .map(p => JSON.stringify(p))
+        .join(', ')}`,
+    )
+  }
+
   const jobs: string[] = []
   const errors: (string | object)[] = []
   const written: Json<SepanaProject>[] = []
@@ -63,7 +73,14 @@ export async function writeSepanaRecords(records: Json<SepanaProject>[]) {
   let page = 0
 
   while (records.length > pageSize * page) {
-    const queue = records.slice(pageSize * page, pageSize * (page + 1))
+    const queue = records
+      .slice(pageSize * page, pageSize * (page + 1))
+      .map(r => ({
+        ...r,
+        // Upserting data in Sepana requires the `_id` param to be included, so we always include it here using `_source.id`
+        // https://docs.sepana.io/sepana-search-api/web3-search-cloud/search-api#request-example-2
+        _id: r.id,
+      }))
 
     await sepanaAxios('read/write')
       .post<{ job_id: string }>(SEPANA_ENDPOINTS.insert, {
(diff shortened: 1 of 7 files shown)


src/lib/sepana/api/generic.ts
```diff
@@ -63,7 +73,14 @@ export async function writeSepanaRecords(records: Json<SepanaProject>[]) {
   let page = 0
 
   while (records.length > pageSize * page) {
-    const queue = records.slice(pageSize * page, pageSize * (page + 1))
+    const queue = records
+      .slice(pageSize * page, pageSize * (page + 1))
+      .map(r => ({
+        ...r,
+        // Upserting data in Sepana requires the `_id` param to be included, so we always include it here using `_source.id`
+        // https://docs.sepana.io/sepana-search-api/web3-search-cloud/search-api#request-example-2
```
tomquirk, line 81: **praise** great explanation and link <3

src/pages/api/sepana/update.page.ts
```diff
@@ -1,132 +1,80 @@
...
 // Synchronizes the Sepana engine with the latest Juicebox Subgraph/IPFS data
 const handler: NextApiHandler = async (req, res) => {
   try {
     // This flag will let us know we should retry resolving IPFS data for projects that are missing it
-    const retryIPFS = req.method === 'POST' && req.body['retryIPFS'] === true
-
-    // Load all projects from Sepana
-    const sepanaProjects = (await queryAll<Json<SepanaProject>>()).data.hits
-      .hits
+    const retryIpfs = req.body['retryIPFS'] === true
```
tomquirk, line 20: **suggestion (non0blocking)** I guess this is equiv to : 
```suggestion
    const retryIpfs = req.body.retryIPFS === true
```
Or
```suggestion
    const retryIpfs = Boolean(req.body.retryIPFS)
```

src/utils/sepana.ts
```diff
@@ -1,10 +1,173 @@
...
+        _hasUnresolvedMetadata: true,
+        _metadataRetriesLeft: 0,
+      } as Json<SepanaProject>,
+    }
+  }
+
+  try {
+    const {
+      data: { name, description, logoUri },
+    } = await infuraApi.get<AnyProjectMetadata>(openIpfsUrl(metadataUri), {
```
tomquirk, line 141: **suggestion** could potentially use `ipfsGetWithFallback` here.

tomquirk (approved):
@peripheralist thanks for sending this my way, was good to get more familiar with sepana. This LGTM just a couple super minor comments

src/utils/sepana.ts
peripheralist: was originally doing this, but that was using the pinata url at the time which was failing on stuff that wasn't pinned via our gateway. will clean this up at some point but gonna leave for this PR

src/pages/api/sepana/update.page.ts
peripheralist: second would resolve to true if the `retryIPFS` param was anything truthy, so would rather leave the explicit `=== true`. first suggestion makes sense!

Issue #2585, opened 2022-11-29
Finish migrating IPFS to Infura


tomquirk:
## Description

Stop using Piñata altogether, use Infura instead.

Incomplete list of things to migrate:
- project metadata GET
- project metadata POST on create
- project metadata POST on reconfigure (settings page, V1 and V2)
- project logos GET
- project logos POST (upload) on create  (see **notes**)
- project logos POST (upload) on reconfigure  (see **notes**)

Finally:
- re-pin all project metadata to Infura
- re-pin all project logos to Infura

Notes: 
- ideally these a stored as `ipfs://` links on chain (and not https links to a gateway)

### Previous efforts:
https://github.com/jbx-protocol/juice-interface/pull/2819
https://github.com/jbx-protocol/juice-interface/pull/2687

PR #3046, opened 2023-02-15
Clean up pinata

tomquirk:
Closes https://github.com/jbx-protocol/juice-interface/issues/2585

Pinata is no longer in use across the site. This PR cleans up any unused code.

Infura has no rate limit issues (150 reqs per second) so I don't think we need the complexities of sign-to-upload.

Genuinely sad and sorry to be removing your hard work here @wraeth-eth. It's very good work and we can def draw from this gained knowledge in the future. I packaged it up in this PR so that we have a reference point if we ever need to come back to it.

LMKWYT.


src/utils/ipfs.ts
```diff
@@ -3,24 +3,14 @@ import { base58 } from 'ethers/lib/utils'
...
 /**
  * Return a URL to our open IPFS gateway for the given cid USING INFURA.
  *
  * The 'open' gateway returns any content that is available on IPFS,
  * not just the content we have pinned.
  */
-export const ipfsOpenGatewayUrl = (cid: string | undefined): string => {
-  return ipfsGatewayUrl(cid, OPEN_IPFS_GATEWAY_HOSTNAME)
+export const ipfsGatewayUrl = (cid: string | undefined): string => {
+  return `https://${OPEN_IPFS_GATEWAY_HOSTNAME}/ipfs/${cid}`
```
wraeth-eth, line 13: Will we need to change things locally?

src/utils/server/ipfs/infuraScheduler.ts
```diff
@@ -0,0 +1,18 @@
+import Bottleneck from 'bottleneck'
+
+// Back off for 3 secs - pinata rate limits for 180/min
```
wraeth-eth, line 3: Change these settings to match infura IPFS backoff

wraeth-eth (approved):
@tomquirk nah this is great! you misunderestimate how much I love deleting code

src/utils/ipfs.ts
tomquirk: nup all g

src/utils/server/ipfs/infuraScheduler.ts
tomquirk: ah good catch, done!

Issue #3010, opened 2023-02-10
Reserved tokens applying 2.5% payout fee in reserved token display


wraeth-eth:
## Summary

_A clear and concise description of what the bug is._

## Discord Link

https://discord.com/channels/775859454780244028/1070622535026225202/1070622539925176350

PR #3047, opened 2023-02-15
Remove fee calculations in text on distribute reserved tokens modal

NakedFool:

Closes https://github.com/jbx-protocol/juice-interface/issues/3010

I just got one question, we are not issuing fees when distributing reserved tokens, right?

Additionally, I'm passing the `showFees` prop to determine if the `splitValue` will be displayed with or without fees. I'm not sure whether we need to add a unique property in `SplitProps` in this situation, but this feels cool to me, wdyt?

image
## Acceptance checklist


src/components/v2v3/shared/SplitItem/SplitAmountValue.tsx
```diff
@@ -60,7 +60,9 @@ export function SplitAmountValue({ props }: { props: SplitProps }) {
         <span>
           (
           <CurrencySymbol currency={curr} />
-          {splitValueFormatted}
```
johnnyd-eth, line 63: I think this would be better named `splitValueAfterFee`

src/components/v2v3/shared/SplitItem/SplitAmountValue.tsx
```diff
@@ -60,7 +60,9 @@ export function SplitAmountValue({ props }: { props: SplitProps }) {
...
+          {props.showFees
+            ? splitValueFormatted
+            : formatWad(splitValue, { precision: 0 })}
```
johnnyd-eth, line 65: **Question:** Why `precision: 0`? Discrete tokens are still possible, and we should account for possibly showing an ETH value in this component when `showFees` is false.

johnnyd-eth, line 65: And would be great to define a `valueFormatted` `const` for this.

johnnyd-eth:
Looks good @NakedFool just a few comments!

src/components/v2v3/shared/SplitItem/SplitAmountValue.tsx
NakedFool: Agreed, much clearer.

src/components/v2v3/shared/SplitItem/SplitAmountValue.tsx
NakedFool: You have a good point, however aren't we only in this situation when `showFees` is false accounts for these values for reserved tokens? If so, ETH value would never display up. Am I missing something? 

With the line `const reservedTokensFormatted = formatWad(reservedTokens, "precision: 0")` , we are precisioning values in the `DistributeReservedTokensModal` to 0.

NakedFool: Should we name it to `valueBeforeFeeFormatted` or smth like that?

NakedFool:
> Looks good @NakedFool just a few comments!

Thank you ser, i agreed with most of it but raised 2 questions above, let me know what you think!

src/components/v2v3/shared/SplitItem/SplitAmountValue.tsx
NakedFool: Agreeing with 'valueFormatted' here at the end! 🍻

NakedFool:
As we can't manipulate the state solely on `showFees` which will corrupt other components like payouts and cause it to display amounts with fees there, I've added the `isReservedTokensSplit` prop to `SplitProps`.

Let me know what you think of the latest approach and if there is a way to do this without passing additional props! 🍻

johnnyd-eth (approved):

Issue #3052, opened 2023-02-16
Sticker click not working on pay modal


wraeth-eth:
## Summary

When I click a sticker on the pay modal, nothing happens. No errors.

## Steps to reproduce

1. Go to https://juicebox.money/v2/p/436
2. Enter some amount and click on 'Contribute'
4. On the pay modal, Add any sticker
5. No response or error is displayed

## What is the _current_ bug behavior?

Clicking on any sticker on the pay modal generates no response or error message.

## What is the expected _correct_ behavior?

Clicking on a sticker should add it to the payment memo.

## Relevant logs and/or screenshots

No console logs are generated when clicking on a sticker.

## Discord Link

https://discord.com/channels/775859454780244028/1075846884654055536/1075846886784774234

PR #3057, opened 2023-02-17
Remove NFTs from sticker selection

tomquirk:

Closes https://github.com/jbx-protocol/juice-interface/issues/3052

- removes NFTs from sticker selection. They ain't stickers, so i dont think they need to be coerced to be stickers
- Removes transactionCanceled logic. Pay form data is correctly retained when a tx is cancelled


src/components/v2v3/V2V3Project/V2V3PayButton/V2V3ConfirmPayModal/V2V3PayForm.tsx
```diff
@@ -62,24 +59,12 @@ export const V2V3PayForm = ({
...
-    })
-  }, [form, nftRewardTiers, transactionCanceled])
-
-  useEffect(() => {
-    if (!form.getFieldValue('beneficiary')) return
-
-    setCustomBeneficiaryEnabled(transactionCanceled)
-  }, [form, transactionCanceled])
-
+  useEffect(
```
tomquirk, line 62: this is pretty inconsequential, but we should add `beneficiaryEnabled` to the form prob.

wraeth-eth (approved):

PR #3061, opened 2023-02-18
Add isZeroAddress util and use errywhere

tomquirk:
title


src/components/v1/V1Project/modals/ConfirmUnstakeTokensModal.tsx
```diff
@@ -50,9 +50,7 @@ export default function ConfirmUnstakeTokensModal({
     )
   }
 
-  const ticketsIssued = tokenAddress
-    ? tokenAddress !== constants.AddressZero
-    : false
+  const ticketsIssued = Boolean(tokenAddress && isZeroAddress(tokenAddress))
```
tomquirk, line 53:
```suggestion
  const ticketsIssued = Boolean(tokenAddress && !isZeroAddress(tokenAddress))
```

PR #3083, opened 2023-02-23
Fix various NFT MP4 loading states

johnnyd-eth:

- **Activity feed:**
BEFORE:
Screen Shot 2023-02-23 at 3 51 31 pm

AFTER:
Screen Shot 2023-02-23 at 3 48 22 pm

- **Reconfigure:**

https://user-images.githubusercontent.com/96150256/221079689-35eb6ec8-3878-4675-9e11-1d16b256ec73.mp4

## Acceptance checklist

### DIFF
--- a/src/components/Create/components/JuiceVideoOrImgPreview.tsx
+++ b/src/components/Create/components/JuiceVideoOrImgPreview.tsx
@@ -2,6 +2,7 @@ import { CloseOutlined } from '@ant-design/icons'
 import { Image, ImageProps } from 'antd'
 import { JuiceVideoPreview } from 'components/NftRewards/NftVideo/JuiceVideoPreview'
 import { useContentType } from 'hooks/ContentType'
+import { stopPropagation } from 'react-stop-propagation'
 import { fileTypeIsVideo } from 'utils/nftRewards'
 
 export const JUICE_IMG_PREVIEW_CONTAINER_CLASS =
@@ -22,11 +23,13 @@ export function JuiceVideoOrImgPreview({
   const { data: contentType } = useContentType(src)
   const isVideo = fileTypeIsVideo(contentType)
 
+  const _onClose = stopPropagation(onClose)
+
   return (
-    <div className={JUICE_IMG_PREVIEW_CONTAINER_CLASS} onClick={onClose}>
+    <div className={JUICE_IMG_PREVIEW_CONTAINER_CLASS} onClick={_onClose}>
       <CloseOutlined
         className="absolute top-10 right-10 pl-4 text-2xl text-slate-100"
-        onClick={onClose}
+        onClick={_onClose}
       />
       <div className="text-center">
         {isVideo ? (
--- a/src/components/NftRewards/NftVideo/JuiceVideoThumbnailOrImage.tsx
+++ b/src/components/NftRewards/NftVideo/JuiceVideoThumbnailOrImage.tsx
@@ -3,12 +3,13 @@ import { ImageProps } from 'antd'
 import { JuiceVideoOrImgPreview } from 'components/Create/components/JuiceVideoOrImgPreview'
 import { useContentType } from 'hooks/ContentType'
 import { useState } from 'react'
+import { stopPropagation } from 'react-stop-propagation'
 import { classNames } from 'utils/classNames'
 import { fileTypeIsVideo } from 'utils/nftRewards'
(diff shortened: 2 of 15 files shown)


tomquirk:
@johnnyd-eth possible to skeleton loader it?

johnnyd-eth:
> @johnnyd-eth possible to skeleton loader it?

Are you thinking the UI should be different? The `Skeleton` component would do something like this I guess:

https://user-images.githubusercontent.com/96150256/220831534-a6526321-6336-424f-9d3d-c5ebb43661bf.mp4

This is quite different to the NFT cards' loading state, however. I was trying to keep it as similar as possible. Happy to revamp them all if we want to move away from the spinner though.

tomquirk:
@johnnyd-eth got it, lets ship as is. I think it looks a bit sketchy with the border is all. No worries though, net improvement 👍 
Thanks for this PR 🙏

src/components/Create/components/JuiceVideoOrImgPreview.tsx
```diff
@@ -22,11 +23,13 @@ export function JuiceVideoOrImgPreview({
   const { data: contentType } = useContentType(src)
   const isVideo = fileTypeIsVideo(contentType)
 
+  const _onClose = stopPropagation(onClose)
```
tomquirk, line 26: **question** why the need for stopPropagation?

src/components/NftRewards/NftVideo/JuiceVideoThumbnailOrImage.tsx
```diff
@@ -3,12 +3,13 @@ import { ImageProps } from 'antd'
...
 import { useContentType } from 'hooks/ContentType'
 import { useState } from 'react'
+import { stopPropagation } from 'react-stop-propagation'
 import { classNames } from 'utils/classNames'
 import { fileTypeIsVideo } from 'utils/nftRewards'
 import { JuiceVideoThumbnail, PlayIconPosition } from './JuiceVideoThumbnail'
 
 export function JuiceVideoThumbnailOrImage({
-  isSelected,
+  isSelected = true,
```
tomquirk, line 12: **question** why default this to true? Generally better to make props false by default, then enable where needed.

src/utils/ipfs.ts
```diff
@@ -75,3 +77,8 @@ export function isIpfsCID(cid: string) {
     cid.startsWith('Qm') || cid.startsWith('bafy') || cid.startsWith('bafk')
   )
 }
+
+export function percentFromUploadProgressEvent(e: UploadProgressEvent) {
+  const _percent = (e.loaded / e.total) * 100
```
tomquirk, line 82: **suggestion** 

```suggestion
  const percent = (e.loaded / e.total) * 100
```

src/components/Create/components/JuiceVideoOrImgPreview.tsx
```diff
@@ -2,6 +2,7 @@ import { CloseOutlined } from '@ant-design/icons'
 import { Image, ImageProps } from 'antd'
```
tomquirk, line 2: **suggestion** this component is used outside of the Create flow. Should it be moved to the `components/` directory?

johnnyd-eth: Done

src/utils/ipfs.ts
johnnyd-eth: Done

src/components/NftRewards/NftVideo/JuiceVideoThumbnailOrImage.tsx
johnnyd-eth: If it defaulted to `false` then images/video thumbnails would be darkened by default. We could change the prop to something like `darkened` but it's less clear in the selectable instances.

src/components/Create/components/JuiceVideoOrImgPreview.tsx
johnnyd-eth: Because in the instances of `<JuiceVideoThumbnailOrImage>` in the upload forms, closing this preview triggers a click on the upload field.

tomquirk: aids

tomquirk (approved):

PR #3103, opened 2023-02-27
Consolidate `NftRewardTier` types

johnnyd-eth:

Prep for #3101 

## Acceptance checklist


src/components/Create/components/RewardsList/AddEditRewardModal.tsx
```diff
@@ -107,13 +107,19 @@ export const AddEditRewardModal = ({
...
+      id: editingData?.id ?? parseInt(v4()),
+      name: fields.rewardName,
+      contributionFloor: parseFloat(fields.contributionFloor ?? '0'),
       description: fields.description,
-      maximumSupply: fields.maxSupply ? parseInt(fields.maxSupply) : undefined,
-      url: fields.externalUrl ? `https://${fields.externalUrl}` : undefined,
+      maxSupply: remainingAndMaxSupply,
+      remainingSupply: remainingAndMaxSupply,
+      externalLink: fields.externalUrl
+        ? `https://${fields.externalUrl}`
```
tomquirk, line 121: **suggestion**: could use `withHttps` util here to be safe

src/models/nftRewardTier.ts
```diff
@@ -31,7 +31,7 @@ export type DelegateVersion = JB721DELEGATE_V1 | JB721DELEGATE_V1_1
 
 // How the reward tiers are stored in the contracts
 export type JB721TierParams = {
-  id?: BigNumber //undefined for outgoing tier (in launch or adjustTiers tx)
+  id: BigNumber //undefined for outgoing tier (in launch or adjustTiers tx)
```
tomquirk, line 34: **question** can we get rid of this comment here now?

src/models/nftRewardTier.ts
```diff
@@ -31,7 +31,7 @@ export type DelegateVersion = JB721DELEGATE_V1 | JB721DELEGATE_V1_1
...
   remainingQuantity?: BigNumber //undefined for outgoing tier (in launch or adjustTiers tx)
```
tomquirk, line 35: **suggestion (non-blocking)** ideally we might have a separate type that handles this case. As in, JB721TierParams might have `remainingQuantity`, but then some other type would be `type EditingJB721TierParams = Omit<JB721TierParams, 'remainingQuantity'>`, for example

tomquirk (changes requested):
@johnnyd-eth huge praise for this PR. Love to see it. couple comments 🙏

src/models/nftRewardTier.ts
johnnyd-eth: Yeah this is a good idea, although will be a solid diff. Happy to get this PR in now since it's already quite hefty?

johnnyd-eth:
> @johnnyd-eth huge praise for this PR. Love to see it. couple comments 🙏

Pushed the non-blockers!

johnnyd-eth:
(Verbal approval from @tomquirk who isn't at his computer)

PR #3119, opened 2023-03-01
Lazy load project page tabs

tomquirk:
Performance gainz! Only load the JS for tabs when the tab is active!


tomquirk:
cc @johnnyd-eth

PR #3144, opened 2023-03-03
Add option to reverse filter on projects page

filipviz:

Add option to reverse filter on projects page. Lets you see earliest projects on various versions (or projects with no volume if you care to see them).

## Acceptance checklist

### DIFF
--- a/src/pages/projects/ProjectsFilterAndSort.tsx
+++ b/src/pages/projects/ProjectsFilterAndSort.tsx
@@ -21,6 +21,8 @@ export default function ProjectsFilterAndSort({
   setIncludeV2,
   showArchived,
   setShowArchived,
+  reversed,
+  setReversed,
   orderBy,
   setOrderBy,
 }: {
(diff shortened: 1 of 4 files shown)


src/pages/projects/ProjectsFilterAndSort.tsx
```diff
@@ -87,6 +91,11 @@ export default function ProjectsFilterAndSort({
               checked={showArchived}
               onChange={setShowArchived}
             />
+            <FilterCheckboxItem
+              label={t`Reverse`}
```
tomquirk, line 95: **question** hmm, i wonder if this should be in a different "Sort order" dropdown.

WDYT @strath-m ?

tomquirk (approved):
@filipviz love it! just one UX comment, lmkwyk

filipviz:
@aeolianeth I considered this, but the bar would become crowded — especially on mobile. I think fine to leave as is for now (especially considering Sepana updates will likely change this entire section). Unless @strath-m strongly disagrees!

tomquirk:
@filipviz great point, feeling good about it as is. Let's merge

PR #3169, opened 2023-03-08
Support distributePayoutsOf for JB 3.1

tomquirk:

Closes JB-46

Makes distributePayoutsOf work if the project is using JBETHTerminal 3.1: https://github.com/jbx-protocol/juice-interface/pull/3169/files#r1128905499.

How it works:
- In V2V3ProjectContractsContext, we are loading the specific contracts that the project uses. We also determine the version of these contracts and provide that to the context too: https://github.com/jbx-protocol/juice-interface/blob/90b7112c4f1ee75af308d6d7295f82925ef74685/src/contexts/v2v3/ProjectContracts/V2V3ProjectContracts.ts#L16
- In teh `distributePayoutsOf` hook, we send different arguments depending on which version of the contract is in use. Lucky in this case, the args are pretty much the same!
- For a similar example, see `addToBalanceOf`: https://github.com/jbx-protocol/juice-interface/blob/74c48a6389662b09cfb29c7194c622dffab7d7cd/src/hooks/v2v3/transactor/AddToBalanceTx/AddToBalanceTx.ts#L22-L37

Has some other random unrelated changes too.

## Acceptance checklist

### DIFF
--- a/src/hooks/v2v3/transactor/DistributePayouts.ts
+++ b/src/hooks/v2v3/transactor/DistributePayouts.ts
@@ -1,5 +1,6 @@
 import { BigNumber } from '@ethersproject/bignumber'
 import { t } from '@lingui/macro'
+import { DEFAULT_MIN_RETURNED_TOKENS } from 'constants/transactionDefaults'
 import { ETH_TOKEN_ADDRESS } from 'constants/v2v3/juiceboxTokens'
 import { ProjectMetadataContext } from 'contexts/shared/ProjectMetadataContext'
 import { TransactionContext } from 'contexts/Transaction/TransactionContext'
@@ -14,8 +15,6 @@ type DistributePayoutsTx = TransactorInstance<{
   amount: BigNumber | undefined
   currency: V2V3CurrencyOption | undefined
 }>
-
-const DEFAULT_MIN_RETURNED_TOKENS = 0
 
 export function useDistributePayoutsTx(): DistributePayoutsTx {
   const { transactor } = useContext(TransactionContext)
--- a/src/hooks/v2v3/transactor/DistributePayouts.ts
+++ b/src/hooks/v2v3/transactor/DistributePayouts.ts
@@ -12,26 +12,67 @@ import { V2V3ProjectContractsContext } from 'contexts/v2v3/ProjectContracts/V2V3
 import { TransactorInstance } from 'hooks/Transactor'
 import { V2V3CurrencyOption } from 'models/v2v3/currencyOption'
 import { useContext } from 'react'
+import invariant from 'tiny-invariant'
 import { useV2ProjectTitle } from '../ProjectTitle'
-import { ETH_PAYMENT_TERMINAL_V_3_1 } from '../V2V3ProjectContracts/projectContractLoaders/ProjectPrimaryEthTerminal'
+import {
+  JBETHPaymentTerminalVersion,
+  JB_ETH_PAYMENT_TERMINAL_V_3,
+  JB_ETH_PAYMENT_TERMINAL_V_3_1,
+} from '../V2V3ProjectContracts/projectContractLoaders/ProjectPrimaryEthTerminal'
 
 interface DistributePayoutsTxBaseParams {
   amount: BigNumber | undefined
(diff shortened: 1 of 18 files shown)


src/hooks/v2v3/transactor/DistributePayouts.ts
```diff
@@ -1,5 +1,10 @@
 import { BigNumber } from '@ethersproject/bignumber'
```
tomquirk, line 1: This file is the core of the change.

src/hooks/v2v3/transactor/DistributePayouts.ts
```diff
@@ -8,29 +13,40 @@ import { TransactorInstance } from 'hooks/Transactor'
...
-}>
+}
+
+type DistributePayoutsTx3Params = DistributePayoutsTxBaseParams & {
+  memo?: string
+}
+
+// eslint-disable-next-line @typescript-eslint/naming-convention
+type DistributePayoutsTx3_1Params = DistributePayoutsTxBaseParams & {
+  metadata?: string // JB terminal 3.1 replaced `memo` with `metadata`
```
peripheralist, line 29: 🫡

src/hooks/v2v3/transactor/DistributePayouts.ts
```diff
@@ -8,29 +13,40 @@ import { TransactorInstance } from 'hooks/Transactor'
...
   amount: BigNumber | undefined
   currency: V2V3CurrencyOption | undefined
-}>
+}
+
+type DistributePayoutsTx3Params = DistributePayoutsTxBaseParams & {
+  memo?: string
+}
+
+// eslint-disable-next-line @typescript-eslint/naming-convention
```
peripheralist, line 27: should we just disable this eslint rule? i feel like it's used in multiple places, or could be

tomquirk: yeah good call

src/hooks/v2v3/transactor/DistributePayouts.ts
```diff
@@ -41,11 +57,13 @@ export function useDistributePayoutsTx(): DistributePayoutsTx {
...
       [
         projectId,
-        amount,
-        currency,
+        args.amount,
+        args.currency,
         ETH_TOKEN_ADDRESS, // _token
         DEFAULT_MIN_RETURNED_TOKENS, // _minReturnedTokens
-        memo,
+        versions.JBETHPaymentTerminal === ETH_PAYMENT_TERMINAL_V_3_1
```
peripheralist, line 64: i feel like the neatest way to do this might be having a `useDistributePayoutsTxV3()` and `useDistributePayoutsTxV3_1()` to avoid type casting args. but tbh prolly not be worth the trouble. we'd still have a dependency somewhere on `versions.JBETHPaymentTerminal` which might make it more messy. this makes perfect sense and i'm cool with it

peripheralist (approved):

src/hooks/v2v3/transactor/DistributePayouts.ts
tomquirk: Ya good call, I pretty much agree although I'm thinking it's slightly nicer to have a single `distributePayouts` hook that handles all this logic internally, that the dev doesn't have to concern themselves with "argh which version do i need". AddToBalance is a better example where there's arg factories for each version, and then depending on the version, we use the relevant args https://github.com/jbx-protocol/juice-interface/blob/74c48a6389662b09cfb29c7194c622dffab7d7cd/src/hooks/v2v3/transactor/AddToBalanceTx/AddToBalanceTx.ts#L22-L37.

tomquirk: @peripheralist rolled wid it in latest commit 🙏  It's more verbose but overall clearer i think. Thanks for the call out! https://github.com/jbx-protocol/juice-interface/pull/3169/commits/3d652c35127534052845fd5b9f117a86cd181dc8#diff-5c7a868881aaa286cdc7448bd25449851402814fa24882d0f8b172fee9674c81R45

PR #3195, opened 2023-03-10
fix! Allow correctly calling error

wraeth-eth:

_Provide a description of what this PR does. Link to any relevant GitHub issues, Notion tasks or Discord discussions._

## Acceptance checklist

### DIFF
--- a/src/hooks/Transactor.ts
+++ b/src/hooks/Transactor.ts
@@ -171,13 +171,12 @@ export function useTransactor(): Transactor | undefined {
           let json = message.split('(error=')[1]
           json = json.split(', method=')[0]
           description = JSON.parse(json).message || message
+          options?.onError?.(new DOMException(description))
         } catch (_) {
           description = message
           options?.onError?.(new DOMException(description))
           emitErrorNotification(t`Transaction failed`, { description })
         }
-
-        options?.onDone?.()
 
         return false
       }
--- a/src/hooks/Transactor.ts
+++ b/src/hooks/Transactor.ts
@@ -171,12 +171,13 @@ export function useTransactor(): Transactor | undefined {
           let json = message.split('(error=')[1]
           json = json.split(', method=')[0]
           description = JSON.parse(json).message || message
-          options?.onError?.(new DOMException(description))
         } catch (_) {
           description = message
           options?.onError?.(new DOMException(description))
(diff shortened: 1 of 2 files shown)


wraeth-eth:
* main
  * **PR #3195** <a href="https://app.graphite.dev/github/pr/jbx-protocol/juice-interface/3195" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jbx-protocol/juice-interface/3195?utm_source=stack-comment).

src/hooks/Transactor.ts
```diff
@@ -171,14 +171,13 @@ export function useTransactor(): Transactor | undefined {
...
           json = json.split(', method=')[0]
           description = JSON.parse(json).message || message
+          options?.onError?.(new DOMException(description))
         } catch (_) {
           description = message
           options?.onError?.(new DOMException(description))
           emitErrorNotification(t`Transaction failed`, { description })
         }
 
-        options?.onDone?.()
```
peripheralist, line 180: does this mean onDone won't be called if there's an error?

wraeth-eth: Yep - should it be called? I see `onError` as a failure and it won't continue. We can `onDone` but it means we may see places calling some code twice

tomquirk: @wraeth-eth this will break a lot of code.

Unfortunately we rely on `onDone` being called on error, because we use `onDone` for most transaction error handling. 

Example: https://github.com/jbx-protocol/juice-interface/blob/82fae112d70396eda3d680d93438302633d95005/src/hooks/JB721Delegate/transactor/ReconfigureV2V3FundingCycleWithNftsTx.ts#L80-L81

It's a legacy thing that's certainly not ideal or clear at all, but we'll have to consider this.

tomquirk: yup totally makes sense @wraeth-eth

peripheralist: I'm cool with that. The original `onError()` was meant to fire if the transaction fails (i.e. show an error message) and `onDone()` was to fire when the transaction completes, wether it succeeds or fails. But transactor has changed so much since then I'm not even sure how we're using them now. As long as nothing breaks 🫡

PR #3241, opened 2023-03-17
Show "unlimited" payout in Edited Payout events

peripheralist:
If the new distribution limit in a SetFundAccessConstraintsEvent (Edited Payout event) is the maximum, show "Unlimited" instead of the numerical value.

image

Other events are unchanged:
image


tomquirk (approved):
@peripheralist lurrrv it, ty ty

PR #3255, opened 2023-03-20
Hide "End" on FC reconfig if "Start" is unset

johnnyd-eth:
Closes JB-138 (https://linear.app/peel/issue/JB-138/edit-cycle-preview-cycle-end-time-is-sometimes-wrong)


src/components/v2v3/V2V3Project/V2V3FundingCycleSection/FundingCycleDetails/FundingCycleListItems/index.tsx
```diff
@@ -44,7 +44,7 @@ export function FundingCycleListItems({
     ? fundingCycle.start.mul(1000).gt(BigNumber.from(Date.now()))
```
tomquirk, line 44: @johnnyd-eth we should probably just show start always.

tomquirk (approved):

PR #3267, opened 2023-03-21
Fix can't read 404 page text

johnnyd-eth:
Closes JB-133 (https://linear.app/peel/issue/JB-133/fix-404-page-text-color)

Screen Shot 2023-03-21 at 3 24 17 pm

### DIFF
--- a/src/styles/next-overrides.scss
+++ b/src/styles/next-overrides.scss
@@ -1,0 +1,3 @@
+.next-error-h1, .next-error-h1 + div > h2 {
+  color: white;
+}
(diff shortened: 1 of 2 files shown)


src/styles/next-overrides.scss
```diff
@@ -0,0 +1,3 @@
+.next-error-h1, .next-error-h1 + div > h2 {
```
tomquirk, line 1: christ. fine 😂

tomquirk (approved):

PR #3274, opened 2023-03-22
Get project handles from API

tomquirk:
This PR does a number of things to significantly improve perf:
- fetch project handles from a new API endpoint _cached for a day on Vercel,_ instead of directly from infura (which isn't cached)
- Remove JBProjectHandles and PublicResolver from `V2V3ProjectContracts`. Only load them when needed. This means that these contracts won't be loaded up front on the project page, saving about 500ms in initial load time.

Other: 
- font-display: swap, because Lighthouse told me to.


tomquirk:
@wraeth-eth keen for you to take a quick look at this 🙏

PR #3276, opened 2023-03-22
Cached api route for project id

tomquirk:
- Move the "get project ID from handle" code to an API route
- cache successful resolutions (if theres a project ID found at that given handle)
- Use it in middleware

The goal is to speed up middleware.

### DIFF
--- a/src/middleware.page.ts
+++ b/src/middleware.page.ts
@@ -1,5 +1,3 @@
-import { PV_V2 } from 'constants/pv'
-import { paginateDepleteProjectsQueryCall } from 'lib/apollo'
 import { NextRequest, NextResponse } from 'next/server'
 
 // get the handle name from a URL path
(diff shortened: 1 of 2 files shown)


tomquirk:
wdyt about this @wraeth-eth ?

src/middleware.page.ts
```diff
@@ -21,21 +19,21 @@ export async function middleware(request: NextRequest) {
...
-  let projects
+  let projectId
   try {
-    projects = await paginateDepleteProjectsQueryCall({
-      variables: {
-        where: { pv: PV_V2, handle: handleDecoded },
-      },
-    })
+    projectId = await fetch(
+      `${process.env.NEXT_PUBLIC_BASE_URL}/api/juicebox/project/${handleDecoded}`,
```
tomquirk, line 25: this fails on the preview link because theres no NEXT_PUBLIC_BASE_URL set :/

PR #3441, opened 2023-04-19
Use custom resolver for ENS node instead of default public resolver

peripheralist:
Previously the UI used the default PublicResolver contract for all ENS name contract calls. But if an ENS node is using a different Resolver contract, actions in the UI will fail to properly update the JBProjectHandles contract.

This PR adds a step in loading the Resolver contract, which first gets the correct resolver address using a call to the ENSRegistry contract https://docs.ens.domains/contract-api-reference/ens#get-resolver


src/hooks/ENS/contracts/PublicResolverAbi.json
```diff
@@ -1,5 +1,4 @@
 {
-  "address": "0x19c2d5D0f035563344dBB7bE5fD09c8dad62b001",
```
peripheralist, line 2: we no longer need to store an address for the Resolver contract. instead we use a contract call to get the address for a specific ENS node's resolver

tomquirk (approved):
@peripheralist awesome, niiice improvement

PR #3518, opened 2023-04-28
chore: webp script & package.json reference

filipviz:
Some things to know:
- The script searches and replaces based on the file's basename, meaning it will replace `foo.png` with `foo.webp` *everywhere* in src (including comments).
- The script writes any remaining references containing `png` `jpg` or `jpeg` to `remaining-references.txt`.
- By default, the script uses compression quality `85`. To use a different quality, change the first variable in the script (from 0-100).

The script requires `cwebp` to run. To install `cwebp` on...
- MacOS: brew install webp
- Ubuntu: sudo apt install webp
- Arch: sudo pacman -S libwebp

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,8 @@
     "supabase:diffdb": "supabase db diff",
     "supabase:generate:local": "supabase gen types typescript --local > src/types/database.types.ts",
     "cy:run": "cypress run",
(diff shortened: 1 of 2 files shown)


tomquirk:
@filipviz thanks for the sprint on this! Unfortuantely we have more problems than simply converting everything to webp - namely, resizing the source directly from Figma.

Gonna close this for now but let's revisit if we ever need it in future 🙏

PR #3519, opened 2023-04-28
fix: show token symbol on total issuance rate

filipviz:
Fix https://discord.com/channels/775859454780244028/876252266594721873/1101582750059872356


tomquirk (approved):
Nice catch

PR #3520, opened 2023-04-28
fix: Improve fees explanation

filipviz:
See https://discord.com/channels/939317843059679252/1017284843756847134/1101489973401747477


tomquirk (approved):
love it, thanks for this.

PR #3564, opened 2023-05-09
Fix duplicate split created on edit pay (JB-200)

johnnyd-eth:
Fixes JB-200 (https://linear.app/peel/issue/JB-200/duplicate-field-when-adding-new-token-beneficiary-in-create-flow)


johnnyd-eth:
@wraeth-eth lmk what you think of this. Would be great to give a good test if possible to make sure I haven't broken anything else 🙏

wraeth-eth (approved):

PR #3570, opened 2023-05-12
Apollo project events on V1/V2V3 dashboards

peripheralist:
This PR replaces subgraph queries for project activity on the V1 and V2V3 dashboards, with a single apollo client query.

- All ActivityEventElem `event` argument types have been replaced with generated graphql types
- Duplicated code for rending unique ActivityEventElems for different ProjectEvent types has been consolidated into a single AnyProjectEvent component. All ActivityEvent components have been moved to a single shared directly, where previously they were shared across v1/v2v3 project page directories
- `EventFilter` type for filtering a specific type of project event has been updated to match subgraph entity names, and pulled out from individual V1/V2V3 activity list components
- The apollo query is missing a few perks of the react-query infinite paging pattern: we lose `hasNextPage` and `isLoadingNextPage`. This has been solved using a custom `isFetchingMore` state on containers that render infinite-paging activity lists (both project dashboards). **I feel like we can bake this functionality into the apollo client which is very extensible, but not sure how yet.**

### DIFF
--- a/src/components/activityEventElems/ActivityElement/ActivityElement.tsx
+++ b/src/components/activityEventElems/ActivityElement/ActivityElement.tsx
@@ -77,9 +77,11 @@ export function ActivityEvent({
 }: {
   header: string | JSX.Element
   subject: string | JSX.Element | null
-  event: ActivityElementEvent
+  event: ActivityElementEvent | null | undefined
   extra?: string | JSX.Element | null
 }) {
+  if (!event) return null
+
   return (
     <>
       <div>
--- a/src/components/activityEventElems/AddToBalanceEventElem.tsx
+++ b/src/components/activityEventElems/AddToBalanceEventElem.tsx
@@ -1,19 +1,14 @@
 import { t } from '@lingui/macro'
 import ETHAmount from 'components/currency/ETHAmount'
 import RichNote from 'components/RichNote'
-import { AddToBalanceEvent } from 'models/subgraph-entities/vX/add-to-balance-event'
 
+import { ProjectEventsQuery } from 'generated/graphql'
 import { ActivityEvent } from './ActivityElement'
 
 export default function AddToBalanceEventElem({
   event,
 }: {
-  event:
-    | Pick<
-        AddToBalanceEvent,
-        'amount' | 'timestamp' | 'from' | 'note' | 'id' | 'txHash' | 'terminal'
-      >
-    | undefined
+  event: ProjectEventsQuery['projectEvents'][0]['addToBalanceEvent']
 }) {
   if (!event) return null
 
--- a/src/components/activityEventElems/AnyProjectEvent.tsx
+++ b/src/components/activityEventElems/AnyProjectEvent.tsx
@@ -1,0 +1,88 @@
+import { ProjectEventsQuery } from 'generated/graphql'
+
+import AddToBalanceEventElem from './AddToBalanceEventElem'
+import BurnEventElem from './BurnEventElem'
+import DeployedERC20EventElem from './DeployedERC20EventElem'
+import PayEventElem from './PayEventElem'
+import ProjectCreateEventElem from './ProjectCreateEventElem'
+import RedeemEventElem from './RedeemEventElem'
+import ReservesEventElem from './v1/ReservesEventElem'
+import TapEventElem from './v1/TapEventElem'
+import V1ConfigureEventElem from './v1/V1ConfigureEventElem'
+import ConfigureEventElem from './v2v3/ConfigureEventElem'
+import DeployETHERC20ProjectPayerEventElem from './v2v3/DeployETHERC20ProjectPayerEventElem'
+import DistributePayoutsElem from './v2v3/DistributePayoutsElem'
+import DistributeReservedTokensEventElem from './v2v3/DistributeReservedTokensElem'
+import SetFundAccessConstraintsEventElem from './v2v3/SetFundAccessConstraintsEventElem'
+
+/**
+ * Returns a formatted element for any project event
+ * @param event Project event object
+ * @param tokenSymbol Must be defined for projectEvents with a `burnEvent`
+ */
+export function AnyProjectEvent({
+  event,
+  tokenSymbol,
+}: {
+  event: ProjectEventsQuery['projectEvents'][number]
+  tokenSymbol: string | undefined
+}) {
+  if (event.addToBalanceEvent) {
+    return <AddToBalanceEventElem event={event.addToBalanceEvent} />
+  }
+  if (event.burnEvent) {
+    return <BurnEventElem event={event.burnEvent} tokenSymbol={tokenSymbol} />
+  }
+  if (event.configureEvent) {
+    return <ConfigureEventElem event={event.configureEvent} />
+  }
+  if (event.deployedERC20Event) {
+    return <DeployedERC20EventElem event={event.deployedERC20Event} />
+  }
+  if (event.deployETHERC20ProjectPayerEvent) {
+    return (
+      <DeployETHERC20ProjectPayerEventElem
+        event={event.deployETHERC20ProjectPayerEvent}
+      />
+    )
+  }
+  if (event.distributePayoutsEvent) {
+    return <DistributePayoutsElem event={event.distributePayoutsEvent} />
+  }
+  if (event.distributeReservedTokensEvent) {
+    return (
+      <DistributeReservedTokensEventElem
+        event={event.distributeReservedTokensEvent}
+      />
+    )
+  }
+  if (event.payEvent) {
+    return <PayEventElem event={event.payEvent} />
+  }
+  if (event.printReservesEvent) {
+    return <ReservesEventElem event={event.printReservesEvent} />
+  }
+  if (event.projectCreateEvent) {
+    return <ProjectCreateEventElem event={event.projectCreateEvent} />
+  }
+  if (event.redeemEvent) {
+    return <RedeemEventElem event={event.redeemEvent} />
+  }
+  if (event.setFundAccessConstraintsEvent) {
+    return (
+      <SetFundAccessConstraintsEventElem
+        event={event.setFundAccessConstraintsEvent}
+      />
+    )
+  }
+  if (event.tapEvent) {
+    return <TapEventElem event={event.tapEvent} />
+  }
+  if (event.v1ConfigureEvent) {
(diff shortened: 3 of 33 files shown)


src/components/activityEventElems/AddToBalanceEventElem.tsx
```diff
@@ -1,19 +1,14 @@
...
 export default function AddToBalanceEventElem({
   event,
 }: {
-  event:
-    | Pick<
-        AddToBalanceEvent,
-        'amount' | 'timestamp' | 'from' | 'note' | 'id' | 'txHash' | 'terminal'
-      >
-    | undefined
+  event: ProjectEventsQuery['projectEvents'][0]['addToBalanceEvent']
```
peripheralist, line 11: we now use generated types instead of the custom subgraph type with Pick<...>

by using the `ProjectEventsQuery[...][...]` type instead of the generated `AddToBalanceEvent` type, we can ensure this component's typing is directly dependent on the entity keys included in the ProjectEventsQuery graphql declaration

src/components/v1/V1Project/ProjectActivity/index.tsx
```diff
@@ -7,321 +7,120 @@ import DeployedERC20EventElem from 'components/activityEventElems/DeployedERC20E
...
-          'currency',
-          'ticketPrintingIsAllowed',
-          'payIsPaused',
-        ],
-      },
-    ],
-    orderDirection: 'desc',
-    orderBy: 'timestamp',
-    where,
+  const { data, fetchMore, loading } = useProjectEvents({
```
peripheralist, line 38: looooots of code replaced with simple apollo hook

src/components/v1/V1Project/ProjectActivity/index.tsx
```diff
@@ -7,321 +7,120 @@ import DeployedERC20EventElem from 'components/activityEventElems/DeployedERC20E
...
-            >
-              {elem}
-            </div>
-          )
-        }),
-      ),
+      projectEvents?.map(e => {
+        let elem: JSX.Element | undefined = undefined
+
+        if (e.payEvent) {
```
peripheralist, line 54: this rendering function as well as the one used in the v2v3 dashboard could be replaced with a single PV-agnostic function. leaving out of the scope of this PR

src/components/v1/V1Project/ProjectActivity/index.tsx
```diff
@@ -7,321 +7,120 @@ import DeployedERC20EventElem from 'components/activityEventElems/DeployedERC20E
...
+    if (count === 0 && !loading) {
       return (
         <div className="border-t border-smoke-200 pt-5 text-grey-500 dark:border-grey-600 dark:text-grey-300">
           <Trans>No activity yet</Trans>
         </div>
       )
     }
 
-    if (hasNextPage) {
+    if (count % PAGE_SIZE === 0) {
```
peripheralist, line 113: awkward. works, except for when the activity event count is divisible by PAGE_SIZE :/

src/components/v2v3/V2V3Project/ProjectActivity/eventElems/SetFundAccessConstraintsEventElem.tsx
```diff
@@ -33,7 +24,9 @@ export default function SetFundAccessConstraintsEventElem({
           ) : (
             <AmountInCurrency
               amount={event.distributionLimit}
-              currency={V2V3CurrencyName(event.distributionLimitCurrency)}
+              currency={V2V3CurrencyName(
+                event.distributionLimitCurrency as V2V3CurrencyOption,
```
peripheralist, line 28: ok Apollo also doesn't type `V2V3CurrencyOption`, but this is infrequently used and i don't mind using `as`. Otherwise we could update the pv override script to also override this property

src/components/v2v3/V2V3Project/ProjectActivity/hooks/useProjectActivity.ts
```diff
@@ -1,316 +0,0 @@
-import { PV_V2 } from 'constants/pv'
```
peripheralist, line 1: byeeeeee

src/graphql/documents/projectEvents.graphql
```diff
@@ -0,0 +1,168 @@
+query ProjectEvents(
```
peripheralist, line 1: new single source of truth for all project activity queries

src/graphql/documents/projectEvents.graphql
```diff
@@ -0,0 +1,168 @@
...
+      id
+      timestamp
+      txHash
+      from
+      beneficiary
+      beneficiaryTransferAmount
+      netTransferAmount
+    }
+    v1ConfigureEvent {
+      id
```
peripheralist, line 153: there doesn't seem to be a way to `fragment` all these repeated property names, since these are all unique entities

subgraph schema doesn't support entities that `implement` an interface

https://www.apollographql.com/docs/react/data/fragments/

src/hooks/useProjectEvents.ts
```diff
@@ -0,0 +1,79 @@
...
+type ProjectEventsQueryArgs = {
+  filter?: ProjectEventFilter
+  pv?: PV
+  wallet?: string
+  projectId?: number
+  skip?: number
+  first?: number
+}
+
+export function useProjectEvents({
```
peripheralist, line 39: light wrapper around the Apollo generated `useProjectEventsQuery` that simplifies arguments

src/lib/apollo/client.ts
```diff
@@ -34,7 +34,20 @@ const scalarsLink = withScalars({ schema, typesMap })
...
 
 const client = new ApolloClient({
-  cache: new InMemoryCache(),
+  cache: new InMemoryCache({
+    typePolicies: {
+      Query: {
+        fields: {
+          projectEvents: {
+            keyArgs: [],
+            merge(existing = [], incoming) {
```
peripheralist, line 43: necessary for `fetchMore` functionality. maybe more work to be done here

https://www.apollographql.com/docs/react/pagination/core-api

src/graphql/documents/projectEvents.graphql
tomquirk: @peripheralist could we fragment each event type though (e.g. `TapEvent`? I think that might work

src/hooks/useProjectEvents.ts
tomquirk: I like this a lot. I wonder if we should group all subgraph-related hooks in a directory. NBD for now.

src/components/activityEventElems/ActivityElement/ActivityElement.tsx
```diff
@@ -77,9 +77,11 @@ export function ActivityEvent({
 }: {
   header: string | JSX.Element
   subject: string | JSX.Element | null
-  event: ActivityElementEvent
+  event: ActivityElementEvent | null | undefined
   extra?: string | JSX.Element | null
 }) {
+  if (!event) return null
```
wraeth-eth, line 83: Is it possible to use undefined instead of null? We use undefined everywhere in the codebase already so should try to keep the same I think

src/components/v2v3/V2V3Project/ProjectActivity/eventElems/SetFundAccessConstraintsEventElem.tsx
wraeth-eth: lgtm! lfg

wraeth-eth (approved):
Awesome work peri!

src/components/activityEventElems/ActivityElement/ActivityElement.tsx
peripheralist: i don't think functional components can return undefined? not seeing where we currently do that, i see lots of `return null`

src/components/activityEventElems/AnyProjectEvent.tsx
```diff
@@ -0,0 +1,88 @@
...
+ * @param tokenSymbol Must be defined for projectEvents with a `burnEvent`
+ */
+export function AnyProjectEvent({
+  event,
+  tokenSymbol,
+}: {
+  event: ProjectEventsQuery['projectEvents'][number]
+  tokenSymbol: string | undefined
+}) {
+  if (event.addToBalanceEvent) {
```
peripheralist, line 30: this is consolidation of logic that was largely duplicated across v1 and v2v3 dashboard pages. not directly necessary for the apollo integration, but felt like the right time to do this and "complete" all project activity-related cleanup on both dashboards

src/components/v1/V1Project/ProjectActivity/index.tsx
```diff
@@ -1,307 +1,61 @@
...
-  | 'deployERC20'
-  | 'projectCreate'
-  | 'configure'
-
 const PAGE_SIZE = 10
 
 export function V1ProjectActivity() {
   const { projectId } = useContext(ProjectMetadataContext)
   const { tokenSymbol } = useContext(V1ProjectContext)
+  const [isFetchingMore, setIsFetchingMore] = useState<boolean>()
```
peripheralist, line 21: custom addition to replicate `isFetchingNextPage` functionality of `useQuery`, referenced in PR description

src/components/v1/V1Project/ProjectActivity/index.tsx
```diff
@@ -1,307 +1,61 @@
...
-            </div>
-          )
-        }),
-      ),
+      projectEvents?.map(e => (
+        <div
+          className="mb-5 border-b border-smoke-200 pb-5 dark:border-grey-600"
+          key={e.id}
+        >
+          <AnyProjectEvent event={e} tokenSymbol={tokenSymbol} />
```
peripheralist, line 51: this component replaces the big chunk of code above, for both dashboards

src/graphql/documents/projectEvents.graphql
peripheralist: @tomquirk not sure I know what you mean, but if you're talking about creating a fragment on TapEvent that would apply to all these sub-events i.e. TapEvent, that's what i've tried and doesn't work. to do so, all sub-events would need to implement a similar interface, which the subgraph compiler doesn't let us do :c

peripheralist: we could fragment TapEvent and apply it to all TapEvent references in our graphql files, but so far this is the only place we're referencing TapEvent

PR #3573, opened 2023-05-15
Show amounts in Edit Payouts modal

johnnyd-eth:
https://github.com/jbx-protocol/juice-interface/assets/96150256/fd98b3cf-303b-4ec7-8a86-ffed5b58fc57

Fixes JB-91


src/components/v2v3/shared/Allocation/components/PercentageInput.tsx
```diff
@@ -15,6 +21,17 @@ export const PercentageInput = ({
...
     isPercent: true,
   })
+
+  const { totalAllocationAmount, allocationCurrency } =
+    Allocation.useAllocationInstance()
+
+  const totalAllocationAmountIsInfinite = useMemo(
+    () =>
+      !totalAllocationAmount ||
+      totalAllocationAmount.eq(MAX_DISTRIBUTION_LIMIT),
```
tomquirk, line 31: can use `isInfiniteDistributionLimit` maybe

tomquirk (approved):
Thanks @johnnyd-eth. I don't love that we have to do this, but I think this is def a net improvement.

Do we have a linear issue for the following: "Allow payouts to be edited by amounts (when dist limit isnt inifinte)"?

johnnyd-eth:
Yep, working on it next: https://linear.app/peel/issue/JB-336/redesign-of-edit-payouts-modal-to-include-before-and-after-amounts

PR #3574, opened 2023-05-15
Fix loading pay button not full width

johnnyd-eth:
Fixes JB-332

Problem was that AntD overrides the given `w-full` className when button is disabled. Quick and easy fix is just using inline style in this case. 

https://github.com/jbx-protocol/juice-interface/assets/96150256/df4d85c3-706c-4c95-afe8-07430549926d


tomquirk (approved):
done deal! nice catch ser

PR #3585, opened 2023-05-16
fix: ballot strategy translation bug

tomquirk:


src/constants/v2v3/ballotStrategies/index.ts
```diff
@@ -116,7 +116,3 @@ export function ballotStrategiesFn(network?: NetworkName): BallotStrategy[] {
     },
   ]
 }
-
-export const ballotStrategies = ballotStrategiesFn()
```
tomquirk, line 120: trans strings above ^ were being evaluated outside the i18n provider scope. BUGZ

PR #3591, opened 2023-05-18
Replace useSubgraphQuery instances with Apollo queries

peripheralist:
Replace all subgraph query instances with Apollo queries, excluding:
- projects queries—these should be replaced with database queries
- infinite queries—these will be replaced later once an infinite apollo query pattern is in place

Logic behind organization of .graphql queries and files:
- We should define a new .graphql query anytime a unique set of property keys or arguments are needed. For example, if we were querying Projects in multiple places, but each use case required different Project properties, or one required filtering by PV while another doesn't. This avoids cases where a single "use-everywhere" query instance would return more properties than are necessary in most cases.
- Variables should be defined as strictly as possible for the use case of a query—i.e. accepting only a `wallet` argument in the WalletContributions query, vs. an entire `Participant_filter` object.
- Graphql queries should be defined in files named after the query, and organized into directories based on the entity they return.
- Generated graphql queries may be wrapped in hooks when additional logic or restrictions are needed that may be impossible or cumbersome to define in the graphql file (e.g. `useProjectEvents`)


src/components/v2v3/V2V3Project/ProjectActivity/hooks/useProjectActivity.ts
peripheralist, line 1: dense subgraph hook replaced by .graphql query

src/graphql/documents/query.graphql
peripheralist, line 1: deleted this to replace contained queries by separate files. previous "Projects" query was unused and hasn't been replaced

src/hooks/JB721Delegate/useNftAccountBalance.ts
```diff
@@ -10,23 +11,13 @@ export function useNftAccountBalance({
...
-          ],
-        }
-      : null,
-  )
+  return useJb721DelegateTokensQuery({
+    client,
+    variables: {
+      where: {
+        ...(dataSourceAddress ? { address: dataSourceAddress } : {}),
+        ...(accountAddress ? { owner: accountAddress } : {}),
```
peripheralist, line 19: wrapping this .graphql query in a hook makes it easier to apply conditional variables at the component end

tomquirk (approved):
@peripheralist LGTM and thanks for laying out dot points in description - all makes a ton of sense.

tomquirk (approved):
@peripheralist LGTM and thanks for laying out dot points in description - all makes a ton of sense.

PR #3594, opened 2023-05-19
Fix NFT modal console error in create flow

johnnyd-eth:
Fixes JB-226

- Problem was that `Form.Item` was passing `UploadNoProps` value, and `Upload` doesn't want it
- Changed `UploadNoProps` to still take value, use it where necessary, but not pass it to `Upload`


tomquirk (approved):
nice, good one

PR #3619, opened 2023-05-25
DBProjects updates + Owned/Created projects on account page

peripheralist:
Another PR in a sequence to deprecate all subgraph utils from the UI.

This began as replacing the `useMyProjects()` subgraph query hook with a DB query. However there were still some missing (and some recently added) Project properties that haven't been added to the database, which were required to make this query, so this PR also updates the DBProject model and update routine to store all Project properties that currently exist in the subgraph.

Also includes a very simple update to the account page to show both "Created" and "Owned" projects (until a recent subgraph update, only `Project.owner` was available, but we can now also make use of `Project.creator`)

Note on this pattern update: We can still query projects from the subgraph using the new Apollo queries. However the ProjectCard component uses a data type that requires DBProject properties (resolved metadata properties) which can't be retrieved from the subgraph. This is specifically relevant to this update since the query results are being passed to ProjectCard lists, but I intend to replace all Projects queries with DB queries as a general practice.

image
image

### DIFF
--- a/src/components/AccountDashboard/AccountDashboard.tsx
+++ b/src/components/AccountDashboard/AccountDashboard.tsx
@@ -19,7 +19,7 @@ import {
   useWalletContributionsQuery,
 } from 'generated/graphql'
 import useMobile from 'hooks/useMobile'
-import { useMyProjectsQuery } from 'hooks/useProjects'
+import { useDBProjectsQuery } from 'hooks/useProjects'
 import { useWalletSignIn } from 'hooks/useWalletSignIn'
 import { useWallet } from 'hooks/Wallet'
 import { client } from 'lib/apollo/client'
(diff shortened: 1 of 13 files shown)


src/components/AccountDashboard/AccountDashboard.tsx
```diff
@@ -112,15 +112,55 @@ function ContributedList({ address }: { address: string }) {
   )
 }
 
-function MyProjectsList({ address }: { address: string }) {
```
peripheralist, line 115: Splits previous `MyProjectsList` into two components: `OwnedProjectsList` and `CreatedProjectsList` which show projects Owned and Created by the wallet. Both now use the `useDBProjectsQuery()` hook, instead of a subgraph query.

src/hooks/useProjects.ts
```diff
@@ -319,28 +320,6 @@ function useProjectsOfParticipants(where: ProjectsOfParticipantsWhereQuery) {
   }
 }
 
-export function useMyProjectsQuery(wallet: string | undefined) {
```
peripheralist, line 322: replaced with `useDBProjects()` query

src/lib/api/supabase/projects/api.ts
```diff
@@ -77,6 +77,8 @@ export async function queryDBProjects(
   if (opts.archived) query = query.is('archived', true)
   else query = query.not('archived', 'is', true)
   if (opts.pv?.length) query = query.in('pv', opts.pv)
+  if (opts.owner) query = query.ilike('owner', opts.owner)
+  if (opts.creator) query = query.ilike('creator', opts.creator)
```
peripheralist, line 81: the /projects api endpoint now supports filtering projects by `owner` and `creator`

src/models/subgraph-entities/vX/project.ts
```diff
@@ -24,6 +24,9 @@ export type Project = {
   projectId: number
   pv: PV
   owner: string
+  creator: string
+  contributorsCount: number
+  nftsMintedCount: number
```
peripheralist, line 29: subgraph entities will be deprecated soon, but to avoid making this PR any thicker I've just updated the current subgraph Project entity type for compatibility, which is still used in the DB update routine

src/utils/sgDbProjects.ts
```diff
@@ -23,6 +23,14 @@ export const sgDbCompareKeys: SGSBCompareKey[] = [
   'metadataUri',
```
peripheralist, line 23: updates to this file are just allowing us to store newly added (and previously missing) Project properties in the DB

supabase/migrations/20230525200130_add_project_columns.sql
```diff
@@ -0,0 +1,11 @@
+alter table public.projects 
+add COLUMN "owner" char(42) not null,
+add COLUMN "creator" char(42) not null,
+add COLUMN "contributors_count" int not null,
+add COLUMN "redeem_count" int not null,
+drop column "total_paid",
```
peripheralist, line 6: this column was just renamed to "volume". `rename` not working, so we just drop and add the new "volume" column

tomquirk (approved):

PR #3646, opened 2023-06-02
feat: new project header

tomquirk:
| state | img | 
| --- | --- |
| everything | Screenshot 2023-06-06 at 2 11 33 PM |
| mobile | Screenshot 2023-06-06 at 2 07 53 PM |
| project with "infinity" growth and no handle | Screenshot 2023-06-06 at 2 08 00 PM |
| project handle tooltip | Screenshot 2023-06-06 at 2 10 33 PM |


src/components/ProjectDashboard/components/ProjectHeader/ProjectHeader.tsx
```diff
@@ -34,25 +28,30 @@ export const ProjectHeader = () => {
...
         </h1>
       </div>
-      <div className="flex justify-between">
+
+      <div className="flex flex-col justify-between gap-8 md:flex-row md:gap-12">
         <div className="flex flex-col gap-7">
           <div className="text-lg text-grey-700 dark:text-slate-50">
-            {subtitle}
+            {subtitle ? (
+              <Paragraph description={subtitle} characterLimit={100} />
```
tomquirk, line 36: just for now

tomquirk:
@wraeth-eth sending your wei for review 🙏

src/components/ProjectDashboard/hooks/useProjectHeader.ts
```diff
@@ -1,12 +1,42 @@
...
+import { useProjectMetadata } from './useProjectMetadata'
+
+export interface ProjectHeaderData {
+  title: string | undefined
+  subtitle: string | undefined
+  handle: string | undefined
+  projectId: number | undefined
+  owner: string | undefined
+  payments: number | undefined
+  totalVolume: BigNumber | undefined
```
wraeth-eth, line 14: Avoid passing out BigNumber - I am trying to cut back on Big Number being exposed in our code where possible. It's mostly confusing and error prone

wraeth-eth: lemme know what you think - happy to discuss this, but I feel pretty strongly that us passing BigNumber everywhere is an antipattern

tomquirk: im down - what do you suggest instead?

wraeth-eth: transforming data (kind of like a presenter) before passing back out to the component to be used. So for instance, totalVolume would be a string of something like "$1,000,000USD" or whatever we're doing, depending on currency

wraeth-eth (approved):

src/components/ProjectDashboard/hooks/useProjectHeader.ts
tomquirk: @wraeth-eth I don't mind this but I'm also not entirely clear on the benefit for this case.

I want to continue using ETHAmount and USDAmount where possible, for maximum consistency throughout the app (has USD tooltip, formatted correctly etc).

```
    <ProjectHeaderStat
        label={t`Total volume`}
        stat={<ETHAmount amount={totalVolume} precision={2} />}
      />
      ```

I might make the argument for thinking of data in the opposite direction: keep the data unaltered from source (chain) until the component layer, which choses to transform the data based on its desired behavior.

Think im gonna keep as-is for now but keen to continue discussion - i do hate bignumber.

PR #3668, opened 2023-06-05
Replace remaining `querySubgraph` and `querySubgraphExhaustive` instances

peripheralist:
- Add a bunch of new .graphql queries based on existing file pattern, and replace remaining `querySubgraph` and `querySubgraphExhaustive` instances
- Updates to `paginateDepleteQuery` for apollo queries, which replaces `useSubgraphExhaustive`

### DIFF
--- a/codegen.yml
+++ b/codegen.yml
@@ -18,3 +18,6 @@ config:
     enumValues: keep
   scalars:
     BigInt: BigNumber
+  avoidOptionals:
+    field: true
+  skipTypename: true
--- a/src/lib/apollo/paginateDepleteProjectsQuery.ts
+++ b/src/lib/apollo/paginateDepleteProjectsQuery.ts
@@ -3,6 +3,8 @@ import {
   ProjectsQuery,
   QueryProjectsArgs,
 } from 'generated/graphql'
+
+import { paginateDepleteQuery } from './paginateDepleteQuery'
 import { serverClient } from './serverClient'
 
 /**
@@ -15,25 +17,9 @@ export async function paginateDepleteProjectsQueryCall({
 }: {
   variables: QueryProjectsArgs
 }) {
-  const first = variables.first ?? 1000
-  const skip = variables.skip
-
-  const { data } = await serverClient.query<ProjectsQuery, QueryProjectsArgs>({
-    query: ProjectsDocument,
-    variables: { ...variables, first, skip },
+  return paginateDepleteQuery<ProjectsQuery, QueryProjectsArgs>({
+    client: serverClient,
+    document: ProjectsDocument,
+    variables,
   })
-  let projects = data.projects
-
-  if (projects.length) {
-    const _projects = await paginateDepleteProjectsQueryCall({
-      variables: {
-        ...variables,
-        first,
-        skip: first + (skip ?? 0),
-      },
-    })
-    projects = projects.concat(_projects)
-  }
-
-  return projects
 }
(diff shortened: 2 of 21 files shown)


codegen.yml
```diff
@@ -17,4 +17,7 @@ config:
   namingConvention:
     enumValues: keep
   scalars:
-    BigInt: BigNumber
\ No newline at end of file
+    BigInt: BigNumber
+  avoidOptionals:
+    field: true
+  skipTypename: true
```
peripheralist, line 23: - `avoidOptionals` prevents typing some things as `undefined` that in practice will only ever be defined or `null`
- `skipTypename` avoids expecting a `typename: string` property in Apollo responses, which creates some type compatibility issues

src/lib/apollo/paginateDepleteProjectsQuery.ts
```diff
@@ -15,25 +17,9 @@ export async function paginateDepleteProjectsQueryCall({
 }: {
   variables: QueryProjectsArgs
 }) {
-  const first = variables.first ?? 1000
-  const skip = variables.skip
-
-  const { data } = await serverClient.query<ProjectsQuery, QueryProjectsArgs>({
-    query: ProjectsDocument,
-    variables: { ...variables, first, skip },
+  return paginateDepleteQuery<ProjectsQuery, QueryProjectsArgs>({
```
peripheralist, line 20: this is just a wrapper for the frequently-used projects-`paginateDepleteQuery` that makes reuse cleaner

src/lib/apollo/paginateDepleteQuery.ts
```diff
@@ -0,0 +1,65 @@
...
+import { DocumentNode } from 'graphql'
+
+/**
+ * Loads all pages for the query.
+ * @param client ApolloClient to use for query.
+ * @param document GraphQL docuemnt to define query.
+ * @param variables Variables to apply to query.
+ * @returns
+ */
+export async function paginateDepleteQuery<
```
peripheralist, line 12: new generic paginateDepleteQuery that replaces `useSubgraphExhaustive`

src/utils/csvDownloadHelpers.ts
```diff
@@ -1,8 +1,35 @@
...
+} from 'generated/graphql'
+import { client } from 'lib/apollo/client'
+import { paginateDepleteQuery } from 'lib/apollo/paginateDepleteQuery'
 import { PV } from 'models/pv'
 import { downloadCsvFile } from 'utils/csv'
 import { fromWad } from 'utils/format/formatNumber'
-import { querySubgraphExhaustive } from 'utils/graph'
 import { emitErrorNotification } from 'utils/notifications'
 
 export async function downloadParticipants(
```
peripheralist, line 35: just replacing a lot of subgraph queries with apollo queries

src/utils/csvDownloadHelpers.ts
```diff
@@ -295,7 +281,7 @@ export async function downloadV1Payouts(
       rows.push([
         date,
         fromWad(p.modCut),
-        fromWad(p.modCutUSD),
+        p.modCutUSD ? fromWad(p.modCutUSD) : 'n/a',
```
peripheralist, line 284: although it shouldn't ever be in practice, `amountUSD` is nullable in the subgraph and correctly typed via Apollo, so we handle that type case here

codegen.yml
tomquirk: ay

tomquirk (approved):

PR #3670, opened 2023-06-05
Remove all deprecated Subgraph types + utils 🚮

peripheralist:
End of an era
image


tomquirk (approved):
absolutely heroic. lfg <3

PR #3672, opened 2023-06-06
Use reducer for NFT->payMetadata hook

johnnyd-eth:
Fixes JB-383

- Converts the management of the selection and deselection of NFT tiers for `payMetadata?.tierIdsToMint` from a state to a reducer function
- Function updates `payMetadata?.tierIdsToMint` in response to actions `SELECT_TIER` and 
  `DESELECT_TIER`

**Why:**
- More readable 
- Better performance and less unnecessary re-rendering

### DIFF
--- a/src/components/ProjectDashboard/components/NftRewardsPanel/hooks/nftPayMetadataReducer.ts
+++ b/src/components/ProjectDashboard/components/NftRewardsPanel/hooks/nftPayMetadataReducer.ts
@@ -1,0 +1,46 @@
+export const ACTIONS = {
+  SELECT_TIER: 'select_tier',
+  DESELECT_TIER: 'deselect_tier',
+}
+
+interface State {
+  tierIdsToMint: number[]
+}
+
+interface Action {
+  type: string
+  tierId?: number
+  quantity?: number
+}
+
+export function nftPayMetadataReducer(state: State, action: Action): State {
+  switch (action.type) {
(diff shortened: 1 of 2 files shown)


wraeth-eth:
Hey @johnnyd-eth this is getting close to what I mean for the reducer.
Here is an example of a reducer that I wrote for the cart: https://github.com/jbx-protocol/juice-interface/pull/3665/files#diff-af0f8315350a031ad00b8e4d129621933c5d2d1c8179466c9fd364b5ab64b5f1R1

What i've tried to do is take all the user interaction with UI and make them actions for the reducer, ie addPayment, removePayment, removeTokens, etc.

Dispatch is then used in code to call the action. for example, in `usePayProjectCard`:
```
  const addPay = useCallback(
    (
      values: PayProjectCardValues,
      formikHelpers: FormikHelpers<PayProjectCardValues>,
    ) => {
      dispatch({
        type: 'addPayment',
        payload: {
          amount: values.payAmount.amount,
          currency: values.payAmount.currency,
        },
      })
      formikHelpers.resetForm({
        values: {
          payAmount: {
            amount: undefined as unknown as number,
            currency: values.payAmount.currency,
          },
        },
      })
    },
    [dispatch],
  )
```

My feeling about this is that as soon as two states change on a callback, we should switch to using a reducer to break up complex logic.

PR #3673, opened 2023-06-06
Bookmarked projects

peripheralist:
## Bookmarked projects are here!

- "Saved" terminology used in UI
- new "user_bookmarks" DB table
- new "Saved projects" tab on account page
- new "Save project" button added next to current Subscribe button in V2V3ProjectHeader
- add saved icon to ProjectCard and TrendingProjectCard

### Considerations
- `useWalletBookmarkedProjects()` hook may need more love
- would be cool to show the bookmarked icon on ProjectCard components
- would be cool to show "save count" somewhere on project page
- we should also support saving V1 projects

image
image
image
image
image
image
image

### DIFF
--- a/src/components/BookmarkButton/hooks/useBookmarkButton.ts
+++ b/src/components/BookmarkButton/hooks/useBookmarkButton.ts
@@ -34,9 +34,9 @@ export const useBookmarkButton = ({
 
   const getIsBookmarked = useCallback(
     async ({ project, userId }: { project: string; userId: string }) => {
-      const { data, error } = await supabase
+      const { error, count } = await supabase
         .from('user_bookmarks')
-        .select('*')
+        .select('*', { count: 'exact', head: true })
         .eq('user_id', userId)
         .eq('project', project)
 
@@ -45,7 +45,7 @@ export const useBookmarkButton = ({
         return
       }
 
-      return data.length > 0
+      return !!count
     },
     [supabase],
   )


src/components/BookmarkButton/hooks/useBookmarkButton.ts
```diff
@@ -0,0 +1,137 @@
...
+
+/**
+ * Hook to control the bookmark button.
+ * @param projectId The projectId of the project to bookmark.
+ * @param pv The PV of the project to bookmark.
+
+ * @returns The loading state, the bookmarked state, and the bookmark function.
+ * @example const { loading, isBookmarked, bookmark } = useBookmarkButton({ project: '1-17' })
+ */
+export const useBookmarkButton = ({
```
peripheralist, line 18: loosely copied from `useSubscribeButton`

src/components/SubscribeButton/SubscribeButton.tsx
```diff
@@ -34,11 +33,7 @@ const _SubscribeButton = ({
           onClick={onSubscribeButtonClicked}
           loading={loading}
           icon={<SubscribeButtonIcon isSubscribed={isSubscribed} />}
-        >
-          <span className="font-base text-sm">
-            <Trans>Get Updates</Trans>
```
peripheralist, line 39: Removed this text to reduce clutter in the header after adding the second Bookmark button

src/hooks/useWalletBookmarkedProjects.ts
```diff
@@ -0,0 +1,53 @@
...
+          ids,
+          pageSize: 1000,
+        }
+      : null,
+  )
+
+  // Sort projects by timestamp of when they were saved
+  return {
+    projects: projects?.sort((a, b) =>
+      createdAtIds[a.id] < createdAtIds[b.id] ? -1 : 1,
```
peripheralist, line 49: this whole pattern feels a little hacky, may revisit it if i can find a better way to make it work

the goal is to sort results from the Projects table by the `user_bookmarks.created_at` property, so that we can show most recently saved projects first

src/pages/projects/ProjectsTabs.tsx
```diff
@@ -30,6 +30,12 @@ export default function ProjectsTabs({
 }: {
   selectedTab: ProjectCategory
 }) {
+  const { userAddress } = useWallet()
+
+  const TABS: ProjectCategory[] = ['all', 'trending', 'new']
+
+  if (userAddress) TABS.push('saved')
```
peripheralist, line 37: only show "saved" tab if a wallet is connected

src/hooks/useWalletBookmarkedProjects.ts
```diff
@@ -0,0 +1,53 @@
...
+          ids,
+          pageSize: 1000,
+        }
+      : null,
+  )
+
+  // Sort projects by timestamp of when they were saved
+  return {
+    projects: projects?.sort((a, b) =>
+      createdAtIds[a.id] > createdAtIds[b.id] ? -1 : 1,
```
peripheralist, line 49: this feels a bit hacky, may revisit later. goal is to sort Projects by the `user_bookmarks.created_at` value

for this reason we also don't use infinite paging here and instead default to a pageSize of 1000

supabase/migrations/20230606180645_add_bookmarked_projects.sql
```diff
@@ -0,0 +1,24 @@
+create table public.user_bookmarks (
+    "project" text not null,
```
tomquirk, line 2: could this be a foreign key to the `id` of the `projects` table?

```suggestion
  project_id bigint not null references public.projects(id),
```

src/hooks/useWalletBookmarkedProjects.ts
```diff
@@ -0,0 +1,53 @@
...
+  wallet,
+}: {
+  wallet: string | undefined
+}) {
+  const [createdAtIds, setCreatedAtIds] = useState<{ [id: string]: number }>({})
+
+  const supabase = useSupabaseClient<Database>()
+
+  useEffect(() => {
+    supabase
```
tomquirk, line 17: how does this actually work under the hood? How are we able to execute an sql query on the frontend?

src/components/BookmarkButton/hooks/useBookmarkButton.ts
```diff
@@ -0,0 +1,137 @@
...
+
+      if (error) {
+        console.error(error)
+        return
+      }
+
+      return data.length > 0
+    },
+    [supabase],
+  )
```
tomquirk, line 51: **suggestion/question** is it better to wrap this in a `useQuery` so we can avoid the useEffect below?

src/hooks/useWalletBookmarkedProjects.ts
```diff
@@ -0,0 +1,53 @@
...
+export function useWalletBookmarkedProjects({
+  wallet,
+}: {
+  wallet: string | undefined
+}) {
+  const [createdAtIds, setCreatedAtIds] = useState<{ [id: string]: number }>({})
+
+  const supabase = useSupabaseClient<Database>()
+
+  useEffect(() => {
```
tomquirk, line 16: same here, wondering is a `useQuery` is the better approach

supabase/migrations/20230606180645_add_bookmarked_projects.sql
```diff
@@ -0,0 +1,24 @@
...
+    "user_id" uuid not null,
```
tomquirk, line 3: same with this, i think we could/should foriegn key this to users table

```
  user_id uuid not null references public.users(id),
```

supabase/migrations/20230606180645_add_bookmarked_projects.sql
```diff
@@ -0,0 +1,24 @@
...
+    "wallet" char(42) not null,
```
tomquirk, line 4: could we derive wallet from the users table instead? in other words, do we need this column?

supabase/migrations/20230606180645_add_bookmarked_projects.sql
tomquirk: edit: i see this is happening below: https://github.com/jbx-protocol/juice-interface/pull/3673/files#diff-2a0597f9a314f672eebbf92f1be5442b2a5a8f757b530aefde412d854cd5ffccR22

Might be clearer to bundle all the changes up into the single `create table`, but either works i guess!

src/hooks/useWalletBookmarkedProjects.ts
```diff
@@ -0,0 +1,53 @@
...
+  const ids = Object.keys(createdAtIds ?? {})
+
+  const { data: projects, isLoading } = useDBProjectsQuery(
+    ids.length
+      ? {
+          ids,
+          pageSize: 1000,
+        }
+      : null,
+  )
```
tomquirk, line 44: **suggestion** I think we can get all of this happening in the one query with a `join`. Not tested but something like this:

```
SELECT * FROM projects
WHERE user_bookmarks.wallet = {wallet_address}
INNER JOIN user_bookmarks ON user_bookmarks.project_id = projects.id
ORDER BY user_bookmarks.created_at
```

tomquirk (changes requested):
peri absolute beast mode! left some comments but on the whole lfgtm

supabase/migrations/20230606180645_add_bookmarked_projects.sql
peripheralist: yeah for sure. i had quickly copied much of this over from previous migrations to make sure i was getting syntax right, and forgot to go back over this for some polish

src/hooks/useWalletBookmarkedProjects.ts
peripheralist: ohhh thank you, this is what i needed. couldn't remember how to do this earlier

src/components/BookmarkButton/hooks/useBookmarkButton.ts
peripheralist: great call

src/hooks/useWalletBookmarkedProjects.ts
peripheralist: should we not be able to? i've only used it locally so maybe it would break in prod. can always create an api endpoint for it

src/hooks/useWalletBookmarkedProjects.ts
peripheralist: @tomquirk shiiiiit just figured out some neat stuff 
image

src/hooks/useWalletBookmarkedProjects.ts
tomquirk: im honestly just curious how this actually works haha. Like what's actually happening here? Does supabaseClient send this query to the server via http?

src/hooks/useWalletBookmarkedProjects.ts
tomquirk: yooo. clean af, look at thattt

src/hooks/useWalletBookmarkedProjects.ts
peripheralist: iiuc that's the job of supabaseClient. we configure it with supabase .env vars so it can make those requests

tomquirk: ya haha all good. i was just curious how it actulaly works. and it works how i expected: they build a query string out of the 'sql query', fire off a HTTP req to the supabase server and it gives back the db query results. 

Screenshot 2023-06-12 at 5 01 16 PM

src/components/BookmarkButton/hooks/useBookmarkButton.ts
```diff
@@ -0,0 +1,136 @@
...
+        .select('*')
+        .eq('user_id', userId)
+        .eq('project', project)
+
+      if (error) {
+        console.error(error)
+        return
+      }
+
+      return data.length > 0
```
tomquirk, line 48: **suggestion** kinda inconsequential at our scale but a `count` query might be better here:

https://stackoverflow.com/questions/65612167/how-to-get-count-in-supabase might work?

src/components/BookmarkButton/hooks/useBookmarkButton.ts
```diff
@@ -0,0 +1,136 @@
...
+  )
+
+  const toggleBookmark = useCallback(async () => {
+    // If the user is not connected to a wallet, don't do anything
+    if (!wallet.userAddress) return
+
+    const { data: users } = await supabase
+      .from('users')
+      .select('*')
+      .eq('wallet', wallet.userAddress)
```
tomquirk, line 60: **suggestion** slight performance boost:

```suggestion
      .eq('wallet', wallet.userAddress)
       .limit(1)
       .single()
```

src/components/BookmarkButton/hooks/useBookmarkButton.ts
```diff
@@ -0,0 +1,136 @@
...
+
+    let _session = session
+
+    if (!_session || !users?.[0]) {
+      _session = await signIn()
+    }
```
tomquirk, line 66: **question** does `session` already update when after `signIn()`

LT if not, then we might consider a new hook that wraps `useSession` and handles this logic for us. Outside this pr scope though

src/components/BookmarkButton/hooks/useBookmarkButton.ts
```diff
@@ -0,0 +1,136 @@
...
+
+  // Check if the project is bookmarked
+  useEffect(() => {
+    if (!session?.user.id) return
+
+    setLoading(true)
+
+    const userId = session.user.id
+
+    getIsBookmarked({ project, userId }).then(_isBookmarked => {
```
tomquirk, line 108: **question** feels like `getIsBookmarked` could also be a `useQuery` - wdyt? Think it'll let us nuke this `useEffect`.

src/hooks/useWalletBookmarkedProjects.ts
```diff
@@ -0,0 +1,56 @@
...
+  wallet: string | undefined
+}) {
+  const supabase = useSupabaseClient<Database>()
+
+  return useQuery(
+    ['user-bookmarks', wallet],
+    async (): Promise<DBProject[] | undefined> =>
+      supabase
+        .from('projects')
+        .select('*, user_bookmarks!inner(created_at, users(wallet))')
```
tomquirk, line 24: based af

tomquirk (approved):
@peripheralist fuck yeaaaasss. lfg. I left a few more comments about queries n such, but will pre-approve.

PR #3692, opened 2023-06-11
fix: nft rewards with same price

tomquirk:


src/utils/nftRewards.ts
```diff
@@ -344,35 +344,45 @@ export function buildJB721TierParams({
...
         if (
           (a as JB_721_TIER_PARAMS_V3_2).price.lt(
             (b as JB_721_TIER_PARAMS_V3_2).price,
           )
-        )
+        ) {
           return -1
+        }
+
+        return 0
```
tomquirk, line 367: the addition of this line is the fix :|

PR #3761, opened 2023-06-19
Re-add TransactionsList with updated styles

peripheralist:
Adds transactions icon + expandable list to the site navigation


wraeth-eth:
image
The loading icon looks a bit odd here imo. It looks like a failed image, maybe we could consider using a different icon for pending state? cc @strath-m

wraeth-eth:
I also think, if possible, can we put the transaction list as a part of the wallet drop down? The nav bar already has a lot of icons and I think we want to add less to it. We could add the count, and pending icon as a part of the wallet button potentially

src/components/Navbar/SiteNavigation.tsx
```diff
@@ -79,6 +80,10 @@ export default function SiteNavigation() {
                         <NavLanguageSelector className="md:order-2" />
                         <ThemePicker className="md:order-3" />
                         <QuickProjectSearch className="md:order-1" />
+                        <TransactionsList
+                          className="relative md:order-4"
```
wraeth-eth, line 84: I wonder if we can 'bake' these class names into the TransactionList itself. Do you think we'll need the flexibility to reuse the tx list with different classes elsewhere?

peripheralist:
> I also think, if possible, can we put the transaction list as a part of the wallet drop down? The nav bar already has a lot of icons and I think we want to add less to it. We could add the count, and pending icon as a part of the wallet button potentially

I tend to agree, but this was the design we eventually settled on in discord design chat 🤷

src/components/Navbar/SiteNavigation.tsx
peripheralist: good call, done

src/components/Navbar/components/TransactionList/TxStatusIcon.tsx
```diff
@@ -0,0 +1,41 @@
+import {
+  CheckCircleIcon,
+  ExclamationCircleIcon,
+} from '@heroicons/react/24/outline'
+import Loading from 'components/Loading'
+import { TxStatus } from 'models/transaction'
+import { twMerge } from 'tailwind-merge'
+
+export default function TxStatusIcon({ status }: { status: TxStatus }) {
+  const containerClassName =
```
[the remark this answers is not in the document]
peripheralist: into it, done

wraeth-eth (approved):

PR #3767, opened 2023-06-20
Update new project page NFT cards [2/2]

johnnyd-eth:
- Adds "Remove NFT" on hover when selected (Fixes JB-513)

https://github.com/jbx-protocol/juice-interface/assets/96150256/f8b80aad-b8f9-4fa2-9c19-33dbcb6298db

- Fixes gap in hover on selected (Fixes JB-527)

- Adds dropshadow to NFT card (Fixes JB-514)

### DIFF
--- a/src/components/ProjectDashboard/components/NftRewardsPanel/NftReward/NftReward.tsx
+++ b/src/components/ProjectDashboard/components/NftRewardsPanel/NftReward/NftReward.tsx
@@ -9,11 +9,13 @@ import { ipfsUriToGatewayUrl } from 'utils/ipfs'
 import { AddNftButton } from './AddNftButton'
 import { NftDetails } from './NftDetails'
 import { NftThumbnail } from './NftThumbnail'
+import { RemoveNftButton } from './RemoveNftButton'
 
 type NftRewardProps = {
   rewardTier?: NftRewardTier
   loading?: boolean
   onSelect: (quantity?: number) => void
+  onDeselect: VoidFunction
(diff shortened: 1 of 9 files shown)


src/components/ProjectDashboard/components/NftRewardsPanel/NftReward/NftReward.tsx
```diff
@@ -59,7 +62,7 @@ export function NftReward({
   return (
     <div
       className={twMerge(
-        'group relative flex h-full w-[252px] cursor-pointer select-none flex-col rounded-lg border border-grey-200 dark:border-slate-500',
+        'group relative flex h-full w-[252px] cursor-pointer select-none flex-col rounded-nft border border-grey-200 shadow-nftCard dark:border-slate-500',
```
wraeth-eth, line 65: I'd avoid making a custom config just for nft rounding, you can say `rounded-[10px]` instead for this.

I think you can do the same with shadow as:
`shadow-[0px_4px_14px_0px_rgba(0, 0, 0, 0.0392156862745098)]`
https://tailwindcss.com/docs/drop-shadow#arbitrary-values

johnnyd-eth: Yeah good call, done. Spent a while trying to figure out why the shadow class you gave wasn't working and it's because `(0, 0, 0, 0.0392156862745098)` has spaces in it 🤣 😢

PR #3831, opened 2023-06-28
Fix wallet connect modal appearing by transaction modals

johnnyd-eth:
Fixes JB-591 (Fixes JB-588)

https://github.com/jbx-protocol/juice-interface/assets/96150256/799fcbb9-2544-41a1-9864-9b39a3d6f706

### DIFF
--- a/src/styles/antd-overrides/modal.scss
+++ b/src/styles/antd-overrides/modal.scss
@@ -29,7 +29,7 @@
 
 .ant-modal-wrap,
 .ant-modal-mask {
-  z-index: 20 !important;
+  z-index: 19 !important;
 }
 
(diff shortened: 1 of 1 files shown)


src/styles/antd-overrides/modal.scss
```diff
@@ -29,7 +29,7 @@
 
 .ant-modal-wrap,
 .ant-modal-mask {
-  z-index: 20 !important;
+  z-index: 19 !important;
```
johnnyd-eth, line 32: Wallet connect z-index is 20 so made it 1 less, to minimise risk of anything else appearing above the ant modal

wraeth-eth (approved):

PR #3870, opened 2023-07-05
Keyp wallet

peripheralist:
This PR adds Keyp wallet as an option for connecting to the app. Keyp wallet will appear and function the same as a standard EOA wallet, with the exception of showing the keyp account username in the wallet component instead of the wallet address.

## Keyp wallet
[https://docs.usekeyp.com/](https://docs.usekeyp.com/)
Keyp provides an Ethereum wallet that is not custodied by the end user, but can be accessed via Keyp account authentication.

### Authentication
Connecting a Keyp wallet requires the user to authenticate with the Keyp API. The API then redirects back to the app, where user details (i.e. wallet address, username) are stored in the session using next-auth.

### Chain
Keyp wallets are cross-chain. For contract writes, the chain is specified in the configuration of the Keyp client. The client will be determined by a .env var `KEYP_CLIENT_ID`.

### useJBWallet() hook
Previously we've used `useWallet()` to get connected wallet details, like `userAddress` and `chain`. We should now rely on `useJBWallet()`, which wraps `useWallet()` as well as the connected Keyp wallet details via `useSession()`. The `useJBWallet().keyp.chain` object is simply a helper property for consistency with  that matches the chain of `readProvider`. 
> **90% of this PR is just replacing `useWallet()` instances.**

### WalletContext / WalletProvider
Previously we've used the `useWallet().connect` function to connect a wallet, which includes its own modal UI. We'll now use components provided in WalletProvider to connect a wallet, which includes a modal for choosing either an EOA wallet or a keyp wallet, then allowing the user to connect accordingly. By placing these components in a context provider at the top of the tree, we can enable the login sequence anywhere (i.e. from the Navbar button, the pay modal, or the transactor hook)

### Transactions
The Transactor automatically handles transactions from Keyp wallets alongside EOA wallets. Keyp wallet transactions are sent to the Keyp API, added to the TxHistoryContext, and tracked the same as transactions from EOA wallets. Components initiating transactions don't need any changes.

### Keyp db user authentication
Keyp wallet API does not support signatures. Because the user must authenticate via Keyp to connect their Keyp wallet, we bypass the wallet signature when creating or logging in a DB user with a Keyp wallet.


src/hooks/Wallet/useJBWallet.ts
```diff
@@ -0,0 +1,48 @@
+import { readNetwork } from 'constants/networks'
+import { WalletContext } from 'contexts/Wallet/WalletContext'
+import { signOut, useSession } from 'next-auth/react'
+import { useContext } from 'react'
+import { useWallet } from './useWallet'
+
+export function useJBWallet() {
```
peripheralist, line 7: new wrapper around `useWallet()` that includes EOA wallet connection as well as new keyp wallet

src/hooks/useTransactor.ts
```diff
@@ -48,6 +62,38 @@ function logTx({
...
+  //   { address },
+  //   { abi },
+  //   { accessToken },
+  //   contract.interface.fragments,
+  // )
+
+  return writeContract({
+    accessToken,
+    address,
+    abi: abiFunctionName(functionName, contract.interface),
```
peripheralist, line 91: we need to send the fn signature (i.e. `Pay(uint256, uint256, ...)`) instead of the full abi

src/contexts/Wallet/WalletProvider.tsx
```diff
@@ -0,0 +1,72 @@
+import { Trans } from '@lingui/macro'
+import { signInKeyp } from '@usekeyp/js-sdk'
+import { LoginPortal } from '@usekeyp/ui-kit'
+import { Button, Modal } from 'antd'
+import { useWallet } from 'hooks/Wallet'
+import { useCallback, useState } from 'react'
+
+import { WalletContext } from './WalletContext'
+
+export const WalletProvider: React.FC<React.PropsWithChildren<unknown>> = ({
```
peripheralist, line 10: new WalletProvider wraps the wallet connection component, allowing us to render the connect wallet UI from anywhere in the app (i.e. from any transactor instance if there's no wallet connected). previously we could just call `useWallet().connect()` which opened its own UI, but now we need a 2-step modal flow to allow a user to select a Keyp wallet or an EOA wallet

peripheralist: we'll want to replace all instances of `useWallet()` with `useJBWallet()`

src/hooks/useWalletSignIn.ts
```diff
@@ -50,15 +45,26 @@ export const useWalletSignIn = () => {
...
+        wallet: eoa.userAddress,
+      })
+      const signature = await eoa.signer.signMessage(challengeMessage)
+      accessToken = await AuthAPI.walletSignIn({
+        wallet: eoa.userAddress,
+        signature,
+        message: challengeMessage,
+      })
+    } else if (keyp.address) {
+      accessToken = await AuthAPI.keypWalletSignIn({ wallet: keyp.address })
```
peripheralist, line 62: @wraeth-eth here we handle signing in a keyp wallet without a challenge message

src/pages/api/auth/keyp-sign-in.page.ts
```diff
@@ -0,0 +1,35 @@
+import { sudoPublicDbClient } from 'lib/api/supabase/clients'
+import { NextApiRequest, NextApiResponse } from 'next'
+import { buildJwt, getOrCreateUser } from './utils'
+
+/**
+ * Sign in a keyp wallet without requiring a signature.
+ *
+ * Only POST requests are supported.
+ */
+const handler = async (req: NextApiRequest, res: NextApiResponse) => {
```
peripheralist, line 10: @wraeth-eth new api endpoint for signing in a wallet without a challenge message

tomquirk:
guessing we can close this ? :(

tomquirk:
@peripheralist gonna close, just doing some cleanup. Will pin this PR somewhere in Discord incase we need to come back and steal some stuff from it.

PR #3909, opened 2023-07-21
[3/3] Payouts table [BASE] - Use table to edit payout amounts

johnnyd-eth:
- Implements a payout table with **sub-total** (the sum of all the payouts that will actually be received by the recipients - "after fee"), **fee** amount and "Total" = `distributionLimit` = subTotal + feeAmount

https://github.com/jbx-protocol/juice-interface/assets/96150256/ed3cd297-bbb7-4711-84ba-da7f209153f3

**Still to do:**
- Add/Edit recipient modal (need to call you about this @wraeth-eth. I tried to implement `AddEditAllocationModal` but it started getting messy, so will leave it for the next PR after we plan it out better.
- Implement table designs for %
- Switch `% -> amount` modal


src/components/v2v3/V2V3Project/V2V3ProjectSettings/pages/NewEditCyclePage/PayoutsSection/PayoutsTable/PayoutsTable.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { PayoutsTableCell } from './PayoutsTableCell'
+import { PayoutsTableRow } from './PayoutsTableRow'
+
+const Row = PayoutsTableRow
+const Cell = PayoutsTableCell
+
+export function PayoutsTable() {
+  const { editCycleForm, initialFormData } = useEditCycleFormContext()
+
+  // useWatch is always undefined. Goal was have these update and trigger re-render when the field is updated in another component
```
johnnyd-eth, line 15: Start here @wraeth-eth

src/components/v2v3/V2V3Project/V2V3ProjectSettings/pages/NewEditCyclePage/PayoutsSection/PayoutsTable/PayoutSplitRow.tsx
```diff
@@ -0,0 +1,131 @@
...
+import { PayoutTitle } from './PayoutTitle'
+import { PayoutsTableCell } from './PayoutsTableCell'
+import { PayoutsTableRow } from './PayoutsTableRow'
+
+const Cell = PayoutsTableCell
+
+export function PayoutSplitRow({ payoutSplit }: { payoutSplit: Split }) {
+  const { editCycleForm } = useEditCycleFormContext()
+
+  // useWatch not loading
```
johnnyd-eth, line 27: If you switch these to `editCycleForm.getFieldValue` everything will load properly, but then when you go to edit the amount of one in the row, everything takes a while to update

peripheralist:
@johnnyd-eth what's the purpose of enforcing adding the prefix in the output of the parser? wouldn't we specifically _not_ want the prefix in the parser, so that the value is just the number value?

image

PR #3932, opened 2023-08-03
Filter invalid tags from project metadata

peripheralist:
Filter out any invalid tags from project metadata.

I noticed a bug (on goerli) where some older projects are using tags that are now "invalid" (not recognized by the UI) which cause breaking runtime errors. I suspect this issue is nonexistent in mainnet prod, but in the case of a project having invalid tags for any reason in the future, this will prevent them from making it to the UI and causing errors.

### DIFF
--- a/src/utils/sgDbProjects.ts
+++ b/src/utils/sgDbProjects.ts
@@ -1,7 +1,11 @@
 import { ipfsGet } from 'lib/api/ipfs'
 import { DBProject, DBProjectRow, SGSBCompareKey } from 'models/dbProject'
 import { Json } from 'models/json'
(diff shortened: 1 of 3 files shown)


src/utils/sgDbProjects.ts
```diff
@@ -185,6 +189,10 @@ export function getChangedSubgraphProjects({
         return true
       }
 
+      if (dbProject.tags?.some(t => !isValidProjectTag(t))) {
```
peripheralist, line 192: if a db project has any invalid tags, force it to update

wraeth-eth (approved):

PR #3941, opened 2023-08-04
fix: wrap non anchor links in anchor

wraeth-eth:

Closes [JB-682 : Links missing from project descriptions](https://linear.app/peel/issue/JB-682/links-missing-from-project-descriptions)

## Acceptance checklist

- [ ] (if relevant) I have tested this PR in [all supported browsers](https://github.com/jbx-protocol/juice-interface/blob/main/CONTRIBUTING.md#supported-browsers).
- [ ] (if relevant) I have tested this PR in dark mode and light mode (if applicable).

### DIFF
--- a/src/components/ProjectDashboard/hooks/useAboutPanel.ts
+++ b/src/components/ProjectDashboard/hooks/useAboutPanel.ts
@@ -23,12 +23,26 @@ export const useAboutPanel = () => {
     projectMetadata?.telegram,
     projectMetadata?.twitter,
   ])
-  const description = projectMetadata?.description
+  const description = useMemo(
+    () =>
+      projectMetadata?.description
+        ? wrapNonAnchorsInAnchor(projectMetadata?.description)
+        : undefined,
+    [projectMetadata?.description],
+  )
+
   return {
     description,
     socialLinks,
     projectName: projectMetadata?.name,
   }
+}
+
+const wrapNonAnchorsInAnchor = (text: string) => {
+  const urlRegex = /\b((http|https):\/\/?[\w\S(-.:#~!$&'()*+,;=%)]+)/g
+  return text.replace(urlRegex, url => {
+    return '<a href="' + url + '">' + url + '</a>'
+  })
 }
 
 const linkOrUndefined = (link: string | undefined) => {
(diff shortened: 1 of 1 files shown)


wraeth-eth:
* main
  * **PR #3941** <a href="https://app.graphite.dev/github/pr/jbx-protocol/juice-interface/3941" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jbx-protocol/juice-interface/3941?utm_source=stack-comment).

src/components/ProjectDashboard/hooks/useAboutPanel.ts
```diff
@@ -23,14 +23,28 @@ export const useAboutPanel = () => {
...
+
   return {
     description,
     socialLinks,
     projectName: projectMetadata?.name,
   }
 }
 
+const wrapNonAnchorsInAnchor = (text: string) => {
+  const urlRegex = /\b((http|https):\/\/?[\w\S(-.:#~!$&'()*+,;=%)]+)/g
```
filipviz, line 42: I think this regex is too permissive:
1. The `?` after the second slash means this would match http:/example.com (do we want to do this)
2. The `\S` in the bracket group means this regex will match `'` and `>` which could be dangerous if this text eventually goes into something that interprets the HTML. I think this would allow for arbitrary HTML (and by extension JS) execution. We should explicitly match reserved characters described in RFC 3986, along these lines:

```js
const urlRegex = /\b((http|https)://[a-zA-Z0-9-._~:/?#@\[\]!$&'()*+,;=%]+\b)/g
```

filipviz: I might be off on details, not an expert on this kind of thing.

https://www.rfc-editor.org/rfc/rfc3986

wraeth-eth: Thanks for the suggestion! Have updated. I am less worried about arbitrary code as we are sanitising after this, but still good practice none the less!

filipviz (approved):
> Thanks for the suggestion! Have updated. I am less worried about arbitrary code as we are sanitising after this, but still good practice none the less!

With this context, LGTM!

PR #3945, opened 2023-08-04
feat: add project logo api

tomquirk:
per title. used in:
- homepage project logos
- payouts

Can be added in other places later (like activity feed).

Please validate locally!

Notes:
- Cached for 1 day but that may be too long - might want to shorten.
- Some projects with massive logos may fail (cos of Next/vercel API limits)
- Open question: https://github.com/jbx-protocol/juice-interface/pull/3945#discussion_r1284519801

This is an MVP. An optimized solution would:
- replicate (send) images to some image CDN
- the image CDN would auto-downscale/size images 
- we could request whatever size of the image we want
- it would be cached, but we can refresh the cache on-demand, if the project changes their logo.

Screenshot 2023-08-04 at 4 12 25 PM

### DIFF
--- a/src/pages/api/juicebox/projectLogo/[projectId].page.ts
+++ b/src/pages/api/juicebox/projectLogo/[projectId].page.ts
@@ -1,0 +1,93 @@
+import { createServerSupabaseClient } from '@supabase/auth-helpers-nextjs'
+import axios from 'axios'
+import { getLogger } from 'lib/logger'
+import { NextApiRequest, NextApiResponse } from 'next'
+import { Database } from 'types/database.types'
+import {
+  cidFromUrl,
+  ipfsGatewayUrl,
+  ipfsUriToGatewayUrl,
+  isIpfsUri,
+} from 'utils/ipfs'
+import { getProjectMetadata } from 'utils/server/metadata'
+
+const logger = getLogger('api/juicebox/projectLogo/[projectId]')
+
+async function findLogoFromDb(
+  req: NextApiRequest,
+  res: NextApiResponse,
+  projectId: string,
+) {
+  const supabase = createServerSupabaseClient<Database>({ req, res })
+
+  const query = await supabase
+    .from('projects')
+    .select('logo_uri')
+    .eq('project_id', projectId)
+
+  if (query.error) throw query.error
+
+  return query.data?.[0]?.logo_uri as string | undefined
+}
+
+async function findLogoOnChain(projectId: string) {
+  const { logoUri } = (await getProjectMetadata(projectId as string)) ?? {}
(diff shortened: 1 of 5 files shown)


src/pages/api/juicebox/projectLogo/[projectId].page.ts
```diff
@@ -0,0 +1,93 @@
...
+  try {
+    const { projectId } = req.query
+    if (!projectId) {
+      return res.status(400).json({ error: 'projectId is required' })
+    }
+
+    // try getting logo from db. if for some reason its not there, try getting it from chain.
+    const logoUri =
+      (await findLogoFromDb(req, res, projectId as string)) ??
+      (await findLogoOnChain(projectId as string))
```
Comment (tomquirk) [inline src/pages/api/juicebox/projectLogo/[projectId].page.ts:64]: if the project legit doesnt have a logo, then we are doing 2 sync GETS for nothing.

Should we just remove this line and solely rely on the DB?

src/pages/api/juicebox/projectLogo/[projectId].page.ts
```diff
@@ -0,0 +1,93 @@
...
+ *
+ * Note, if DB request fails, we still try to fetch the logo from chain. This will be 2 more requests.
+ *
+ * Thus this endpoint should be cached.
+ * However, the ideal duration is TBD.
+ * Please consider changing (lowering) if it causes issues with users.
+ */
+const handler = async (req: NextApiRequest, res: NextApiResponse) => {
+  if (req.method !== 'GET') {
+    return res.status(404)
```
Comment (wraeth-eth) [inline src/pages/api/juicebox/projectLogo/[projectId].page.ts:52]: method not allowed `405`

Comment (wraeth-eth) [reply src/pages/api/juicebox/projectLogo/[projectId].page.ts:52]: I think you also need to include .json, or something else as well

src/pages/api/juicebox/projectLogo/[projectId].page.ts
Comment (wraeth-eth) [reply src/pages/api/juicebox/projectLogo/[projectId].page.ts:64]: I think so 👍

wraeth-eth (commented):
Comments - still need to test locally.

Love this @tomquirk , looks really good!!

wraeth-eth:
Just tested! So fresh!

src/pages/api/juicebox/projectLogo/[projectId].page.ts
Comment (tomquirk) [reply src/pages/api/juicebox/projectLogo/[projectId].page.ts:64]: rad. will do

src/pages/api/juicebox/projectLogo/[projectId].page.ts
Comment (tomquirk) [reply src/pages/api/juicebox/projectLogo/[projectId].page.ts]: great call, done

tomquirk:
Thanks @wraeth-eth for review! <3

PR #3946, opened 2023-08-05
fix: small regex correction

filipviz:
image
Currently matches periods, causing some 404s. Adding an extra `\b` (outside of parens) to match word boundaries (before punctuation).


filipviz:
image
Working now

PR #3954, opened 2023-08-07
Allow wad precision in JuiceInputNumber

peripheralist:
This PR removes the Antd number input from JuiceInputNumber and replaces it with custom logic. This avoids an issue where numbers were previously being limited to JS MAX_INT, presumably because the Antd component was doing some Number logic under the hood.

### DIFF
--- a/src/components/inputs/JuiceInputNumber.tsx
+++ b/src/components/inputs/JuiceInputNumber.tsx
@@ -17,11 +17,11 @@ export const JuiceInputNumber = ({
   value?: string
   formatter?: (s?: string | undefined) => string
   parser?: (s?: string | undefined) => string
-  onChange?: (s: string) => void
+  onChange?: (s?: string) => void
   className?: string
-} & DetailedHTMLProps<
-  InputHTMLAttributes<HTMLInputElement>,
-  HTMLInputElement
+} & Omit<
+  DetailedHTMLProps<InputHTMLAttributes<HTMLInputElement>, HTMLInputElement>,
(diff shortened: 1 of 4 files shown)


src/components/inputs/JuiceInputNumber.tsx
```diff
@@ -1,14 +1,47 @@
...
+
   return (
-    <InputNumber
-      {...props}
+    <input
+      value={val}
       className={twMerge(
-        'border-smoke-300 bg-smoke-50 text-black dark:border-slate-300 dark:bg-slate-600 dark:text-slate-100 dark:placeholder:text-slate-300',
-        props.className,
+        'text-primary stroke-secondary rounded-lg border border-smoke-300 bg-smoke-50 px-4 py-2 text-black dark:border-slate-300 dark:bg-slate-600 dark:text-slate-100 dark:placeholder:text-slate-300',
```
peripheralist, line 36: adds some styling that was previously defined in css for .ant-number-input

johnnyd-eth:
Code looks dope @peripheralist! Just gave it a whirl and noticed typing it decimals points don't seem to be working anywhere. LGTM outside of that

johnnyd-eth (changes requested):
(Decimals not working)

peripheralist:
thanks for taking a careful look @johnnyd-eth 🙏 fixed now

image

johnnyd-eth (approved):
Sweet, LGTM then!

PR #3967, opened 2023-08-09
Add tools drawer to new v2v3 dashboard

peripheralist:
image
image


wraeth-eth:
Awesome PR!! 



looks like the button is mis-matched in dark mode

wraeth-eth:
It is also missing on mobile, but that might be intentional

wraeth-eth (changes requested):
above comments 🙏

peripheralist:
thanks @wraeth-eth , fixed.

wraeth-eth (approved):

PR #3987, opened 2023-08-12
fix: combine pay checkboxes

filipviz:
Combine the pay notice checkboxes as per [Strath's recommendation](https://discord.com/channels/939317843059679252/1060691268759462028/1139389906565742693).

I haven't combined the checkboxes in `src/components/v2v3/V2V3Project/V2V3PayButton/V2V3ConfirmPayModal/V2V3PayForm.tsx` (the old project page). Can make the change if helpful.

### DIFF
--- a/src/components/ProjectDashboard/components/PayProjectModal/PayProjectModal.tsx
+++ b/src/components/ProjectDashboard/components/PayProjectModal/PayProjectModal.tsx
@@ -138,7 +138,7 @@ export const PayProjectModal: React.FC = () => {
                         )}
                       >
                         <Trans>
-                          I understand and accept{' '}
+                          I understand and accept
                           {projectPayDisclosure
                             ? "this project's notice and "
                             : ' '}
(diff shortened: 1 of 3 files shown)


src/components/ProjectDashboard/components/PayProjectModal/PayProjectModal.tsx
```diff
@@ -139,45 +138,18 @@ export const PayProjectModal: React.FC = () => {
                         )}
                       >
                         <Trans>
-                          I accept the{' '}
+                          I understand and accept
```
wraeth-eth, line 141: This might have unintended issues with translations - could we split it into two separate strings?

wraeth-eth (changes requested):

src/components/ProjectDashboard/components/PayProjectModal/PayProjectModal.tsx
tomquirk: not that it matters too much, but for future ref: strings marked for translation should always be whole sentences. 

made the change here:
cc @filipviz

tomquirk (approved):

src/components/ProjectDashboard/components/PayProjectModal/PayProjectModal.tsx
filipviz: Thank you!

PR #4009, opened 2023-08-21
Edit cycle [4/n]: Tx success modal + fix memo

johnnyd-eth:
- Adds a modal after config tx successful:

https://github.com/jbx-protocol/juice-interface/assets/96150256/8c55842c-d3d3-4c4d-afaf-e6e7fb6e4f00

- Fixes memo functionality (moves `<ReviewConfirmModal />` into the edit cycle `<Form>`

- Updates memo styles according to latest designs:
Screen Shot 2023-08-22 at 1 08 27 am

### DIFF
--- a/src/styles/antd-overrides/input.scss
+++ b/src/styles/antd-overrides/input.scss
@@ -2,6 +2,7 @@
 .ant-input-affix-wrapper .ant-input,
 .ant-input-textarea textarea {
   @apply text-primary;
+  @apply py-1.5;
 }
 
 .ant-input-number,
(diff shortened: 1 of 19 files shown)


src/styles/antd-overrides/input.scss
```diff
@@ -2,6 +2,7 @@
 .ant-input-affix-wrapper .ant-input,
 .ant-input-textarea textarea {
   @apply text-primary;
+  @apply py-1.5;
```
wraeth-eth, line 5: does this affect any other antd text areas on the site?

johnnyd-eth: Yeah, they've all haven't got enough top padding

johnnyd-eth: @wraeth-eth It's only a 2px change, just keeping inline with Strath's latest.

Before:
Screen Shot 2023-08-22 at 12 31 33 pm

After:
Screen Shot 2023-08-22 at 12 30 36 pm

PR #4033, opened 2023-08-25
Edit cycle bugs [1/n]: Only save when changes have been made

johnnyd-eth:
https://github.com/jbx-protocol/juice-interface/assets/96150256/ac1ccabb-da5d-4bd8-848f-caedb9ca36ca


src/components/v2v3/V2V3Project/V2V3ProjectSettings/pages/NewEditCyclePage/EditCyclePage.tsx
```diff
@@ -43,6 +50,7 @@ export function EditCyclePage() {
           form={editCycleForm}
           layout="vertical"
           initialValues={initialFormData}
+          onValuesChange={handleFormValuesChange}
```
wraeth-eth, line 53: Is this strong enough to fix toggling? For instance, if a user has 100 reserved tokens, changes it to 0, the back to 100, will the button still be disabled?

johnnyd-eth: > Is this strong enough to fix toggling? For instance, if a user has 100 reserved tokens, changes it to 0, the back to 100, will the button still be disabled?

Nup you're right, is this case the button would stay enabled. I planned this initially, but it proved to be a right pain in the A to handle live diff checks in this `onValuesChange`. 

In any case, on second thought, I figured it made sense that if the user has made changes to the form and then back (to no changes), they might not realise everything is actually still the same, and therefore expect the button to work. In this case, we then show them clearly there is no diff in the Review modal, and the launch button from this modal definitely doesn't work if the form has no diff.

johnnyd-eth:
@wraeth-eth made a few more changes to this because I realised `onValuesChange` wasn't being called when the payouts or reserved list changes. So I just moved the `formHasUpdated` state to the context and call it manually from those spots

wraeth-eth (approved):

PR #4042, opened 2023-08-29
Edit cycle bugs [4/n]: Better error state for payouts and reserved splits overflow

johnnyd-eth:
### **- Shows better error states for payouts and reserved splits exceed 100%:**

Screen Shot 2023-08-29 at 10 32 06 pm

Screen Shot 2023-08-29 at 10 31 34 pm

### **- Prevents form from being submitted when the errors exist:**##

Screen Shot 2023-08-29 at 10 33 04 pm
Screen Shot 2023-08-29 at 10 33 36 pm
Screen Shot 2023-08-29 at 10 32 12 pm

### DIFF
--- a/src/components/v2v3/V2V3Project/V2V3ProjectSettings/pages/NewEditCyclePage/EditCyclePage.tsx
+++ b/src/components/v2v3/V2V3Project/V2V3ProjectSettings/pages/NewEditCyclePage/EditCyclePage.tsx
@@ -1,5 +1,5 @@
 import { Trans } from '@lingui/macro'
-import { Button, Form } from 'antd'
+import { Button, Form, Tooltip } from 'antd'
 import Loading from 'components/Loading'
 import { ExternalLinkWithIcon } from 'components/ProjectDashboard/components/ui/ExternalLinkWithIcon'
 import { ProjectMetadataContext } from 'contexts/shared/ProjectMetadataContext'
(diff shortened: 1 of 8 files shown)


wraeth-eth:
I think we want to show a permanent error message, rather than a tooltip for this. wdyt?

wraeth-eth:
Found a bug:

https://github.com/jbx-protocol/juice-interface/assets/104132113/1ea04738-ef66-4fc3-908a-3773870499fd

### Steps to repeat:

1. Insert a valid percentage payout
2. input 100.0000001 (probably doesnt need that much padding, 100.01 would likely work too)
3. Observe form locked (as expected)
4. Remove the trailing 1 and click out of form to allow auto formatting
5. Observe that form is still locked


In all honesty, I think these auto formatting inputs are more headache than they are worth. How would you feel about rolling back to a simple number input that allows silly looking things like 100.00000?

wraeth-eth (changes requested):
as above, looking great though @johnnyd ❤️

johnnyd-eth:
> I think we want to show a permanent error message, rather than a tooltip for this. wdyt?

Not sure what

> Found a bug:
> 
>  Screen.Recording.2023-08-30.at.7.39.51.am.mov 
> ### Steps to repeat:
> 1. Insert a valid percentage payout
> 2. input 100.0000001 (probably doesnt need that much padding, 100.01 would likely work too)
> 3. Observe form locked (as expected)
> 4. Remove the trailing 1 and click out of form to allow auto formatting
> 5. Observe that form is still locked
> 
> In all honesty, I think these auto formatting inputs are more headache than they are worth. How would you feel about rolling back to a simple number input that allows silly looking things like 100.00000?

Great catch here sir. I've thought of a way to approach this: only show the massive precision when necessary (when subTotal exceeds 100%). LMKWYT
- i.e. set the roundingPrecision for the whole table based on the decimal value exceeding the subTotal -> 100.001 produces RP of 3. Looks like this:
 
https://github.com/jbx-protocol/juice-interface/assets/96150256/eacfa3b6-759b-4589-99db-cc38d7798890

PR #4043, opened 2023-08-29
Edit cycle bugs [5/n]: Delete payouts bug

johnnyd-eth:
- Deletes bug where deleting a project will also delete the payout with the same address as the project's token beneficiary:

https://github.com/jbx-protocol/juice-interface/assets/96150256/b52a295a-9190-43f1-8724-54e300d87362

- Also sneaks in widening the number input:
BEFORE:
Screen Shot 2023-08-29 at 11 09 05 pm

AFTER:
Screen Shot 2023-08-29 at 11 05 41 pm


wraeth-eth (changes requested):
EDIT: nvm - ignore this. still reviewing

johnnyd-eth:
> EDIT: nvm - ignore this. still reviewing
> 
> EDIT EDIT: no this is still a problem.
> 
> You can input duplicate payouts

Agreed we probably shouldn't allow this from the front-end, but it won't cause any major problems for now since the contracts do accept duplicate payouts. Should we just get these ones in now and I fix in an upcoming?

johnnyd-eth:
> > EDIT: nvm - ignore this. still reviewing
> > EDIT EDIT: no this is still a problem.
> > You can input duplicate payouts
> 
> Agreed we probably shouldn't allow this from the front-end, but it won't cause any major problems for now since the contracts do accept duplicate payouts. Should we just get these ones in now and I fix in an upcoming?

(Agreed to merge in Discord)

PR #4047, opened 2023-08-30
Only show version appropriate ballots in listings

johnnyd-eth:
- We're currently showing all the sets of ballots for both V2 and V3 projects in any ballot listing (BEFORE):
Screen Shot 2023-08-30 at 6 19 22 pm

- This PR filters out the V2 ballots for the listings on V3 projects and vice-versa (AFTER):
Screen Shot 2023-08-30 at 6 14 07 pm

### DIFF
--- a/src/components/Create/hooks/useLoadInitialStateFromQuery.ts
+++ b/src/components/Create/hooks/useLoadInitialStateFromQuery.ts
@@ -73,7 +73,7 @@ const parseCreateFlowStateFromInitialState = (
   }
 
   const reconfigurationRuleSelection =
-    ballotStrategiesFn().find(s =>
+    ballotStrategiesFn({}).find(s =>
       isEqualAddress(s.address, initialState.fundingCycleData.ballot),
     )?.id ?? 'threeDay'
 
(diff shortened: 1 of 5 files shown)


src/components/Create/hooks/useLoadInitialStateFromQuery.ts
```diff
@@ -73,7 +73,7 @@ const parseCreateFlowStateFromInitialState = (
   }
 
   const reconfigurationRuleSelection =
-    ballotStrategiesFn().find(s =>
+    ballotStrategiesFn({}).find(s =>
```
wraeth-eth, line 76: This feels like a bug - is it possible to allow this to take in undefined instead?

wraeth-eth (changes requested):

src/components/Create/hooks/useLoadInitialStateFromQuery.ts
johnnyd-eth: Fixed this sir :pray:

tomquirk (approved):
@johnnyd-eth happy to approve and merge this. one comment but can be follow up if u wish

PR #4054, opened 2023-09-06
New edit cycle: apply issuance reduction rate to mint rate field

johnnyd-eth:
Fixes https://discord.com/channels/775859454780244028/1148469126839423057/1148469129536340049

### DIFF
--- a/src/components/v2v3/V2V3Project/V2V3ProjectSettings/pages/NewEditCyclePage/hooks/useLoadEditCycleData.tsx
+++ b/src/components/v2v3/V2V3Project/V2V3ProjectSettings/pages/NewEditCyclePage/hooks/useLoadEditCycleData.tsx
@@ -11,12 +11,14 @@ import {
 } from 'utils/format/formatTime'
 import { V2V3CurrencyName } from 'utils/v2v3/currency'
 import { distributionLimitStringtoNumber } from 'utils/v2v3/distributions'
+import { deriveNextIssuanceRate } from 'utils/v2v3/fundingCycle'
 import {
   formatDiscountRate,
   formatIssuanceRate,
   formatRedemptionRate,
   formatReservedRate,
 } from 'utils/v2v3/math'
+import { deserializeV2V3FundingCycleData } from 'utils/v2v3/serializers'
 import { useInitialEditingData } from '../../ReconfigureFundingCycleSettingsPage/hooks/useInitialEditingData'
 import { EditCycleFormFields } from '../EditCycleFormFields'
 
@@ -45,9 +47,16 @@ export const useLoadEditCycleData = () => {
       ).toNumber()
       const fundingCycleData = currentProjectData.fundingCycleData
       const fundingCycleMetadata = currentProjectData.fundingCycleMetadata
-      const mintRate = fundingCycleData?.weight
-        ? parseFloat(formatIssuanceRate(fundingCycleData.weight))
-        : parseFloat(DefaultTokenSettings.initialMintRate)
+      const mintRate = parseFloat(
+        formatIssuanceRate(
+          deriveNextIssuanceRate({
+            weight: BigNumber.from(0),
+            previousFC: deserializeV2V3FundingCycleData(
+              currentProjectData.fundingCycleData,
+            ),
+          }).toString(),
+        ),
+      )
       const reservedTokens = fundingCycleMetadata.reservedRate
         ? parseFloat(formatReservedRate(fundingCycleMetadata.reservedRate))
         : DefaultTokenSettings.reservedTokensPercentage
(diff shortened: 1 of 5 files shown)


src/components/v2v3/V2V3Project/V2V3ProjectSettings/pages/NewEditCyclePage/hooks/useLoadEditCycleData.tsx
```diff
@@ -45,9 +47,16 @@ export const useLoadEditCycleData = () => {
       ).toNumber()
       const fundingCycleData = currentProjectData.fundingCycleData
       const fundingCycleMetadata = currentProjectData.fundingCycleMetadata
-      const mintRate = fundingCycleData?.weight
-        ? parseFloat(formatIssuanceRate(fundingCycleData.weight))
-        : parseFloat(DefaultTokenSettings.initialMintRate)
+      const mintRate = parseFloat(
```
johnnyd-eth, line 50: Loads next issuance rate into edit cycle form (after issuance reduction)

src/utils/v2v3/fundingCycle.ts
```diff
@@ -138,19 +139,21 @@ export const deriveNextIssuanceRate = ({
   weight,
   previousFC,
 }: {
-  weight: BigNumber
-  previousFC: V2V3FundingCycle | undefined
+  weight?: BigNumber
```
johnnyd-eth, line 142: Changes this function a bit so this parameter is optional

src/components/v2v3/V2V3Project/V2V3ProjectSettings/pages/NewEditCyclePage/hooks/usePrepareSaveEditCycleData.tsx
```diff
@@ -87,9 +89,9 @@ export const usePrepareSaveEditCycleData = () => {
 
   const editingFundingCycleData: V2V3FundingCycleData = {
     duration: BigNumber.from(durationSeconds),
-    weight: BigNumber.from(
-      issuanceRateFrom(formValues.mintRate?.toString() ?? '0'),
-    ),
+    weight: mintRateHasDiff
+      ? newMintRate
+      : reduxConfig.editingFundingCycleData.weight,
```
tomquirk, line 94: if no change, shouldn't this be `WEIGHT_UNCHANGED`?

src/utils/v2v3/fundingCycle.ts
```diff
@@ -138,19 +139,21 @@ export const deriveNextIssuanceRate = ({
   weight,
   previousFC,
```
tomquirk, line 140: non blocking but wonder if its just easier to pass in exactly what we need, rather than the whole fc object 🤔

src/components/v2v3/V2V3Project/V2V3ProjectSettings/pages/NewEditCyclePage/hooks/usePrepareSaveEditCycleData.tsx
tomquirk, line 94: if weight doesnt change, shouldnt this be `WEIGHT_UNCHANGED`?

tomquirk (changes requested):
@johnnyd-eth one comment worth noting https://github.com/jbx-protocol/juice-interface/pull/4054#discussion_r1319926747

src/components/v2v3/V2V3Project/V2V3ProjectSettings/pages/NewEditCyclePage/hooks/usePrepareSaveEditCycleData.tsx
johnnyd-eth: Good catch, but no - this is eventually passed into this function inside `useReconfigureFundingCycle`. As it works now, passing `WEIGHT_UNCHANGED` (`0`) as `newFundingCycleWeight` would be interpreted as the user wanting a weight `0`, in which case we pass `1` to the contracts. 

Screen Shot 2023-09-09 at 3 01 57 pm

src/utils/v2v3/fundingCycle.ts
johnnyd-eth: Good call, should definitely have just made this function take `oldWeight` and `oldDiscountRate` as opposed to the whole FC object. However, changing it will create quite a diff - perhaps not appropriate for this PR. I will get to it in a follow up

Screen Shot 2023-09-09 at 3 06 29 pm

johnnyd-eth:
(Discussed in Discord, all good to merge)

PR #4067, opened 2023-09-14
Dev blocked projects and block Metakeys copy-cat

johnnyd-eth:
For projects listed in `BLOCKED_PROJECT_IDS`:
- Hides from project listings 
- Shows banner on project page
- Disables pay button

Screen Shot 2023-09-14 at 2 31 27 pm

### DIFF
--- a/src/constants/blocklist.ts
+++ b/src/constants/blocklist.ts
@@ -1,4 +1,16 @@
+import { NetworkName } from 'models/networkName'
+import { readNetwork } from './networks'
+
+const PROJECT_IDS = {
+  METAKEYS_COPYCAT: 564, // copycat of 563
+}
+
 // List of delisted projects
-export const BLOCKED_PROJECT_IDS = [
-  '564', //copycat of 563
-]
+const V2_BLOCKLISTED_PROJECT_IDS_BY_NETWORK: Partial<
+  Record<NetworkName, number[]>
+> = {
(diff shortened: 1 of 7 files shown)


src/hooks/useProjects.ts
```diff
@@ -68,7 +69,13 @@ export function useDBProjectsInfiniteQuery(
             pageSize,
           })}`,
         )
-        .then(res => (res.data ? res.data.map(parseDBProject) : []))
+        .then(res =>
```
johnnyd-eth, line 72: Adds filters to "all" (same as the "new" query) and "trending" projects queries

src/constants/blocklist.ts
```diff
@@ -0,0 +1,4 @@
+// List of delisted projects
+export const BLOCKED_PROJECT_IDS = [
+  '564', //copycat of 563
```
peripheralist, line 3: we'll want these to be project `id`s, not `projectId`s (otherwise this would block a v1 project 564). we also want it to be specific to mainnet

peripheralist: or, we could copy the pattern used for archived projects (storing `V2ArchivedProjectIdsByNetwork` that are all pv == 2), but we need to add the `pv == 2` statement in the filter

peripheralist: image
image

peripheralist (approved):

PR #4074, opened 2023-09-18
New payouts table in create flow behind feature flag

johnnyd-eth:
Apologies diff has gone a bit cooked because I changed the file names. It's not that big. Worth just checking out the new `PayoutsPage` locally. `OldPayoutsPage` is completely unchanged from what `PayoutsPage` previously.

https://github.com/jbx-protocol/juice-interface/assets/96150256/b7c3a757-d959-4486-ab2f-779058dd9efe

Think it's worth letting everyone have a play with this behind FF to make sure nothing's broken!

Still need to implement something better in the "Review and deploy" section of create:
- Currently show before fee amounts for all payouts which is misleading


src/components/Create/components/pages/PayoutsPage/OldPayoutsPage.tsx
```diff
@@ -0,0 +1,293 @@
+import { StopOutlined } from '@ant-design/icons'
```
johnnyd-eth, line 1: Absolutely nothing changed in this file from when it was the OG `PayoutsPage.tsx`

tomquirk: Heads up: pushing a commit to reverse the naming of these. When we clean up feature flag, we can rename `NewPayoutsPage` to `PayoutsPage`

PR #4081, opened 2023-09-22
[DO NOT MERGE] Use subgraph for historical funding cycles query

peripheralist:
**Do not merge, must be shipped alongside subgraph version migration**

- Creates a new `fundingCycles` graphql query used to page funding cycle history for V2V3 projects
- Adapts `ConfigureEvent` element for compatibility with new subgraph type
- Adds utils for converting between subgraph fundingCycle type and the `V2V3FundingCycle`/`V2V3FundingCycleMetadata` types already native to the app


johnnyd-eth (approved):
So clean! Can we not scrap the `useHistorySubPanel` hook with this as well?

peripheralist:
@johnnyd-eth yessir good call

PR #4085, opened 2023-09-26
[3/3] Add more obvious currency selector to payouts table

johnnyd-eth:
https://github.com/jbx-protocol/juice-interface/assets/96150256/dcad95d6-83f1-4a80-a9d0-01da0ca7fd2f

### DIFF
--- a/src/utils/format/formatCurrency.ts
+++ b/src/utils/format/formatCurrency.ts
@@ -1,6 +1,7 @@
 import { BigNumber, BigNumberish } from 'ethers'
 
 import { CurrencyName } from 'constants/currency'
+import round from 'lodash/round'
 import { parseWad } from './formatNumber'
 
 export class CurrencyUtils {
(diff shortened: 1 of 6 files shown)


src/utils/format/formatCurrency.ts
```diff
@@ -59,7 +60,7 @@ export class CurrencyUtils {
       return parseWad(this.weiToUsd(amount)?.toString())
     if (targetCurrency === 'ETH' && this.usdPerEth !== undefined)
       return BigNumber.from(amount)
-        .div(this.usdPerEth * 100)
+        .div(round(this.usdPerEth * 100))
```
johnnyd-eth, line 63: Turns out I don't think we've ever actually converted USD->ETH in the app. There was a bug here that `this.usdPerEth * 100` could be a decimal, which breaks the site

tomquirk: classic. nice catch

tomquirk (approved):

PR #4095, opened 2023-10-13
Enable buyback delegate metadata for pay() txs to juicebox

peripheralist:
**Changes**
- New `encodeDelegateMetadata()` pattern as a new catch-all metadata encoding util for all delegate types
  - Encodes metadata for multiple delegates, using the new [metadata-lib pattern](https://github.com/jbx-protocol/juice-delegate-metadata-lib/tree/main), and the [juicebox-metadata-helper](https://www.npmjs.com/package/juicebox-metadata-helper) package
  - Also supports legacy encoding of jb721Delegate metadata where version < 3.4 (not compatible with buyback delegate)
  - Will be extensible for any future cases where any project makes use of multiple metadata-lib-compatible delegates
- Enables buyback delegate metadata encoding for Juicebox
  - Minted tokens count must not exceed half of available AMM liquidity (arbitrary) to avoid slippage. If condition is not met, buyback delegate metadata will be omitted
  - If included, buyback delegate metadata will instruct protocol to use ALL ETH in pay amount to purchase tokens on AMM. In order to reduce arbitrage opportunities, the client will estimate the output of the swap based on the market price, and encode a minimum of 95% of that estimate in the delegate metadata. This is arbitrary—we only care about getting a minimum of the number of tokens that would otherwise be minted by the protocol, but we don't want to _request_ a minimum that is significantly smaller than the actual swap output, as it would increase the likelihood of the tx being front run.
  - Support for buyback delegate with Juicebox project is hard coded to projectId. We could determine this dynamically, but since there's no expectation of any other project using the buyback delegate this is prolly fine for now
- Adds a notice in Juicebox Pay > TokensReceived section that indicate a payer may receive more tokens than what are estimated. 
image

### DIFF
--- a/src/components/AMMPrices/hooks/useERC20UniswapPrice.ts
+++ b/src/components/AMMPrices/hooks/useERC20UniswapPrice.ts
@@ -35,8 +35,8 @@ interface State {
 }
 
 type Props = {
-  tokenSymbol: string
-  tokenAddress: string
+  tokenSymbol: string | undefined
+  tokenAddress: string | undefined
 }
 
 /**
@@ -126,6 +126,8 @@ export function useUniswapPriceQuery({ tokenSymbol, tokenAddress }: Props) {
   return useQuery(
     [`${tokenSymbol}-uniswap-price`],
     async () => {
+      if (!tokenAddress || !tokenSymbol) return null
+
       try {
         const poolAddress = await getPoolAddress()
         if (!poolAddress) {
@@ -179,6 +181,7 @@ export function useUniswapPriceQuery({ tokenSymbol, tokenAddress }: Props) {
     },
     {
       refetchInterval: 30000, // refetch every 30 seconds
+      enabled: !!(tokenAddress && tokenSymbol),
     },
   )
 }
--- a/src/components/AMMPrices/hooks/useERC20UniswapPrice.ts
+++ b/src/components/AMMPrices/hooks/useERC20UniswapPrice.ts
@@ -2,6 +2,7 @@ import { Token } from '@uniswap/sdk-core'
 import IUniswapV3FactoryABI from '@uniswap/v3-core/artifacts/contracts/interfaces/IUniswapV3Factory.sol/IUniswapV3Factory.json'
 import IUniswapV3PoolABI from '@uniswap/v3-core/artifacts/contracts/interfaces/IUniswapV3Pool.sol/IUniswapV3Pool.json'
 import {
+  FeeAmount,
   Pool,
   FACTORY_ADDRESS as UNISWAP_V3_FACTORY_ADDRESS,
 } from '@uniswap/v3-sdk'
@@ -43,7 +44,12 @@ type Props = {
  * Pools are created at a specific fee tier.
  * https://docs.uniswap.org/protocol/concepts/V3-overview/fees#pool-fees-tiers
  */
-const UNISWAP_FEES_BPS = [10000, 3000, 500]
+const UNISWAP_FEES_BPS = [
+  FeeAmount.LOWEST,
+  FeeAmount.LOW,
+  FeeAmount.MEDIUM,
+  FeeAmount.HIGH,
+]
 const networkId = readNetwork.chainId
 
 /**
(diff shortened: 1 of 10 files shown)


src/components/AMMPrices/hooks/useERC20UniswapPrice.ts
```diff
@@ -126,6 +132,8 @@ export function useUniswapPriceQuery({ tokenSymbol, tokenAddress }: Props) {
   return useQuery(
```
tomquirk, line 132: wow, its awesome that this code was useful! haha

src/components/v2v3/V2V3Project/ManageNftsSection/RedeemNftsModal/RedeemNftsModal.tsx
```diff
@@ -21,7 +21,7 @@ import {
   encodeJB721DelegateV3_2RedeemMetadata,
   encodeJB721DelegateV3_4RedeemMetadata,
   encodeJB721DelegateV3RedeemMetadata,
-} from 'utils/encodeJb721DelegateMetadata/encodeJb721DelegateMetadata'
+} from 'utils/delegateMetadata/encodeJb721DelegateMetadata'
```
tomquirk, line 24: probably gonna move this to the sdk fwiw

src/components/v2v3/V2V3Project/ProjectDashboard/components/PayProjectModal/components/ReceiveTokensItem.tsx
```diff
@@ -31,7 +39,18 @@ export const ReceiveTokensItem = ({ className }: { className?: string }) => {
             <Trans>{badgeTitle} Token</Trans>
           </CartItemBadge>
         </div>
-        <div>{receivedTickets}</div>
+        {buybackDelegateEnabled ? (
+          <div>
+            ≥ {receivedTickets}
+            <TooltipIcon
+              tip={
+                'This project may purchase tokens from a secondary market instead of minting new tokens, which may result in more tokens distributed depending on the swap price.'
```
tomquirk, line 47:
```suggestion
                'Your payment may purchase tokens from a secondary market instead of minting new tokens. You might receive more tokens depending on the swap price.'
```

src/components/v2v3/V2V3Project/ProjectDashboard/hooks/usePayProjectModal/usePayProjectTx.ts
```diff
@@ -97,15 +110,54 @@ export const usePayProjectTx = ({
...
+        BigNumber.from(priceQuery?.projectTokenPrice.numerator.toString()).div(
+          BigNumber.from(priceQuery?.projectTokenPrice.denominator.toString()),
+        ),
       )
 
+      const shouldUseBuybackDelegate =
+        projectId &&
+        priceQuery &&
+        requiredTokens &&
+        BUYBACK_DELEGATE_ENABLED_PROJECT_IDS.includes(projectId) &&
```
tomquirk, line 130: (non-blocking) instead of this, I think we could use the Delegates Registry to check if their fc datasource is a buy-back delegate. But this works for now 🙏

src/utils/delegateMetadata/encodeDelegateMetadata.ts
```diff
@@ -0,0 +1,100 @@
...
+} from 'utils/delegateMetadata/encodeJb721DelegateMetadata'
+
+/**
+ * Encode pay metadata for project delegate contracts.
+ *
+ * Supports delegate contracts:
+ *  - jb721Delegate
+ *  - jbBuybackDelegate
+ *
+ * Note: jbBuybackDelegate metadata uses the Metadata Lib pattern (https://github.com/jbx-protocol/juice-delegate-metadata-lib) and CANNOT be encoded with jb721Delegate contracts older than version 3.4.
```
tomquirk, line 18: faaaantastic. this looks great. will also move most of this to sdk methinks

tomquirk:
@peripheralist heads up: pushing a rebase 🙏

tomquirk (approved):
tested and working!!!! lfg

PR #4109, opened 2023-10-28
fix: various image handling improvements

filipviz:
1. Small fix to allow apng header images.
2. Add an image URI override for the Breadfruit logo (it's hefty). Also separate image URI overrides for mainnet and goerli.
3. Improve the middleware function. Instead of downloading the entire image and streaming it to the client, just check the headers and first few bytes, and return a redirect if they look good.
4. Add a check to NFT cards for the old IPFS endpoint in which was causing some NFT artwork to fail to load.
5. Other small changes.


filipviz:
@aeolianeth put a bunch of small changes in here. The image middleware should be much more robust now – instead of downloading the entire image and streaming it to the user (which can bump into Vercel's 4.5 MB size limit), it now uses a Range header to just check the headers and file signature (3-4 bytes). We can add new filetypes to this relatively trivially.

SUGGESTION: we should consider routing *all* requests for user-defined images through this endpoint. NFT artwork and banners are susceptible to the same risks as project logos, and we should at least check the signatures + headers. Let me know what you think – I can create another PR.

The endpoint is relatively simple. `/api/image/[uri]`. Just pass a URL encoded with `encodeURIComponent`.

src/pages/api/image/[url].ts
```diff
@@ -11,18 +11,40 @@ export default async function handler(
 
   const imageUrl = req.query?.url as string
   if (!imageUrl) {
```
Comment (tomquirk) [inline src/pages/api/image/[url].ts:13]: @filipviz this LGTM. one nice improvement we could make would be to cache responses for this endpoint. Given we know the URI will always point to the same file (otherwise, it would have a different CID), the cache duration could be pretty long.

tomquirk (approved):
@filipviz nice iteration, lgtm!

PR #4110, opened 2023-10-29
Filter out blocklisted projects at api level

peripheralist:
We previously implemented a filter on the client side to avoid showing blocklisted projects in the UI. This created a convoluted bug where:
1. Items were filtered out of project DB responses, _after_ the response was received
2. The client may receive a smaller-than-expected pageSize of results
3. The client would assume there were no more objects to query, breaking infinite paging

This filter has been moved to the API level, so that blocklisted projects will be filtered in the query to the projects DB.

### DIFF
--- a/src/hooks/useProjects.ts
+++ b/src/hooks/useProjects.ts
@@ -1,6 +1,4 @@
 import axios from 'axios'
-import { V2_BLOCKLISTED_PROJECT_IDS } from 'constants/blocklist'
-import { PV_V2 } from 'constants/pv'
 import { BigNumber } from 'ethers'
 import { DBProject, DBProjectQueryOpts, DBProjectRow } from 'models/dbProject'
 import { Json } from 'models/json'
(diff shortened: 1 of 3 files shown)


src/hooks/useProjects.ts
```diff
@@ -16,11 +14,6 @@ import { parseDBProject, parseDBProjectJson } from 'utils/sgDbProjects'
 
 const DEFAULT_STALE_TIME = 60 * 1000 // 60 seconds
 
-// Filter applied to all DB project query results
-// TODO this could be applied at the /api/projects level, but would take some more finessing. we're working quick here
```
peripheralist, line 20: Frens, always do the finessing. Especially when it's actually not a lot of finessing

_Narrator: This was not a lot of finessing_

tomquirk: Prime finesse right here.

tomquirk (approved):
sweeeeet, thanks @peripheralist

PR #4115, opened 2023-10-31
feat: add start date to project dashboard cycle tab

filipviz:
- Add cycle start date to dropdown Rules tab on project dashboard in this format: `1970-01-01, Wednesday, 07:00:00 PM EST`
- Change word "Configuration" to "Rules" in dropdown (should we go with "Details" instead?)
- Tests.

NOTE: All tests on CI fail because of the same error: `Cannot find module './src/locales/en/messages.js' from '../jest.setup.ts'`. Works locally, not sure how to debug this.

image
image


tomquirk:
@filipviz this LGTM but the test is failing, i think the tz/locale mock didnt work. I tried a quick fix but no luck - mind taking a look?

filipviz:
> @filipviz this LGTM but the test is failing, i think the tz/locale mock didnt work. I tried a quick fix but no luck - mind taking a look?

Should be good @aeolianeth. If it keeps breaking we can remove this test. It's finnicky to test because of timezones, but the component itself is quite simple.

tomquirk (approved):
lfg

PR #4120, opened 2023-11-01
fix: display upcoming cycles with no duration

filipviz:
Address issue reported by filipv [here](https://discord.com/channels/775859454780244028/866040669712678942/1167579298039869500) and jango [here](https://discord.com/channels/775859454780244028/866040669712678942/1169305100083286067).

Currently, if the upcoming cycle has no duration, the upcoming cycle tab says there's no upcoming cycle. This is inaccurate, and this should check if the *current* cycle has no duration. This PR fixes that.

image
image


src/components/v2v3/V2V3Project/ProjectDashboard/components/CyclesPayoutsPanel/hooks/useCurrentUpcomingSubPanel.ts
```diff
@@ -35,6 +35,12 @@ export const useCurrentUpcomingSubPanel = (type: 'current' | 'upcoming') => {
     return upcomingFundingCycle?.duration?.isZero() ?? true
   }, [fundingCycle?.duration, type, upcomingFundingCycle?.duration])
 
+  /** Determines if the CURRENT cycle is unlocked.
+   * This is used to check if the upcoming cycle can start at any time. */
+  const currentCycleUnlocked = useMemo(() => {
+    return fundingCycle?.duration?.isZero() ?? true
```
tomquirk, line 41: **suggestion**: Is this equivalent?

```suggestion
    return Boolean(fundingCycle?.duration?.isZero())
```

filipviz: @aeolianeth no. If `fundingCycle?.duration?.isZero()` is null or undefined, the null coalescing operator returns `true`, where casting as a Boolean would return `false` (it is falsy).

Although I didn't look TOO deeply into this, I assume that if the funding cycle's duration were undefined, it means that there isn't a current cycle (maybe launched without one, for example). I didn't double check this, but it matches a pattern used in the other cycleUnlocked check.

tomquirk: true fact, all g

tomquirk (approved):

PR #4130, opened 2023-11-06
Token distribution chart

peripheralist:
Adds visualization of token distribution to Participants modal. User can toggle between a pie chart and an area chart by clicking icons below. Also adds the total number of wallets.

**Pie chart**
Shows the relative shares held by the top 100 (arbitrary) wallets for a project, with the remaining wallets lumped into a single slice. Hovering each slice shows the wallet name, % share, and token balance, while hovering the remainder slice shows the number of hidden wallets and their combined total share.
image
image

**Area chart**
Visualizes the spread of token distribution across _all_ wallets using an area chart format. All wallets, ordered by token balance, are clustered in up to 12 groups. Each group is plotted on the X axis, with the group's combined token share on the Y axis.
image
image

Video:
https://github.com/jbx-protocol/juice-interface/assets/79433522/d7044fc7-b7e1-4bf0-a466-7e3a01cb2ed0

Also refactors the ParticipantsModal to use a more current implementation of Apollo graph query, and organize components. Number of all holder wallets is now shown at the top of the modal.

Update 01: Gap in pie chart has been removed
image


src/components/modals/ParticipantsModal/HoldersList.tsx
```diff
@@ -0,0 +1,198 @@
...
+  },
+  {
+    label: t`Last paid`,
+    value: Participant_OrderBy.lastPaidTimestamp,
+  },
+]
+
+const pageSize = 100
+
+export default function HoldersList({
```
peripheralist, line 48: copied over from original ParticipantsModal component. participants query now uses Apollo query pattern

src/components/TokenDistributionChart/TokenPieChart.tsx
```diff
@@ -0,0 +1,156 @@
...
+
+    // Only show (arbitrary) max number of wallets to avoid chart clutter
+    const maxVisibleWallets = 100
+
+    const visibleWallets: typeof participants = []
+    const remainderWallets: typeof participants = []
+
+    participants.forEach((p, i) =>
+      (i < maxVisibleWallets ? visibleWallets : remainderWallets).push(p),
+    )
```
tomquirk, line 42: **suggestion** is something like this an equivalent?

```suggestion
    const visibleWallets = participants.slice(0, maxVisibleWallets)
    const remainderWallets = participants.slice(maxVisibleWallets, participants.length)
```

src/components/TokenDistributionChart/TokenPieChart.tsx
```diff
@@ -0,0 +1,156 @@
...
+
+    const _chartData = visibleWallets.reduce(
```
tomquirk, line 44: **suggestion** `map` would work here but nbd

tomquirk (approved):
@peripheralist lgtm!!!! i left just one minor comment

peripheralist:
thanks @aeolianeth good calls, implemented both. code got leftover from some previous logic and not properly cleaned up lol. merging

PR #4150, opened 2023-11-21
:wrench: Additional settings: OFAC check

0xPoint:
This check blocks the ability to buy nft , for wallets that are on the sanction list


Pasha8914:
### if OFAC check successful
Screenshot 2023-11-20 at 20 29 25

Screenshot 2023-11-20 at 20 29 35




### if OFAC check declined
Screenshot 2023-11-20 at 20 31 06

Screenshot 2023-11-20 at 20 31 24

### OFAC setting setup
Screenshot 2023-11-20 at 20 33 09

### OFAC setting edit
Screenshot 2023-11-20 at 20 33 20

src/contexts/v2v3/V2V3ProjectOFACProvider.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+  const checkIsAddressListedInOFAC = useCallback(
+    async (address: string) => {
+      try {
+        const { data } = await axios.get<{ isGoodAddress: boolean }>(
+          `${OFAC_API}?address=${address}`,
+        )
+        setIsListed(!data.isGoodAddress)
+      } catch (err) {
+        setIsListed(true)
```
wraeth-eth, line 34: IN the case of an error, I think it would be safer to set to false

src/contexts/v2v3/V2V3ProjectOFACProvider.tsx
```diff
@@ -0,0 +1,53 @@
...
+      }
+    },
+    [setIsListed],
```
wraeth-eth, line 37: You shouldn't need to have `setIsListed` in the dependencies as it is a state setter. https://legacy.reactjs.org/docs/hooks-reference.html#usestate

This is from the old docs, but the behaviour is the same

src/contexts/v2v3/V2V3ProjectOFACProvider.tsx
```diff
@@ -0,0 +1,53 @@
...
+  )
+
+  useEffect(() => {
+    if (shouldCheckOfac && userAddress) {
+      checkIsAddressListedInOFAC(userAddress)
+    }
+    // unexpected dependency checkIsAddressListedInOFAC
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
wraeth-eth, line 45: Making the above change (removing `setIsListed`) will allow you to add `checkIsAddressListedInOFAC` back into this dependency tree with no bugs

.nvmrc
```diff
@@ -1 +1 @@
-v14.17.0
+v18.0.0
```
wraeth-eth, line 1: Is this change needed?

src/components/Create/components/pages/ReviewDeploy/components/ProjectDetailsReview/ProjectDetailsReview.tsx
```diff
@@ -83,6 +84,17 @@ export const ProjectDetailsReview = () => {
           ) : null
         }
       />
+      {projectRequiredOFACCheck && (
+        <ReviewDescription
+          className="col-span-3"
+          title={t`OFAC Sanctions screening`}
+          desc={
+            <div className="overflow-hidden text-ellipsis text-base font-medium">
+              {String(projectRequiredOFACCheck)}
```
wraeth-eth, line 93: If undefined, this will show as `"undefined"`. Better to use explicit change, or set a default value somewhere

src/components/v2v3/V2V3Project/ProjectDashboard/hooks/index.ts
```diff
@@ -0,0 +1,10 @@
+export * from './useAboutPanel'
```
wraeth-eth, line 1: Is it possible to remove this barrel file? We are trying to clean up all usage of barrel files where we can.

I think @aeolianeth recently nuked all barrels from the project

https://dev.to/tassiofront/barrel-files-and-why-you-should-stop-using-them-now-bc4

wraeth-eth (changes requested):
Thanks for the PR! Made some comments about some of the code.

Unfortunately I am out of loop of the OFAC decision at JB but left some markers about some changes I'd recommend

src/components/v2v3/V2V3Project/ProjectDashboard/hooks/index.ts
tomquirk: All good - lets leave it for now! Thanks for the callout though @wraeth-eth 🙏

.nvmrc
Pasha8914: > Is this change needed?

If this file will not have the current version (as in package.json) then why do I need this file?

0xPoint: > Is this change needed?

If this file will not have the current version (as in [package.json](https://github.com/jbx-protocol/juice-interface/blob/main/package.json#L59)) then why do I need this file?

src/contexts/v2v3/V2V3ProjectOFACProvider.tsx
0xPoint: > IN the case of an error, I think it would be safer to set to false

a sanctioned user may block a request in the browser, and it is critical for us to prevent such a user from doing so

src/components/Create/components/pages/ReviewDeploy/components/ProjectDetailsReview/ProjectDetailsReview.tsx
0xPoint: if there is `undefiend`, the component will not showing (look at 87 row)

I thought I'd check it out
image

there is nothing to display

image

If the variable is a bool, it looks like this
image

0xPoint: If you have internal rules/code style for such cases, please describe it to me, I will fix it according to your rules/code style.

tomquirk:
@0xPoint thanks for making those changes! Again, really appreciate you doing the heavy-lifting on this PR.

I'm going to merge into upstream `feat/ofac`, check some things and then merge to main. I'll let you know when it's live 🚀

PR #4151, opened 2023-11-22
Update settings to handle 1 cycle multi reconfigs

johnnyd-eth:
Couldn't reliably reproduce the original problem from Jango (https://discord.com/channels/775859454780244028/1169306840836882532/1169306843395408054). Took some digging but I believe Jango's issue was a lingering AntD state to show. Resetting on save seems to fix 

Fixes JB-809 (https://linear.app/peel/issue/JB-809/small-but-important-payout-bug)


tomquirk (approved):

tomquirk:
nice catch, lgtm probs

PR #4163, opened 2023-11-30
feat: make whole redeem nft card able to be selected

johnnyd-eth:
- ATM, only the bottom of the card can be used to select/deselect the card
- Removes unnecessary complexity from the component in question, whose `preview` is currently not being used anywhere


tomquirk (approved):
@johnnyd-eth lgtm.

it would be nice to nuke NftTierCard in favor of whatever we use on the new project page 👍

PR #4167, opened 2023-12-06
:bug: Block payment for sanctioned address

0xPoint:
If users are not connected, they can add an NFT or donation to the cart. Once connected, the app does not check their OFAC status

DEFAULT
image

IF BUTTON DISABLED AND BUTTON HOVERED
image


0xPoint:
Fixed white space

BEFORE
image

AFTER
image

tomquirk (approved):

tomquirk:
@0xPoint this LGTM, thanks for catching and for the quick fix. Tests are failing because the PR is from a fork - will need to look into that.

Merging!

PR #4196, opened 2024-01-15
feat: process held fees page

filipviz:
Add a settings page which allows projects to process held fees, if they have any.

If a project has no held fees, the page looks like this:

image

If a project does have held fees, the page looks like this:

image

In settings menu:

image

Dark mode:

image


filipviz:
NOTE: This PR also fixes the read/process held fees transactors. Instead of always using `JBETHPaymentTerminal`, it uses the project's primary ETH terminal.

src/components/v2v3/V2V3Project/V2V3ProjectSettings/pages/ProcessHeldFeesPage/ProcessHeldFeesPage.tsx
```diff
@@ -0,0 +1,117 @@
...
+          </Trans>
+        </p>
+      </MinimalCollapse>
+      <TransactorButton
+        onClick={processHeldFees}
+        loading={processingHeldFees}
+        type="primary"
+        connectWalletText={<Trans>Not able to process fees</Trans>}
+        text={<Trans>Process held fees</Trans>}
+        disabled={!canProcessHeldFees || !heldFees}
```
tomquirk, line 106: **suggestion**: i think we can just hide the button if there's no held fees. you've already got the empty state above, which is great and sufficient I think!

src/components/v2v3/V2V3Project/V2V3ProjectSettings/pages/ProcessHeldFeesPage/ProcessHeldFeesPage.tsx
```diff
@@ -0,0 +1,117 @@
...
+      />
+      {!canProcessHeldFees ? (
+        <p>
+          <Trans>Your wallet isn't allowed to process held fees.</Trans>
```
tomquirk, line 110: **praise** great!

src/hooks/v2v3/contractReader/useHeldFeesOf.ts
```diff
@@ -1,25 +1,24 @@
...
 import { sumHeldFees } from 'utils/v2v3/math'
+import { useProjectPrimaryEthTerminal } from '../V2V3ProjectContracts/projectContractLoaders/useProjectPrimaryEthTerminal'
 
 export function useHeldFeesOf() {
   const [heldFees, setHeldFees] = useState<number>()
 
   const { projectId } = useContext(ProjectMetadataContext)
-  const { contracts } = useContext(V2V3ContractsContext)
+  if (!projectId) return
+  const { JBETHPaymentTerminal } = useProjectPrimaryEthTerminal({ projectId })
```
tomquirk, line 11: **suggestion** We should just refactor this hook to use `useContractReader` like normal.

tomquirk (changes requested):
@filipviz thanks for this!

So, i totally forgot that we already have the 'process held fees' section in the Project Tools drawer 🙈 .

I think what you've done here is better though.

We should remove `src/components/Project/ProjectToolsDrawer/HeldFeesSection.tsx` and its uses - WDYT?

src/components/v2v3/V2V3Project/V2V3ProjectSettings/pages/ProcessHeldFeesPage/ProcessHeldFeesPage.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+  return (
+    <>
+      {heldFees ? (
+        <>
+          <p>
+            <Trans>Process this project's held fees.</Trans>
+          </p>
+          <Statistic
+            title={<Trans>Current held fees</Trans>}
```
tomquirk, line 53:
```suggestion
            title={<Trans>Fees held</Trans>}
```

filipviz:
> We should remove `src/components/Project/ProjectToolsDrawer/HeldFeesSection.tsx` and its uses - WDYT?

Same page. It's actually broken right now. Removing.

src/components/v2v3/V2V3Project/V2V3ProjectSettings/pages/ProcessHeldFeesPage/ProcessHeldFeesPage.tsx
filipviz: SGTM! adding

src/components/v2v3/V2V3Project/V2V3ProjectSettings/pages/ProcessHeldFeesPage/ProcessHeldFeesPage.tsx
filipviz: thank you :^)

src/hooks/v2v3/contractReader/useHeldFeesOf.ts
tomquirk: dw bout it for now tho @filipviz.

But quick question: whats the reason behind the changes here?

filipviz:
image
Without held fees now

src/hooks/v2v3/contractReader/useHeldFeesOf.ts
tomquirk: Oh. @filipviz i see! It wasn't using the projects terminal.

In this case, we should already have the primary terminal in context.

Try getting `JBETHPaymentTerminal` from `contracts` here:
```
  const { contracts } = useContext(V2V3ProjectContractsContext)
```

Also, will need to remove line 10: `if (!projectId) return`. React doesn't like hooks running conditionally 🙏

filipviz: > But quick question: whats the reason behind the changes here?

The old version just got the `JBETHPaymentTerminal`, so if a project was using 3_1, 3_1_1, etc. it wouldn't work

filipviz: Ok, will leave for now! If we're leaving, this should be good now.

filipviz: > Oh. @filipviz i see! It wasn't using the projects terminal.
> 
> In this case, we should already have the primary terminal in context.
> 
> Try getting `JBETHPaymentTerminal` from `contracts` here:
> 
> ```
>   const { contracts } = useContext(V2V3ProjectContractsContext)
> ```
> 
> Also, will need to remove line 10: `if (!projectId) return`. React doesn't like hooks running conditionally 🙏

Oh didn't see this. Ok will do – sorry for some of this. Not too familiar with the contract abstractions on JBM

filipviz:
@aeolianeth anything else needed?

tomquirk (approved):
@filipviz lets gooooo

PR #4237, opened 2024-01-28
fix: toString bug

tomquirk:


src/components/v2v3/V2V3Project/ProjectDashboard/components/Cart/hooks/useRewardEligibilityCallout.ts
```diff
@@ -20,7 +20,7 @@ export const useRewardEligibilityCallout = () => {
 
   const weiPayAmount = useWeiConverter({
     currency: payAmount?.currency,
-    amount: payAmount?.amount.toString(),
+    amount: payAmount?.amount?.toString(),
```
tomquirk, line 23: idk if this is even an issue. just came across it, and thought it might be given the typing issue below

wraeth-eth (approved):

PR #4297, opened 2024-04-06
fix: formatCurrencyAmount for ETH

aeolianeth:


src/utils/format/formatCurrencyAmount.ts
```diff
@@ -27,7 +31,13 @@ export const formatCurrencyAmount = ({
   if (withScale) {
     formattedAmount = amount !== undefined ? formatAmountWithScale(amount) : '0'
   } else {
-    formattedAmount = amount !== undefined ? formatAmount(amount) : '0'
+    formattedAmount =
+      amount !== undefined
+        ? formatAmount(amount, {
+            maximumFractionDigits:
+              currency === V2V3_CURRENCY_ETH ? PRECISION_ETH : PRECISION_USD,
```
aeolianeth, line 38: this is effectively the fix

PR #4355, opened 2024-06-10
fix: sometimes tx's never confirming on mainnet

johnnyd-eth:
See comment for explanation

### DIFF
--- a/src/contexts/Transaction/TxHistoryProvider.tsx
+++ b/src/contexts/Transaction/TxHistoryProvider.tsx
@@ -22,18 +22,25 @@ const pollTransaction = async (
   if (!txLog.tx?.hash || txLog.status !== TxStatus.pending) {
     return txLog
   }
-
   const response = await readProvider.getTransaction(txLog.tx.hash)
-  // if no response, then the tx is cancelled.
+  const oneMinuteAgo = nowSeconds() - 1 * 60
   if (!response) {
-    txLog.callbacks?.onCancelled?.(response)
+    // if no response and tx was created over a minute ago, then assume the tx is cancelled.
+    if (oneMinuteAgo > txLog.createdAt) {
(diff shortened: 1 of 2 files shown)


src/contexts/Transaction/TxHistoryProvider.tsx
```diff
@@ -22,20 +22,27 @@ const pollTransaction = async (
   if (!txLog.tx?.hash || txLog.status !== TxStatus.pending) {
     return txLog
   }
-
   const response = await readProvider.getTransaction(txLog.tx.hash)
-  // if no response, then the tx is cancelled.
+  const oneMinuteAgo = nowSeconds() - 1 * 60
+  // if no response and tx was created over a minute ago, then the tx is cancelled.
   if (!response) {
```
johnnyd-eth, line 28: I think the problem that this was getting called too quickly. It would give a null response instantly, tx would get marked as failed, and not be checked again. So I added a delay here, so tx doesn't get marked as failed unless the null response occurs 1min after the tx was created

src/contexts/Transaction/TxHistoryProvider.tsx
```diff
@@ -22,18 +22,25 @@ const pollTransaction = async (
   if (!txLog.tx?.hash || txLog.status !== TxStatus.pending) {
     return txLog
   }
-
   const response = await readProvider.getTransaction(txLog.tx.hash)
-  // if no response, then the tx is cancelled.
+  const oneMinuteAgo = nowSeconds() - 1 * 60
   if (!response) {
-    txLog.callbacks?.onCancelled?.(response)
+    // if no response and tx was created over a minute ago, then the tx is cancelled.
```
aeolianeth, line 28: maybe just say here "we assume it's failed". Cos in theory i guess it might not actually have failed.

aeolianeth (approved):

PR #4367, opened 2024-06-12
refactor: organize version-specific stuff by version directory

aeolianeth:
Idea: completely isolate all JB version-specific modules under a single "version" directory.

Why:

- allows dev to clearly identify what version a specific piece of code is for
- allow for a clearer distinction between shared/common and version-specific code
- (future thinking) in theory, allows us to actually compile each package separately, meaning that each version could have their own dependencies.

What:

- Create "version" directories under `packages` (`packages/v1`, `packages/v2v3`)
- Move all hooks, components, constants, utils etc within their respective package directories


aeolianeth:
Closing, POC

PR #4369, opened 2024-07-02
Init v4 project page with wagmi and sdk

aeolianeth:


src/packages/v4/hooks/useNativeTokenSurplus.ts
```diff
@@ -0,0 +1,20 @@
...
+/**
+ * @todo add to sdk
+ */
+export function useNativeTokenSurplus() {
+  const {
+    projectId,
+    contracts: { primaryNativeTerminal },
+  } = useJBContractContext()
+
+  return useReadJbMultiTerminalCurrentSurplusOf({
```
aeolianeth, line 16: All the read hooks come from the SDK. This hook (`useNativeTokenSurplus`) is an just an example of where we're wrapping one of the sdk hooks. But per the TODO, this hook can probably live in the SDK (it's a super common hook that many people could use).

src/packages/v4/hooks/useNativeTokenSymbol.ts
```diff
@@ -0,0 +1,22 @@
...
+ */
+export function useNativeTokenSymbol() {
+  const symbols: { [k: number]: string } = {
+    [sepolia.id]: 'SepETH',
+    [optimismSepolia.id]: 'OPSepETH',
+    [arbitrumSepolia.id]: 'ArbSepETH',
+    [baseSepolia.id]: 'BaseSepETH',
+  }
+
+  const chainId = useJBChainId()
```
aeolianeth, line 20: JBChainId is whatever we pass in to the JBProjectProvider. Its used for all the contract reads (e.g. do all the reads on `sepolia` instead of `mainnet`.

src/packages/v4/components/ActivityList/ActivitiyList.tsx
```diff
@@ -0,0 +1,95 @@
...
+export function ActivityList() {
+  const { projectId } = useJBContractContext()
+  const { data } = useSubgraphQuery(PayEventsDocument, {
+    orderBy: PayEvent_OrderBy.timestamp,
+    orderDirection: OrderDirection.desc,
+    where: {
+      // pv: PV2,
+      projectId: Number(projectId),
+    },
+  })
```
aeolianeth, line 77: New approach to querying the subgraph. Use `useSubgraphQuery`, then pass in the generated graphql document (e.g. `PayEventsDocument`), followed by the query args.

It might be nice to have these as nice wrapper hooks (e.g. create a hook called `usePayEvents` that just calls this under the hood.

But lets only do this if it becomes a huge pain. THis good for now.

src/pages/v4/[chainName]/p/[projectId]/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+  )
+}
+
+const Providers: React.FC<
+  PropsWithChildren & { chainId: JBChainId; projectId: bigint }
+> = ({ chainId, projectId, children }) => {
+  return (
+    <ReactQueryProvider>
+      <WagmiProvider config={wagmiConfig}>
+        <JBProjectProvider
```
Comment (aeolianeth) [inline src/pages/v4/[chainName]/p/[projectId]/index.tsx:48]: JBProjectProvider loads up most of the project's data, and makes various contexts available.

Full list here: https://github.com/Bananapus/juice-sdk-v4/blob/351a46e13a42f3b0341575332a651c1d98aa817c/packages/react/src/contexts/JBProjectProvider/JBProjectProvider.tsx#L46-L56

Example: you should be able to get project metadata anywhere using:

```ts
const { metadata } = useJBProjectMetadataContext()
const { data, loading } = { metadata }
const projectName = data?.name
```

Throughout the SDK, most of the data is supplied with its loading state alongside (as shown above).

src/pages/v4/[chainName]/p/[projectId]/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+  const [hasMounted, setHasMounted] = React.useState(false)
+  React.useEffect(() => {
+    setHasMounted(true)
+  }, [])
+
+  if (!hasMounted) {
+    return null
+  }
+
+  if (!featureFlagEnabled(FEATURE_FLAGS.V4)) {
```
Comment (aeolianeth) [inline src/pages/v4/[chainName]/p/[projectId]/index.tsx:73]: run `jb.enableFeatureFlag('v4')` on `http://localhost:3000`

src/packages/v4/components/ActivityList/ActivitiyList.tsx
johnnyd-eth: Sounds good

src/packages/v4/hooks/useNativeTokenSurplus.ts
johnnyd-eth: Copy

src/packages/v4/hooks/useNativeTokenSymbol.ts
johnnyd-eth: Ok got it

johnnyd-eth (approved):
Got it, thanks for all the comments. LGTM!

PR #4375, opened 2024-07-09
feat: fetch data for v4 project header [2/n]

johnnyd-eth:
- Uses subgraph query to fetch project stats 
- Fetches `projectOwnerAddress` from SDK's `useReadJbProjectsOwnerOf` (currently failing, returning undefined?)


src/packages/v4/views/V4ProjectDashboard/hooks/useV4ProjectHeader.ts
```diff
@@ -24,20 +28,37 @@ export interface ProjectHeaderData {
 export const useV4ProjectHeader = (): ProjectHeaderData => {
   const router = useRouter()
   const { projectMetadata, projectId } = useProjectMetadataContext()
```
aeolianeth, line 30: Yeah I think we should just not use these shared contexts if we can avoid it, in V4 code. Lets try stick to the SDK-given contexts

src/packages/v4/views/V4ProjectDashboard/hooks/useV4ProjectHeader.ts
```diff
@@ -24,20 +28,37 @@ export interface ProjectHeaderData {
...
+    },
+  )
+
+  const projectStatsData = data?.projects?.[0]
+  const {
+    createdAt,
+    volume: totalVolumeStr,
+    trendingVolume: trendingVolumeStr,
+    paymentsCount,
+  } = projectStatsData ?? {
```
aeolianeth, line 53: Hmm, should we have fallbacks like this? Do we need them?

johnnyd-eth: I just wasn't sure how best to handle `projectStatsData` possibly being undefined otherwise?

src/packages/v4/views/V4ProjectDashboard/hooks/useV4ProjectHeader.ts
johnnyd-eth: Yeah ok good idea

PR #4376, opened 2024-07-09
feat: v4 permissions and `hasPermissions` hook used in project header [3/n]

johnnyd-eth:
- Renames `permissions` (v2v3) to `v2v3Permissions`
- Creates `v4Permissions`(permissionIds) (couldn't see `permissionIds` in the SDK -> **should I add them?**)
- Creates `useV4WalletHasPermissions` (checks if current wallet has a particular permission on current project)
- Conditionally show the headers settings button depending on permission


src/packages/v4/hooks/useV4WalletHasPermission.ts
```diff
@@ -0,0 +1,40 @@
...
+import { isEqualAddress } from 'utils/address'
+import { V4OperatorPermission } from '../models/v4Permissions'
+
+export function useV4WalletHasPermission(
+  permission: V4OperatorPermission | V4OperatorPermission[],
+): boolean {
+  const { userAddress } = useWallet()
+
+  const { projectId } = useContext(ProjectMetadataContext)
+  const _projectId = BigInt(projectId ?? 0)
```
aeolianeth, line 15: Could we just use projectId from an SDK context? Then we don't have to convert back to bigint

src/packages/v4/models/v4Permissions.ts
```diff
@@ -0,0 +1,40 @@
+export enum V4OperatorPermission {
```
aeolianeth, line 1: Beast! Thank you! Yeah we should sdk this. Good here for now

src/packages/v4/hooks/useV4WalletHasPermission.ts
johnnyd-eth: Oh yep, which would this be? 😅

PR #4379, opened 2024-07-11
feat: v4 current and upcoming fc data pt. 1

johnnyd-eth:
- Should have real data for these 3 cards in the current and upcoming ruleset tabs: (SDK's `useJBRuleset` and `useReadJbRulesetsLatestQueuedOf` currently both returning undefined)
Screenshot 2024-07-11 at 12 11 30 pm

- Moves some countdown clock components to shared folders


src/models/ballot.ts
```diff
@@ -11,6 +11,15 @@ export enum V2BallotState {
   'Failed' = 2,
 }
 
+export enum V4ApprovalStatus {
```
aeolianeth, line 14: Do u wanna just add a todo to add this to the sdk? should be there imo

src/packages/v4/contexts/RulesetCountdownProvider.tsx
```diff
@@ -0,0 +1,43 @@
...
+}
+
+export const RulesetCountdownContext =
+  createContext<RulesetCountdownContextType>({
+    timeRemainingText: '0d 0h 0m 0s',
+    endEpochSeconds: 0,
+    secondsRemaining: 0,
+  })
+
+export const RulesetCountdownProvider = ({
```
aeolianeth, line 18: Great one

src/models/ballot.ts
johnnyd-eth: Yep will do. Do you want me to start adding stuff to the SDK? I'm not sure how to atm but would love to learn

src/components/Project/hooks/useCountdownClock.ts
```diff
@@ -1,5 +1,5 @@
 import { useEffect, useState } from 'react'
-import { timeSecondsToDateString } from '../utils/timeSecondsToDateString'
+import { timeSecondsToDateString } from '../../../utils/timeSecondsToDateString'
```
aeolianeth, line 2: if its in the root, we can do:

```suggestion
import { timeSecondsToDateString } from 'utils/timeSecondsToDateString'
```

johnnyd-eth: Yeah gotcha, this is happening automatically when I move the files around and I'm missing some. I'll keep a better eye on them going forward.

PR #4380, opened 2024-07-11
feat: v4 current and upcoming fc data pt. 2

johnnyd-eth:
- Pulls data for this section (currently not working because SDK's `useJBRuleset` returning undefined)
Screenshot 2024-07-11 at 5 59 25 pm


src/packages/v2v3/components/V2V3Project/ProjectDashboard/components/CyclesPayoutsPanel/hooks/useConfigurationPanel/useFormatConfigurationCyclesSection.ts
```diff
@@ -8,9 +8,9 @@ import { MAX_DISTRIBUTION_LIMIT } from 'packages/v2v3/utils/math'
 import { useMemo } from 'react'
 import { formatCurrencyAmount } from 'utils/format/formatCurrencyAmount'
 import { fromWad } from 'utils/format/formatNumber'
-import { formatTime } from 'utils/format/formatTime'
+import { formatTimeBN } from 'utils/format/formatTime'
 import { timeSecondsToDateString } from 'utils/timeSecondsToDateString'
-import { pairToDatum } from '../../utils/pairToDatum'
+import { pairToDatum } from '../../../../../../../../../components/Project/ProjectHeader/utils/pairToDatum'
```
aeolianeth, line 13: jezuz

src/utils/format/formatTime.ts
```diff
@@ -141,3 +141,21 @@ export const formatTime = (timestamp: BigNumber | undefined) => {
   const timeWeekdayString = timeDate.toLocaleString('en-US', formatOptions)
   return isoDateString + ', ' + timeWeekdayString
 }
+
+export const formatTime = (timestamp: bigint | undefined) => {
```
aeolianeth, line 145: I feel like if this is the exact same as the other one, we should either just nuke the `formatTimeBN` and replace uses with the new bigint fn.

`formatTime(BigInt(myBigInt.toString()))`

In fact, maybe we should have a function called `toBigInt`

```js
function toBigInt(bn: BigNumber | undefined) {
    return bn ? BigInt(bn.toString()) : undefined
}
```

So then the pattern for 'migrating' to bigint becomes:
- identify hook or utility that could be bigint'd
- rewrite it with bigint, accept bigint params etc
- refactor existing usages to transform bignumber to bigint with `toBigInt`.

WDYT?

johnnyd-eth: Good call, done for this instance

PR #4388, opened 2024-07-16
fix: ruleset countdown, upcoming "Ruleset #" and "Ruleset duration" cards

johnnyd-eth:
Screenshot 2024-07-16 at 1 07 21 pm

Screenshot 2024-07-16 at 1 07 25 pm


src/pages/v4/[chainName]/p/[projectId]/index.tsx
```diff
@@ -53,9 +54,11 @@ const Providers: React.FC<
             metadata: { ipfsGatewayHostname: OPEN_IPFS_GATEWAY_HOSTNAME },
           }}
         >
-          <V4ProjectMetadataProvider projectId={BigInt(projectId)}>
-            {children}
-          </V4ProjectMetadataProvider>
+          <RulesetCountdownProvider>
```
Comment (aeolianeth) [inline src/pages/v4/[chainName]/p/[projectId]/index.tsx:57]: Does the whole page need to be aware of this context? Could it just be scoped to wherever the countdown is?

Comment (johnnyd-eth) [reply src/pages/v4/[chainName]/p/[projectId]/index.tsx]: Good call, moved it

PR #4391, opened 2024-07-17
feat: v4 rules card, "OTHER RULES" section [1/n] 

johnnyd-eth:
Screenshot 2024-07-17 at 1 54 16 pm


src/packages/v4/hooks/useJBUpcomingRuleset.ts
```diff
@@ -0,0 +1,47 @@
...
+export function useJBUpcomingRuleset(): { 
+  ruleset: Ruleset | undefined, 
+  rulesetMetadata: JBRulesetMetadata | undefined, 
+  isLoading: boolean
+ } {
+  const { contracts, projectId } = useJBContractContext()
+  const { data, isLoading } = useReadJbControllerUpcomingRulesetOf({
+    address: contracts.controller?.data ?? undefined,
+    args: [projectId]
+  })
```
aeolianeth, line 16: no biggie but we can use the `select` function on this hook to make the below data transformations a lot easier. DW though, we'l move these to sdk eventually

aeolianeth (approved):

src/packages/v4/hooks/useJBUpcomingRuleset.ts
johnnyd-eth: Ok got it 💯

PR #4402, opened 2024-07-20
refactor: refine tx history types

aeolianeth:
Some more tx history prep. Makes types as minimal as possible, so that we can more easily implement the wagmi tx history provider.

### DIFF
--- a/src/contexts/Transaction/TxHistoryContext.ts
+++ b/src/contexts/Transaction/TxHistoryContext.ts
@@ -1,17 +1,23 @@
-import { providers } from 'ethers'
-import { TransactionCallbacks, TransactionLog } from 'models/transaction'
+import { TransactionCallbacks, TxStatus } from 'models/transaction'
 import { createContext } from 'react'
 
-// Prefer using tx.timestamp if tx has been mined. Otherwise use createdAt timestamp
-export const timestampForTxLog = (txLog: TransactionLog) => {
-  return (
-    (txLog.tx as providers.TransactionResponse)?.timestamp ?? txLog.createdAt
-  )
+export type TransactionType = {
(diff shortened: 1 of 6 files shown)


src/contexts/Transaction/TxHistoryContext.ts
```diff
@@ -1,17 +1,23 @@
...
-import { TransactionCallbacks, TransactionLog } from 'models/transaction'
+import { TransactionCallbacks, TxStatus } from 'models/transaction'
 import { createContext } from 'react'
 
-// Prefer using tx.timestamp if tx has been mined. Otherwise use createdAt timestamp
-export const timestampForTxLog = (txLog: TransactionLog) => {
-  return (
-    (txLog.tx as providers.TransactionResponse)?.timestamp ?? txLog.createdAt
-  )
+export type TransactionType = {
```
aeolianeth, line 4: AFAICT, we only actually use these properties on transactions. Don't need the whole `providers.TransactionResponse`

src/hooks/useTransactor.ts
```diff
@@ -134,10 +134,17 @@ export function useTransactor(): Transactor | undefined {
 
         // add transaction to the history UI
         const txTitle = options?.title ?? functionName
-        addTransaction?.(txTitle, result as providers.TransactionResponse, {
-          onConfirmed: options?.onConfirmed,
-          onCancelled: options?.onCancelled,
-        })
+        addTransaction?.(
+          txTitle,
+          {
```
aeolianeth, line 139: Again, only pass thru what parts of the tx we actually need

johnnyd-eth (approved):
Nicee LGTM

PR #4403, opened 2024-07-20
wagmi tx history

aeolianeth:


src/contexts/Transaction/WagmiTxHistoryProvider.tsx
```diff
@@ -1,14 +1,94 @@
...
+}
+
 export default function WagmiTxHistoryProvider({
   children,
 }: {
   children: ReactNode
 }) {
-  const { transactions, addTransaction, removeTransaction } = useTransactions()
-  // TODO implement polling/wait logic using Wagmi
+  const { transactions, addTransaction, removeTransaction, setTransactions } =
```
aeolianeth, line 45: This all ends up doing the exact same thing as the Ethers variant, except the `pollTransaction` function is wagmi-specific.

I tried to improve it but meh.

johnnyd-eth (approved):

PR #4407, opened 2024-07-24
feat: V4 payouts table [1/n]

johnnyd-eth:
- Copied PayoutsTable directory from V2V3 to V4, switched all v2v3 references to v4 in all the components with the directory
- Also copied the necessary `Allocation` components that `PayoutsTable` relies on to V4, adjust them where necessary
- Use `PayoutsTable` in the V4 distribute payouts modal:

Screenshot 2024-07-24 at 9 29 39 pm


src/components/Create/components/pages/PayoutsPage/components/CreateFlowPayoutsTable.tsx
```diff
@@ -7,10 +7,10 @@ import {
   getV2V3CurrencyOption,
 } from 'packages/v2v3/utils/currency'
 import { MAX_DISTRIBUTION_LIMIT } from 'packages/v2v3/utils/math'
+import { allocationToSplit, splitToAllocation } from 'packages/v2v3/utils/splitToAllocation'
```
aeolianeth, line 10: Hmm... maybe we should move create into packages/v2v3 ...kinda makes sense actually

src/packages/v2v3/components/V2V3Project/ProjectDashboard/components/NftRewardsPanel/NftReward/PreviewAddRemoveNftButton.tsx
```diff
@@ -3,7 +3,7 @@ import { Trans, t } from '@lingui/macro'
 import { Button } from 'antd'
 import { useCallback } from 'react'
 import { twMerge } from 'tailwind-merge'
-import { emitConfirmationDeletionModal } from '../../../utils/modals'
+import { emitConfirmationDeletionModal } from '../../../../../../../../hooks/emitConfirmationDeletionModal'
```
aeolianeth, line 6: possible to sort out some of these bong import paths? 😅

johnnyd-eth: Ah my bad, will fix

src/components/Create/components/pages/PayoutsPage/components/CreateFlowPayoutsTable.tsx
johnnyd-eth: You mean moving entire /create dir to v2v3?

aeolianeth: yeah dw for now. just an idea

src/packages/v4/models/split.ts
```diff
@@ -0,0 +1,11 @@
+import { JBSplit, SplitPortion } from "juice-sdk-core";
+import { zeroAddress } from "viem";
+
+export const defaultSplit: JBSplit = {
```
aeolianeth, line 4: i think we can move defaultSplits to the spot where it's used, and remove this file. WDYT?

src/packages/v4/utils/distributions.ts
```diff
@@ -0,0 +1,260 @@
...
+ * @param splits {Split[]} - list of current splits to possibly have their percents adjusted
+ * @returns {Split[]} splits with their percents adjusted
+ */
+export function ensureSplitsSumTo100Percent({
+  splits,
+}: {
+  splits: Split[]
+}): Split[] {
+  // Calculate the percent total of the splits
+  const currentTotal = splits.reduce((sum, split) => sum + split.percent.toFloat(), 0)
```
aeolianeth, line 111: Just a comment, no need to fix now: wouldn't it be more accurate to use bigint throughout this function?

aeolianeth (approved):

src/packages/v4/utils/distributions.ts
johnnyd-eth: It would probably be more clear, but IIRC, the frontend code is as precise as it possibly can be in terms of rounding. It's actually limited by the contract payouts which support ~10dp. 

Even at this level of precision sometimes situations can arise where we try to submit payouts to the contracts where the total payout percent is 100.000000001, which causes a error.  So this function is a final safeguard against that.

src/packages/v4/models/split.ts
johnnyd-eth: Good idea, done

PR #4420, opened 2024-08-07
V4 project page: volume charts [1/n]

johnnyd-eth:
Screenshot 2024-08-07 at 12 22 28 pm

### DIFF
--- a/src/components/VolumeChart/hooks/useProjectTimeline.ts
+++ b/src/components/VolumeChart/hooks/useProjectTimeline.ts
@@ -1,9 +1,12 @@
 import { useQuery } from '@tanstack/react-query'
+import { PV_V4 } from 'constants/pv'
 import { readProvider } from 'constants/readProvider'
 import EthDater from 'ethereum-block-by-date'
-import { useProjectTlQuery } from 'generated/graphql'
+import { ProjectTlQuery, useProjectTlQuery } from 'generated/graphql'
 import { client } from 'lib/apollo/client'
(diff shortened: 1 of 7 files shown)


src/components/VolumeChart/hooks/useProjectTimeline.ts
```diff
@@ -74,22 +77,19 @@ export function useProjectTimeline({
       ...blocks,
     },
   })
-  // TODO: const { data: v4QueryResult } = useSubgraphQuery(ProjectTlDocument, {
-  //   where: {
-  //     projectId,
-  //   },
-  // })
+  const { data: v4QueryResult } = useSubgraphQuery(ProjectTlDocument, {
```
aeolianeth, line 80: Both `useProjectTlQuery` and `useSubgraphQuery` are going to run though, when we only want one or the other to run.

Maybe the easiest fix is to set the `enabled` flag on each query, to only be enabled when the project has the correct version.

https://tanstack.com/query/v4/docs/framework/react/guides/disabling-queries

johnnyd-eth: Good catch, I've modified the definition of the `useSubgraphQuery` function so we can pass it an object with `enabled` like here:

Screenshot 2024-08-08 at 10 42 30 am

And then have used the equivalent of `enabled` for Apollo on the old v1v2v3 query above (`skip`)

aeolianeth (approved):

PR #4444, opened 2024-08-29
feat: V4 create behind feature flag [1/n]

johnnyd-eth:
- Copies v2v3 create dir to v4
- Modify pages/create to enable v4 behind feature flag, plus wrap v4 in required Providers: https://github.com/jbx-protocol/juice-interface/pull/4444/files#r1735464428


src/pages/create/index.tsx
johnnyd-eth, line 1: This is the only diff outside of copied files

src/pages/create/index.tsx
```diff
@@ -1,16 +1,51 @@
...
+        </V2V3CurrencyProvider>
+      </TransactionProvider>
+    </V2V3ContractsProvider>
+  )
+
+  if (featureFlagEnabled(FEATURE_FLAGS.V4)) {
+    contentByVersion = (
+      <WagmiProvider config={wagmiConfig}>
+        {/* Hardcode V4 create to Sepoila for now */}
+        <JBProjectProvider
```
johnnyd-eth, line 37: Sets the network for `useJBContractContext` to use.

src/packages/v4/components/Create/hooks/useLoadInitialStateFromQuery.ts
```diff
@@ -0,0 +1,178 @@
...
+ */
+export function useLoadingInitialStateFromQuery() {
+  const dispatch = useDispatch()
+  const router = useRouter()
+  const { contracts } = useJBContractContext()
+
+  const [loading, setLoading] = useState(true)
+
+  useEffect(() => {
+    if (!router.isReady || !contracts.primaryNativeTerminal.data) return
```
johnnyd-eth, line 125: Terminal address loaded for network given here: https://github.com/jbx-protocol/juice-interface/pull/4444/files#r1735468140

src/pages/create/index.tsx
```diff
@@ -1,16 +1,51 @@
...
+          chainId={NETWORKS_BY_NAME[NetworkName.sepolia].chainId as 11155111}
```
johnnyd-eth, line 38: Sets the network for useJBContractContext to use. Ideally we would read from `process.env.NEXT_PUBLIC_INFURA_NETWORK`, but I think this will require us to first update `jbDirectoryAddress` in the SDK to include mainnet (chainId=1).

Screenshot 2024-08-29 at 10 18 59 am

PR #4517, opened 2024-11-25
hide payouts create zero amount

wraeth-eth:
This PR fixes a bug raised by @mejango where the payouts recipients area is shown when 'zero' amount is selected:
image

This code hides it:
image


aeolianeth (approved):
TY @wraeth-eth lgtm

PR #4624, opened 2025-05-30
Aggregate multi-chain projects in UI

peripheralist:
- Store new `sucker_group_id` column in db projects table
- Create new database view that allows the UI to consume project "aggregates" to display in project lists

The `projects_aggregate` view coalesces projects from the projects table with a shared `suckerGroupId`. Most importantly, the view includes a derived `chain_ids` column that contains the `chainId` of each matching project. For pre-v4 projects this will always be simply `[1]`. Unique columns like `projectId`, `handle` (including `chainId`), metadata properties, etc will be taken from the project with the lowest `chainId` value, ensuring project cards can still link to a valid url generated from `chainId` + `projectId`, preferring the "eth" project if it exists. Importantly, this pattern assumes that the metadata properties of each matching project are identical (in any case, only the properties of the lowest-chainId project will be shown).

### DIFF
--- a/supabase/migrations/20250522201752_add_sucker_group_id_aggregate_view.sql
+++ b/supabase/migrations/20250522201752_add_sucker_group_id_aggregate_view.sql
@@ -1,0 +1,103 @@
+ALTER TABLE public.projects
+RENAME COLUMN redeem_voume_usd to redeem_volume_usd;
+ALTER TABLE public.projects
+ADD COLUMN IF NOT EXISTS sucker_group_id text;
+
+drop view if exists projects_aggregate;
+
+create view projects_aggregate as
+(
+  -- One project per suckerGroupId + array of chainIds
+  select distinct on (p.sucker_group_id)
+    p.id,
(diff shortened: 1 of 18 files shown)


supabase/migrations/20250522201752_add_sucker_group_id_aggregate_view.sql
```diff
@@ -0,0 +1,103 @@
...
+    sub.trending_payments_count as trending_payments_count,
+    sub.trending_score as trending_score,
+    sub.trending_volume as trending_volume,
+    sub.volume as volume,
+    sub.volume_usd as volume_usd
+  from projects p
+  join (
+    select 
+      sucker_group_id, 
+      coalesce(array_agg(distinct chain_id) filter (where chain_id is not null), '{}') as chain_ids,
```
peripheralist, line 47: we use `coalesce` to enforce non-null values at runtime (so we can manually type properties as non-null in the UI)

